PR #3, opened 2020-07-22
add tags

joelhooks:


lib/courses.js
```diff
@@ -1,73 +1,32 @@
...
-export function getAllCourseLessonIds() {
-  return getCourseLessons().map((lesson) => {
-    return {
-      params: {
-        ...lesson,
-        id: lesson.slug,
-      },
+import {request} from 'graphql-request'
+
+export async function loadCourse(slug) {
```
joelhooks, line 3: this is rad imo.

lib/instructors.js
```diff
@@ -1,16 +1,16 @@
...
+  pick(props, find(selector, data)),
 )
 
-const instructorsPath = "./data/instructors.json"
+const instructorsPath = './data/instructors.json'
 
 function readInstructors() {
   const rawdata = fs.readFileSync(instructorsPath)
-  return JSON.parse(rawdata)
+  return uniqBy('slug', JSON.parse(rawdata))
```
joelhooks, line 13: something is weird in the data and there were dupes that caused issues. Might be related to the issue you described @johnlindquist ?

lib/tags.js
```diff
@@ -0,0 +1,33 @@
...
+      params: {
+        slug,
+        label,
+      },
+    }
+  })(tags)
+}
+
+export function getTag(slug) {
+  return find({slug}, readTags())
```
joelhooks, line 32: I wanted the full data so no need to pluck

pages/courses/[slug].js
```diff
@@ -55,16 +61,10 @@ export default function Course({courseData}) {
...
 }
 
-export async function getStaticPaths() {
-  const paths = getAllCourseIds()
-  return {
-    paths,
-    fallback: false,
-  }
-}
+export async function getServerSideProps({res, params}) {
```
Comment (joelhooks) [inline pages/courses/[slug].js:64]: using graphql there is no need to pre-render this statically. It's fast af.

pages/instructors/index.js
```diff
@@ -3,20 +3,20 @@ import Link from 'next/link'
 
 export default function Instructors({instructors}) {
   return (
-    <ul className="list-disc">
+    <div className="grid grid-cols-3 gap-4">
```
joelhooks, line 6: grid is neat

lib/instructors.js
johnlindquist: I investigated and found the dupes, but I think it's just from how we're doing the `loadStaticDataToFiles.js` pagination or something.

They do not appear as dupes in blazer queries or in the graphql request

PR #4, opened 2020-07-26
Jh/auth

joelhooks:


.babelrc
```diff
@@ -0,0 +1,3 @@
+{
+  "presets": ["next/babel"]
```
joelhooks, line 2: this is for the testing stuff that I haven't figured out

.env.development
```diff
@@ -0,0 +1,3 @@
+NEXT_PUBLIC_AUTH_DOMAIN="https://egghead.io"
+NEXT_PUBLIC_CLIENT_ID=d25ada502a252a89c5f80fbbae70320785cd73cd54e46a00226ddcf6a446df0c
+NEXT_PUBLIC_REDIRECT_URI=https://badass.dev
```
joelhooks, line 3: authing redirect to badass.dev for now

context/EggheadUserProvider.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
```
joelhooks, line 1: all of this is imported and frankly it feels confusing and could use some attention and thought 🤔

pages/lessons/[slug].js
```diff
@@ -32,9 +54,32 @@ const Transcript = ({url}) => {
...
+      ...authorizationHeader,
+    },
+  })
+  return graphQLClient.request(query, variables)
+}
+
+export default function Lesson({initialLesson}) {
   const router = useRouter()
   const playerRef = React.useRef(null)
+  const {authToken} = useEggheadUser()
```
Comment (joelhooks) [inline pages/lessons/[slug].js:75]: when and where hooks get called is confusing and a silent killer

pages/lessons/[slug].js
```diff
@@ -32,9 +54,32 @@ const Transcript = ({url}) => {
   return data ? <Markdown className="prose">{data.text}</Markdown> : null
 }
 
-export default function Lesson({lesson}) {
+const lessonLoader = (slug, token) => (query) => {
```
Comment (joelhooks) [inline pages/lessons/[slug].js:57]: Is this a partial application? 🥳

Comment (johnlindquist) [reply pages/lessons/[slug].js:57]: let loadByToken = _.partial(lessonLoader, "some-slug")

would be partial application 🤓

Because you would be taking a function with 2 args and only passing in 1 (resulting in a new function waiting for a `slug`).

pages/lessons/[slug].js
Comment (johnlindquist) [reply pages/lessons/[slug].js:75]: My understanding: Hooks are always called, but most React hooks (useState, etc) are memo'ed/cached.

So when creating custom hooks, if you're using `useState`, those values should be cached even if it looks like your hook is constantly invoked. I think of hooks like "getters" in that sense.

context/EggheadUserProvider.js
johnlindquist: I know the full import for "React" is required when the babel config looks for the jsx pragma.

I'm fairly certain Next.js uses:

https://github.com/vslinko/babel-plugin-react-require

to automatically inject "React" to satisfy the babel/jsx/pragma lookup.

PR #11, opened 2020-07-30
adds back the auth hook and viewer context

joelhooks:
this does NOT turn auth.js into a typed file. That needs to
happen for sure. Also consolidated the hooks and whatnot into a
single file.

### DIFF
--- a/utils/auth.js
+++ b/utils/auth.js
@@ -145,21 +145,26 @@ export default class Auth {
   }
 
   refreshUser(accessToken, loadFullUser = false) {
-    if (typeof localStorage === 'undefined') {
-      return
-    }
-    return http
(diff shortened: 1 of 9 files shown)


utils/auth.js
```diff
@@ -88,13 +88,13 @@ export default class Auth {
     }
   }
 
-  handleAuthentication(location) {
-    if (typeof localStorage === 'undefined') {
-      return
-    }
+  handleAuthentication() {
     return new Promise((resolve, reject) => {
+      if (typeof localStorage === 'undefined') {
```
joelhooks, line 93: todo: based on conversations with Jared we should be using cookies to access the auth from the SSR

PR #16, opened 2020-07-30
Jl/instructors page

johnlindquist:
Adding /instructors functionality

### DIFF
--- a/cypress/integration/instructors_spec.js
+++ b/cypress/integration/instructors_spec.js
@@ -1,0 +1,13 @@
+describe('instructors', () => {
+  it('/instructors loads', () => {
+    cy.visit('/instructors')
+    cy.get('img').should('have.length', 24)
+  })
+})
+
+describe('instructor/john-lindquist', () => {
+  it('instructors/john-lindquist loads', () => {
+    cy.visit('/instructors/john-lindquist')
+    cy.get('img').should('have.length', 1)
+  })
+})
--- a/lib/instructors.ts
+++ b/lib/instructors.ts
@@ -1,36 +1,17 @@
-import fs from 'fs'
-import {compose, curry, map, pick, find, uniqBy} from 'lodash/fp'
+import {request} from 'graphql-request'
 
-const mapProps = compose(map, pick)
-const findPick = curry((selector, props, data) =>
-  pick(props, find(selector, data)),
-)
+export async function loadInstructors(page = 1) {
+  const endpoint = 'https://egghead.io/graphql'
 
-const instructorsPath = './data/instructors.json'
-
-function readInstructors() {
-  const rawdata = fs.readFileSync(instructorsPath)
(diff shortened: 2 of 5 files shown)


cypress/integration/instructors_spec.js
```diff
@@ -0,0 +1,13 @@
+describe('instructors', () => {
+  it('/instructors loads', () => {
+    cy.visit('/instructors')
+    cy.get('img').should('have.length', 24)
```
joelhooks, line 4: Is this testing the number of instructors on the page? I'd like to chat about testing strategy if you're up for it.

lib/instructors.ts
```diff
@@ -1,36 +1,31 @@
...
-function readInstructors() {
-  const rawdata = fs.readFileSync(instructorsPath)
-  return uniqBy('slug', JSON.parse(rawdata.toString()))
-}
-
-export function getInstructors(props) {
-  const instructors = readInstructors()
+export async function loadInstructors(page: number = 1) {
+  const query = `query getInstructors($page: Int!){
+    instructors(per_page: 24, page:$page){
```
joelhooks, line 7: I don't think there's a big need for pagination here. 

Might be wrong but I would generally prefer to not paginate when possible and I think we could load them all and aggressively cache. 

I actually think that instructors, like tags, make sense static. They don't change much and there is t a lot of them

pages/instructors/index.tsx
```diff
@@ -1,38 +1,41 @@
...
-export function getStaticProps() {
-  const instructors = getInstructors([
-    'id',
-    'slug',
-    'full_name',
-    'website',
-    'path',
-    'avatar_480_url',
-  ])
+export async function getServerSideProps({res, params}) {
```
joelhooks, line 32: Because the instructor list changes rarely we might consider this staying SSG? Just a thought.

cypress/integration/instructors_spec.js
johnlindquist: Yeah, it's just testing for 24 images (like egghead has now). I didn't put any thought into it, just wanted to stub it out.

lib/instructors.ts
johnlindquist: We briefly discussed "sorting" and "featuring" instructors based on the query and making /instructors more engaging in general. I think it's worth keeping dynamic to we can play with it more until we settle on features.

cypress/integration/instructors_spec.js
joelhooks: I'd test this `>1` if that was possible. Something like that. 

We've got an ongoing issue of hyper specified tests at egghead so I've got some trauma 😂

johnlindquist: We can test if "body" exists or whatever. It's based on Chrome APIs, so it's a similar mindset to using the Chrome console to inspect the DOM. I'll change it to >1

PR #17, opened 2020-07-31
Jh/better base url

joelhooks:


lib/config.ts
```diff
@@ -0,0 +1,6 @@
+const config = {
```
joelhooks, line 1: this config makes a ton of sense ❤️

.vscode/settings.json
```diff
@@ -1,5 +1,5 @@
 {
   "javascript.format.insertSpaceAfterOpeningAndBeforeClosingNonemptyBraces": false,
   "javascript.preferences.quoteStyle": "single",
-  "typescript.tsdk": "node_modules/typescript/lib",
-}
\ No newline at end of file
+  "typescript.tsdk": "node_modules/typescript/lib"
```
joelhooks, line 4: This one is nice because VS Code mismatches and it gets weird.

components/ReactPlayer/props.js
```diff
@@ -1,5 +1,5 @@
 import PropTypes from 'prop-types'
-import noop from '../../lib/noop'
+import noop from '../../utils/noop'
```
joelhooks, line 2: Util or Lib? lol

Util.

jest.config.js
```diff
@@ -5,7 +5,12 @@ module.exports = {
     '!**/node_modules/**',
   ],
   setupFilesAfterEnv: ['<rootDir>/setupTests.js'],
-  testPathIgnorePatterns: ['/node_modules/', '/.next/', '/.vercel/'],
+  testPathIgnorePatterns: [
+    '/node_modules/',
+    '/.next/',
+    '/.vercel/',
+    '/cypress/',
```
joelhooks, line 12: Jest was picking these up.

lib/search-url-builder.ts
```diff
@@ -0,0 +1,11 @@
+import config from './config'
+import {last, first} from 'lodash'
+
+export const buildTitleFromUrl = (url) => {
```
joelhooks, line 4: Just an initial poke at this idea. It's a giant of statement at the end of the day feels like?

johnlindquist: Was it important to keep versions of titles or only URLs? Probably wanna version maps of titles/urls then add a version arg:

`buildTitleFromUrl = (url, version) => `

joelhooks: It only matters for the URLs as far as search engines are concerned. 

Titles are pretty mutable. 

The URL builder is from search filter/facet set

PR #20, opened 2020-08-05
stubbing out all the pages in the nav

johnlindquist:
This is mainly to fix 404s in the nav.

I also played around with /lessons and /courses. I copy/pasted a tailwind card layout from [here](https://tailwindcss.com/components/cards/#app) and filled them in with data from graphql. I imagine these will be deleted or refactored, but it was cool to see how far you could get with a tailwind component and graphql data.

I spent some time with types in courses and lessons, but back out those changes because that will be a bigger discussion.


lib/lessons.ts
```diff
@@ -1,7 +1,26 @@
 import {request} from 'graphql-request'
 import config from './config'
 
-export async function loadLesson(slug) {
+export async function loadLessons() {
+  const query = /* GraphQL */ `
+    query getLessons {
+      lessons(per_page: 25) {
```
johnlindquist, line 7: This is a placeholder. I'd like to explore the pagination with that new Next.js hook you found.

pages/lessons/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+            <a>{lesson.title}</a>
+          </div>
+        </Link>
+        <p className="text-gray-700 text-base truncate">{lesson.description}</p>
+      </div>
+    </div>
+  )
+}
+
+export default function Courses({lessons}) {
```
joelhooks, line 20:
```suggestion
export default function Lessons({lessons}) {
```

joelhooks: On egghead.io as it is now we dropped /lessons in favor of /search with a filter, and I think this is a good move. 5000+ lessons in a paginated list aren't very structurally predisposed to learning. 😂

pages/courses/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+          {/* <p className="text-gray-700 text-base truncate">
+              {course.description}
+            </p> */}
+        </div>
+      </a>
+    </Link>
+  )
+}
+
+export default function Courses({allCourses}) {
```
joelhooks, line 28: /courses also forwards to /search on egghead.io for similar reasons as /lessons. The paginated list isn't very useful in practice.

I can see an argument for a curated /courses page with on-page filtered search.

lib/lessons.ts
joelhooks: We actively killed the /lessons route in favor of filtered search as a paginated list of 5000+ videos is...

https://twitter.com/chrisbiscardi/status/1172527924945096706?s=20

They capture it in this thread well.

PR #24, opened 2020-08-14
less relative paths for components and such thanks

joelhooks:


package.json
```diff
@@ -68,18 +72,47 @@
...
     "postcss-preset-env": "^6.7.0",
+    "prettier": "^2.0.5",
     "react-test-renderer": "^16.13.1",
     "start-server-and-test": "^1.11.2",
     "tailwindcss": "^1.4.0",
     "typescript": "^3.9.7"
   },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
```
joelhooks, line 92: very consistent formatting!

tsconfig.json
```diff
@@ -13,7 +13,16 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "jsx": "preserve",
-    "noImplicitAny": false
+    "noImplicitAny": false,
+    "baseUrl": "./",
+    "paths": {
+      "@components/*": ["src/components/*"],
```
joelhooks, line 19: this is a convenience method to avoid all the relative imports from `pages`

joelhooks: also moved the code stuff that isn't pages into `src` because the top-level directory was gonna get heinous

johnlindquist: Just noting that the docs use:

```js
"@/components/*": ["components/*"]
```

I'm 99% sure it's only a preference.

joelhooks: I was copying what Dom had done. You could make it “butts” and it would still work so it’s definitely a pref.

johnlindquist: +1 for "butts"

PR #26, opened 2020-08-14
Jh/react page collections

joelhooks:


src/components/app/Main.tsx
```diff
@@ -1,6 +1,6 @@
 export default function Main({children}) {
   return (
-    <main className="max-w-screen-lg w-full mx-auto sm:p-8 p-5 flex flex-col flex-grow">
+    <main className="max-w-screen-lg w-full mx-auto sm:p-8 p-5 flex flex-col flex-grow prose">
```
joelhooks, line 3: was trying to figure out how to format the mdx above so I added the `prose` here, but I don't think this is a great way to do that and I'm wondering how better to manage/handle mdx 🤔

src/lib/collections.ts
```diff
@@ -0,0 +1,9 @@
+import axios from 'axios'
+
+export async function loadCollection(slug: string) {
+  const endpoint = `${process.env.NEXT_PUBLIC_AUTH_DOMAIN}/api/v1/playlists/${slug}`
```
joelhooks, line 4: playlists don't have a graphql endpoint yet

src/components/mdx/course.tsx
```diff
@@ -0,0 +1,22 @@
+import useSWR from 'swr'
+import {loadCourse} from '@lib/courses'
+import Link from 'next/link'
+
+export default ({slug}) => {
+  const {data} = useSWR(slug, loadCourse)
```
joelhooks, line 6: this just works so well 😍

dunno how it handles it from the SSR perspective, but widgets don't really matter and can load lazy/async as far as SEO or whatever is concerned.

pages/learn/react/index.tsx
```diff
@@ -0,0 +1,16 @@
+import Head from 'next/head'
+
+import LearnReactContent from '@content/learn/react/index.mdx'
+
+export default function LearnReact() {
```
joelhooks, line 5: instead of a raw mdx here we can use a React component and inject the mdx. That gives more control of the template and layout.

tailwind.config.js
```diff
@@ -1,9 +1,9 @@
 module.exports = {
   purge: [
-    './components/**/*.js',
+    './src/**/*.js',
```
joelhooks, line 3: whoops

PR #29, opened 2020-08-17
semantic search urls

joelhooks:
This is exploring some of the ideas around "semantic URLs" for site search that are friendly to google.

Basically taking the URL and parsing into Algolia params and then back again.


src/lib/search-url-builder.ts
```diff
@@ -9,3 +10,71 @@ export const buildTitleFromUrl = (url: string) => {
...
+}
+
+// Returns a name from the category slug.
+// The "+" are replaced by spaces and other
+// characters are decoded.
+function getCategoryName(slug) {
+  return slug.split('+').map(decodeURIComponent).join(' ')
+}
+
+export const createUrl = (searchState) => {
```
joelhooks, line 25: all of this gets a little gross, but it has tests ;)

src/pages/search/[[...all]].tsx
```diff
@@ -35,17 +35,15 @@ export default function SearchIndex({initialSearchState, resultsState}) {
...
+      const href = searchStateToUrl(searchState)
 
       //this is all f'd up. The general idea is to build up SEO friendly URLs for search
       //where the url would be broken up like `/react/hooks/courses/by/kent+c+dodds` which is 100%
       //possible but also fairly nuanced and complex 😅
 
-      router.push(href, href, {
-        shallow: true,
-      })
+      // router.push(href, href, {
```
Comment (joelhooks) [inline src/pages/search/[[...all]].tsx:44]: killed this for testing purposes

jaredpalmer:
My suggestion is to kill all the slashes and use a different delimiter. The goal is just and English phrase slugified

jaredpalmer:
My suggestion is to kill all the slashes and use a different delimiter. The goal is just and English phrase slugified

joelhooks:
@jaredpalmer I used `~~t~~` instead, where `~~` is a replacement for the `/`

Is that any better? we've got `type`, `tag`, `query`, and `instructor` currently as the facets...

joelhooks:
actually I like `--` better A E S T H E T I C A L L Y

jaredpalmer:
The closer to English the better. 

react-and-amplify
videos-by-dan-abramov
react-and-redux-videos-by-kent-c-dodds
react-hooks-and-redux-and-netlify

src/pages/s/[[...all]].tsx
```diff
@@ -56,17 +54,18 @@ export default function SearchIndex({initialSearchState, resultsState}) {
     createURL,
     onSearchStateChange,
   }
+
   return (
     <div>
+      <NextSeo noindex={!isEmpty(searchState.query)} title={pageTitle} />
```
Comment (joelhooks) [inline src/pages/s/[[...all]].tsx:60]: if the user has typed in a query we don't want to index anything. We only index when they have applied certain refinements and no query.

PR #33, opened 2020-08-21
Generating a sitemap

johnlindquist:
[nextjs-sitemap-generator](https://github.com/IlusionDev/nextjs-sitemap-generator)

I saw this package recommended a few times. We should be able to tweak it to match any pages we generate.

I imagine we would want to integrate it with your search-url-builder?


johnlindquist:
The `pagesDirectory` option has been a pain: https://github.com/IlusionDev/nextjs-sitemap-generator/issues/40#issuecomment-659584790

If we want to build on vercel, we have to make sure we know what their internals are doing during the build. So I think this has the potential of breaking based on versions #s and vercel internals changes.

I'm still investigating. Just leaving this note for future reference.

johnlindquist:
@joelhooks 

`robots.txt` can support multiple sitemaps, so here's my suggestion:
1. Ruby continues to generate a sitemap based on the database
2. Node generates a sitemap based on Next.js pages

Our robots.txt would then look like:

```
User-agent: *
Allow: /
Host: https://next.egghead.io
Sitemap: https://next.egghead.io/sitemap.xml
Sitemap: http://egghead-sitemaps.s3.amazonaws.com/sitemaps/sitemap.xml.gz
```

I really like this approach. Do you see any downsides?

joelhooks:
@johnlindquist nope, that's excellent. Best of both worlds.

Generating `/s` slugs can probably focus on the top 100-200 combinations and build from there. I don't know realistically what the limit would be 🤔

johnlindquist:
Let's do a third sitemap specifically for /s urls

joelhooks:
sure, that makes sense

joelhooks:
Jared suggested a combinatorics library to generate the sitemaps for search. There’s a limit of 50,000 entries but no limit to the number of sitemaps. I suspect the possible combos is u realistic though 😂

joelhooks:
Jared suggested a combinatorics library to generate the sitemaps for search. There’s a limit of 50,000 entries but no limit to the number of sitemaps. I suspect the possible combos is u realistic though 😂

johnlindquist:
There's an option to generate a `robots.txt`, but I didn't want to touch it until we're ready to go live.

https://www.npmjs.com/package/next-sitemap#full-configuration
Screen Shot 2020-08-26 at 1 02 24 PM

So the "pages" sitemap is done, I'll start investigating the /s sitemap.

joelhooks (approved):

PR #35, opened 2020-08-26
enables analytics via ahoy and video progress tracking

joelhooks:
This enables analytics via ahoy and migrates video progress and end event handlers.

https://github.com/eggheadio/egghead-rails/pull/4058 <- this needs to go first as this requires a couple graphql API additions to function properly


src/components/EggheadPlayer/index.js
```diff
@@ -5,9 +5,10 @@ import {propTypes, defaultProps} from './props'
 import Wistia from './players/Wistia'
 import Bitmovin from './players/Bitmovin'
 import YouTube from './players/YouTube'
+import useEggheadPlayer from './use-egghead-player'
 
-export default class ReactPlayer extends Component {
-  static displayName = 'ReactPlayer'
+export default class EggheadPlayer extends Component {
```
joelhooks, line 10: renamed, this could use a refactor and extraction some time into a lib

src/components/EggheadPlayer/use-egghead-player.ts
```diff
@@ -0,0 +1,213 @@
...
+  })
+}
+
+function onError() {}
+
+function onVideoQualityChanged() {}
+
+function onSubtitleChange() {}
+
+export default function useEggheadPlayer(lesson) {
```
joelhooks, line 205: this isn't actually a hook, but the pattern makes sense so I went with it.

src/components/EggheadPlayer/use-egghead-player.ts
```diff
@@ -0,0 +1,213 @@
...
+  })
+
+const onComplete = (lesson, collection?) => {
+  return getOrCreateLessonView(lesson, collection)
+    .then(trackStartingCourse)
+    .then(completeLesson)
+    .then(trackPercentComplete)
+}
+
+const onEnded = (lesson) => async () => {
```
joelhooks, line 183: this was almost a full port. in the egghead-rails version, it was managing some view state as well which is omitted here.

src/components/EggheadPlayer/use-egghead-player.ts
```diff
@@ -0,0 +1,213 @@
...
+const storeProgress = (playedSeconds, lesson, list?) => {
+  return new Promise((resolve) => {
+    const isSegment = Math.round(playedSeconds) % 30 === 0
+    if (isSegment) {
+      createLessonView(lesson, list).then(incrementViewSegments).then(resolve)
+    }
+  })
+}
+
+const onProgress = (lesson) => (progress) => {
```
joelhooks, line 64: logically this ported over mostly intact. It's missing some details around playlists (collections) that is used primarily to decide what to show in the "next up list" but also as metadata to analytics.

src/pages/api/events.ts
```diff
@@ -0,0 +1,17 @@
+import axios from 'axios'
+
+export default async (req, res) => {
```
joelhooks, line 3: using lambdas solves for a host of CORS issues

johnlindquist: I ❤️ vercel's approach to this. So simple.

src/pages/_app.tsx
```diff
@@ -9,6 +9,15 @@ import mdxComponents from '@components/mdx'
 import defaultSeoConfig from 'src/next-seo.json'
 import {useURL} from '@hooks/useUrl'
 import '../styles/index.css'
+import Router from 'next/router'
+
+declare global {
+  interface Window {
+    ahoy: any
```
johnlindquist, line 16: I love modern development. We try so hard to hide from window, but the hacks to bring it back in just keep getting more interesting.

johnlindquist:
> renamed, this could use a refactor and extraction some time into a lib 

This would be a good one to pair on

PR #37, opened 2020-08-29
😭 this enables the TS restriction for no implicit any

joelhooks:
now we have to type stuff more aggressively and let our friends know when it's OK to party with `any`. Over time we will be able to add better types in general and not be so loose with the goose.

https://twitter.com/jaredpalmer/status/1299702784565149698?s=20

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -13,7 +13,6 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "jsx": "preserve",
-    "noImplicitAny": false,
     "baseUrl": "./",
     "paths": {
       "@components/*": ["src/components/*"],
--- a/@types/browser-cookies.d.ts
+++ b/@types/browser-cookies.d.ts
@@ -1,0 +1,1 @@
+declare module 'browser-cookies'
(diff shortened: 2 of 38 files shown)


tsconfig.json
```diff
@@ -13,7 +13,6 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "jsx": "preserve",
-    "noImplicitAny": false,
```
joelhooks, line 16: removing this opens a large can of worms

johnlindquist:
We'll need to restart the conversation about Types from Data/GraphQL to get out of `any`-land. The Types as they are now are just begging for it.

joelhooks:
I’d vote for hand crafting a couple of these types in the short term

mainly an interface around Resources

https://github.com/eggheadio/egghead-rails/blob/a89ee004898e11f9b6fcb11ec33217b217650e2d/app/models/concerns/content_model.rb

the ContentModel is effectively the api

https://github.com/eggheadio/egghead-rails/blob/a89ee004898e11f9b6fcb11ec33217b217650e2d/db/migrate/20191115184135_create_resources_file.rb (just found an example of the core props)

PR #39, opened 2020-08-29
Jh/remove one any

joelhooks:
[previous PR was merged rapidly]


src/types.ts
```diff
@@ -0,0 +1,14 @@
+export type Resource = {
```
joelhooks, line 1: type makes more sense if we aren't using classes I think?

src/pages/lessons/index.tsx
```diff
@@ -1,52 +0,0 @@
-import Link from 'next/link'
```
joelhooks, line 1: these are actually hard forwarded to `/search` so never displayed

src/types.ts
johnlindquist: Yes. `Interface` is strictly for OOP patterns IMO.

now.json
```diff
@@ -1,6 +1,6 @@
 {
   "redirects": [
-    {"source": "/lessons", "destination": "/s"},
-    {"source": "/courses", "destination": "/s"}
+    {"source": "/lessons", "destination": "/s?type=lesson"},
```
johnlindquist, line 3: This is starting to remind me of how codesandbox does links:

http://codesandbox.io/s/vanilla
http://codesandbox.io/s/github
http://codesandbox.io/s/vue

etc, etc, etc.

PR #41, opened 2020-09-01
cleaning up the builder

johnlindquist:
Took a pass at cleaning up the builder. All the tests pass, but I think we need more tests around tags and years. @joelhooks we could look at this together or maybe you could write some more tests and I could implement?


src/lib/search-url-builder.ts
```diff
@@ -47,18 +49,27 @@ export const titleFromPath = (all: string[] = []) => {
...
-    )} in ${year}`
-  } else if (tags) {
-    return `${config.searchResultCount}${
-      tags ? ` ${humanize(tags.map(toTitleCase))} ` : ' '
-    }Courses for Web Developers in ${year}`
+    return `${count}${humanizedTags}Courses from ${humanizedInstructors} in ${year}`
   }
-  return `${config.searchResultCount} Courses for Web Developers`
+
+  //TODO: I think we need more tests around tags and years here...
```
joelhooks, line 62: surprised to not see any added if this is the case! The tests were definitely non-exhaustive.

joelhooks (approved):

PR #42, opened 2020-09-03
Using "absolute imports" instead of "aliases"

johnlindquist:
Next.js 9.4 added support for absolute imports based on the `baseUrl`:

https://nextjs.org/blog/next-9-4#absolute-imports-and-aliases
https://nextjs.org/docs/advanced-features/module-path-aliases

This removes the need for `@` paths on every directory.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -13,16 +13,7 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "jsx": "preserve",
-    "baseUrl": "./",
-    "paths": {
-      "@components/*": ["src/components/*"],
-      "@utils/*": ["src/utils/*"],
-      "@hooks/*": ["src/hooks/*"],
-      "@context/*": ["src/context/*"],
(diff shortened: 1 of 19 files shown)


tsconfig.json
```diff
@@ -13,16 +13,7 @@
...
-    "paths": {
-      "@components/*": ["src/components/*"],
-      "@utils/*": ["src/utils/*"],
-      "@hooks/*": ["src/hooks/*"],
-      "@context/*": ["src/context/*"],
-      "@lib/*": ["src/lib/*"],
-      "@content/*": ["content/*"],
-      "@types": ["src/types.ts"]
-    }
+    "baseUrl": "./src/"
```
johnlindquist, line 16: Pointing `baseUrl` to `src` allows us to delete all of the `paths` config.

All of the other file changes are removing the `@` from the beginning of the module import.

joelhooks (approved):

PR #43, opened 2020-09-08
add a basic machine

joelhooks:

### DIFF
--- a/src/components/EggheadPlayer/machine.ts
+++ b/src/components/EggheadPlayer/machine.ts
@@ -1,0 +1,63 @@
+import {Machine} from 'xstate'
+
+interface PlayerStateSchema {
+  states: {
+    loading: {}
+    playing: {}
+    quizzing: {}
+    rating: {}
+    subscribing: {}
+    joining: {}
+    completed: {}
+  }
+}
+
+export type PlayerStateEvent =
+  | {type: 'PLAY'}
+  | {type: 'SUBSCRIBE'}
+  | {type: 'JOIN'}
+  | {type: 'QUIZ'}
+  | {type: 'COMPLETE'}
+  | {type: 'RATE'}
+
+const playerMachine = Machine<PlayerStateSchema, PlayerStateEvent>({
+  id: 'player',
+  initial: 'loading',
+  states: {
+    loading: {
+      on: {
+        PLAY: 'playing',
+        SUBSCRIBE: 'subscribing',
+        JOIN: 'joining',
+      },
+    },
+    playing: {
+      on: {
+        QUIZ: 'quizzing',
+        COMPLETE: 'completed',
+        RATE: 'rating',
+      },
+    },
+    quizzing: {
+      on: {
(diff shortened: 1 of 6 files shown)


src/components/EggheadPlayer/machine.ts
```diff
@@ -0,0 +1,63 @@
+import {Machine} from 'xstate'
```
joelhooks, line 1: xstate and typescript requires some love: https://xstate.js.org/docs/guides/typescript.html

src/pages/lessons/[slug].tsx
```diff
@@ -109,6 +112,13 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
 
   if (!lesson) return null
 
+  React.useEffect(() => {
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:115]: if it isn't in a useEffect it cries in the console.

src/pages/lessons/[slug].tsx
```diff
@@ -109,6 +112,13 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
...
+    if (playerState.value === 'loading') {
+      // here we can do some consideration for the actual state.
+      send('PLAY')
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:118]: I just defaulted to play here, but you can do all sorts of fun algorithms here to figure out the state based on data we have if it is pro or whatever to pop a CTA

src/pages/login.tsx
```diff
@@ -11,33 +11,32 @@ const loginSchema = yup.object().shape({
 })
 
 function LoginForm() {
-  const [clicked, setClicked] = React.useState(false)
   const [isSubmitted, setIsSubmitted] = React.useState(false)
+  const [isError, setIsError] = React.useState(false)
```
joelhooks, line 15: I fixed this on Epic React today and migrated it over so it just happened to be in this same PR.

src/pages/lessons/[slug].tsx
Comment (johnlindquist) [reply src/pages/lessons/[slug].tsx:118]: Very cool. Should be a fun puzzle.

src/components/EggheadPlayer/machine.ts
```diff
@@ -0,0 +1,63 @@
...
+
+export type PlayerStateEvent =
+  | {type: 'PLAY'}
+  | {type: 'SUBSCRIBE'}
+  | {type: 'JOIN'}
+  | {type: 'QUIZ'}
+  | {type: 'COMPLETE'}
+  | {type: 'RATE'}
+
+const playerMachine = Machine<PlayerStateSchema, PlayerStateEvent>({
```
johnlindquist, line 23: I bet you enjoyed writing this

src/pages/lessons/[slug].tsx
Comment (johnlindquist) [reply src/pages/lessons/[slug].tsx:115]: I need to dive into the react/xstate/hooks demos

johnlindquist (approved):

PR #44, opened 2020-09-09
Adding a "CommandBar" to issue dev commands

johnlindquist:
Running `yarn dev:tools` dynamically places an app-wide input box. 

TODO: Map commands that the input box can invoke to control the app.

Screen Shot 2020-09-09 at 9 10 55 AM

### DIFF
--- a/src/components/app/Layout.tsx
+++ b/src/components/app/Layout.tsx
@@ -1,12 +1,19 @@
 import {FunctionComponent} from 'react'
+import dynamic from 'next/dynamic'
 import Header from './Header'
 import Footer from './Footer'
 import Main from './Main'
 
+let CommandBar: FunctionComponent = () => <></>
+
+if (process.env.NODE_ENV == 'development' && process.env.NEXT_PUBLIC_DEVTOOLS) {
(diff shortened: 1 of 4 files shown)


src/components/app/Layout.tsx
```diff
@@ -1,12 +1,19 @@
...
+let CommandBar: FunctionComponent = () => <></>
+
+if (process.env.NODE_ENV == 'development' && process.env.NEXT_PUBLIC_DEVTOOLS) {
+  CommandBar = dynamic(() => import('./CommandBar')) as FunctionComponent
+}
 const Layout: FunctionComponent = ({children}) => {
   return (
     <>
       <Header></Header>
+      <CommandBar></CommandBar>
```
johnlindquist, line 16: This can go anywhere, but it will need access to a Context/whatever to `send` commands to.

We could also show/hide with a keyboard shortcut if we want to develop this further.

johnlindquist: We could also conditionally render like `{devtools && <CommandBar/>}`

joelhooks:
What's the scoop on this? Would be nice to see some utility here! Or should it be closed for now?

PR #81, opened 2020-09-14
add tags to Metadata component

DrShpongle:
Image 2020-09-14 at 3 48 33 PM

### DIFF
--- a/src/pages/lessons/[slug].tsx
+++ b/src/pages/lessons/[slug].tsx
@@ -4,7 +4,7 @@ import React, {FunctionComponent} from 'react'
 import Link from 'next/link'
 import {useRouter} from 'next/router'
 import EggheadPlayer from 'components/EggheadPlayer'
-import get from 'lodash/get'
+import {isEmpty, get} from 'lodash'
 import Markdown from 'react-markdown'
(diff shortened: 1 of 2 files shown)


tailwind.config.js
```diff
@@ -15,6 +15,8 @@ module.exports = {
     }),
     extend: {},
   },
-  variants: {},
+  variants: {
+    margin: ['responsive', 'first', 'last'],
```
DrShpongle, line 19: enable `:first` and `:last` pseudoclasses support

src/pages/lessons/[slug].tsx
```diff
@@ -4,7 +4,7 @@ import React, {FunctionComponent} from 'react'
 import Link from 'next/link'
 import {useRouter} from 'next/router'
 import EggheadPlayer from 'components/EggheadPlayer'
-import get from 'lodash/get'
+import {isEmpty, get} from 'lodash'
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:7]: I was curious which way is “better” for Lodash imports and it turns out the way you’ve used is recommended

https://lodash.com/per-method-packages

joelhooks (approved):

PR #83, opened 2020-09-15
Metadata component improvements

DrShpongle:
TypeScript workaround for rest props due to https://stackoverflow.com/questions/40032592/typescript-workaround-for-rest-props-in-react


DrShpongle:
offtop: can we add some license field to the `package.json` so that it will be possible to commit with VS Code (seems it checks `licence` field presence)?

Image 2020-09-15 at 10 42 25 AM

src/pages/lessons/[slug].tsx
```diff
@@ -99,16 +99,18 @@ type MetadataProps = {
...
 const Metadata: FunctionComponent<MetadataProps> = ({
   title,
   instructor,
   tags,
   summary,
+  ...restProps
 }: MetadataProps) => {
   return (
-    <div>
+    <div {...restProps}>
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:113]: Don’t know if this applies here but it was interesting https://twitter.com/markdalgleish/status/1305714449475489792?s=20

joelhooks (approved):

PR #86, opened 2020-09-15
ULTRAwide

joelhooks:
I wanted to make two proposals with this PR:

* an ultrawide breakpoint
* moving the "next up" list way down the page

I think that a lot of folks learn on ultra wides and the XL breakpoint of tailwind is super narrow.

image

The "course info" top bar should provide an opportunity for orientation and navigation in a list and it makes the full list of lessons less interesting. Additionally when a lesson completes we will display what the next lesson is, again, making the full lesson list not very interesting IMO.

Labeled RFC for discussion, but feel free to merge to continue work.


tailwind.config.js
```diff
@@ -6,6 +6,13 @@ module.exports = {
     './src/**/*.tsx',
   ],
   theme: {
+    screens: {
```
joelhooks, line 9: tailwind lets you set this arbitrarily, but 1900 felt pretty good on my 32" screen

DrShpongle:
User should have ability to view the full list of lessons (in case he wants to skip some lesson or evaluate amount of lessons). So I think we should keep it but probably make with maxHeight and/or scrollable/collapsable.
Related question - do we plan to have the same layout options for the theater mode as current egghead has (when lessons list jumps to the right of the video in non-theater mode )?

joelhooks:
I disagree with the assumption that having a list of lessons for review is something that most learners do, it’s just what we’ve always done. 

These are created and presented in a linear format, consumed in a linear way, and skipping around is not efficient or effective for learning. 

I think “theater mode” should be the only mode and that the extended list of lessons should migrate to well below the fold. 

This will allow the video to dominate and also allow us to have more consistent screen real estate for quizzes and other CTAs

Additionally the weird sequence with the list will no longer be something you have to scroll through and we remove the need for “inner scroll”

This assumes the top bar has “next lesson”, lesson count, current position, etc visible as represented in Vojta’s latest wireframes

DrShpongle:
Only "theater" mode makes sense!
Only the thing I want to suggest is probably PrevLesson button near the NextUp? Once we consider the top bar as main navigation control. I think sometimes user wants just review the prev lesson in case he lose the track in current lesson (my situation sometimes 😅)

joelhooks:
From the designs I think “next” would work on the “list bar” above the video. 

Previous is less useful perhaps?

joelhooks:
Also the full list will be available on screen. I think since it can be displayed at full length it is more useful too for see the entire collection of videos/content

DrShpongle:
Ah, I thought that your idea is to cut the list of lessons off. Once we keep it anyway we don't need Prev button for sure. Misunderstood.

DrShpongle (approved):

tailwind.config.js
```diff
@@ -6,6 +6,13 @@ module.exports = {
...
+      sm: '640px',
+      md: '768px',
+      lg: '1024px',
+      xl: '1280px',
+      ultra: '1900px',
```
vojtaholik, line 14: that's quite a jump from xl to ultra. 
how about we add one more and apply more consistent naming (in tailwind world) and make use of traditionally recognized resolutions?

suggestion:
```
xl: 1280
2xl: 1440
3xl: 1920
```

PR #89, opened 2020-09-17
Lesson page refactoring && player controls component

DrShpongle:
Image 2020-09-17 at 3 08 13 PM


src/pages/_app.tsx
```diff
@@ -9,6 +9,7 @@ import mdxComponents from 'components/mdx'
 import defaultSeoConfig from 'next-seo.json'
 import {useURL} from 'hooks/useUrl'
 import '../styles/index.css'
+import '@reach/listbox/styles.css'
```
DrShpongle, line 12: If I'm not mistaken NextJS should allow import css from node_modules on the component level, but I got this error:
Image 2020-09-17 at 3 15 56 PM
So I added css import to the `_app.tsx` for now

src/pages/lessons/[slug].tsx
```diff
@@ -49,44 +50,42 @@ const lessonQuery = /* GraphQL */ `
 
 const fetcher = (url: RequestInfo) => fetch(url).then((r) => r.json())
 
+const getNextUpData = (url: string) => {
+  const {data} = useSWR(url, fetcher)
+  return data
+}
+
```
Comment (DrShpongle) [inline src/pages/lessons/[slug].tsx:57]: Make this request as separate function so that we can get data with one request and use it for few places (there are two cases at least - `data.list.lessons` for the `NextUp` component and `data.next_lesson` for the `PlayerControls` component)

src/pages/lessons/[slug].tsx
```diff
@@ -102,74 +101,6 @@ const Transcript: FunctionComponent<TranscriptProps> = ({
...
-              </a>
-            ))}
-          </div>
-        )}
-      </div>
-      {summary && <Markdown className="mt-4">{summary}</Markdown>}
-    </div>
-  )
-}
-
```
Comment (DrShpongle) [inline src/pages/lessons/[slug].tsx:172]: Moved `Metadata` component to the separate file

src/pages/lessons/[slug].tsx
```diff
@@ -1,20 +1,21 @@
...
+import Metadata from 'components/Metadata'
 import {loadLesson} from 'lib/lessons'
-import {GraphQLClient} from 'graphql-request'
 import {useViewer} from 'context/viewer-context'
-import {GetServerSideProps} from 'next'
-import {LessonResource} from 'types'
-import {useMachine} from '@xstate/react'
-import playerMachine from 'components/EggheadPlayer/machine'
-import Eggo from '../../../public/images/eggo.svg'
+import * as Types from 'types'
```
Comment (DrShpongle) [inline src/pages/lessons/[slug].tsx:18]: As a suggestion: what if we import types in this way `import * as Types from 'types'` and use it as `Types.Lesson` for example? Just to avoid intersection between `Lesson` type and `Lesson` component? Just a matter of naming convention

src/pages/lessons/[slug].tsx
```diff
@@ -186,7 +117,7 @@ const lessonLoader = (slug: any, token: any) => (query: string) => {
 }
 
 type LessonProps = {
-  initialLesson: LessonResource
+  initialLesson: Types.LessonResource
 }
```
Comment (DrShpongle) [inline src/pages/lessons/[slug].tsx:121]: due to suggestion from the comment above about types import and naming convention

src/pages/lessons/[slug].tsx
```diff
@@ -49,44 +50,42 @@ const lessonQuery = /* GraphQL */ `
...
+const getNextUpData = (url: string) => {
+  const {data} = useSWR(url, fetcher)
+  return data
+}
+
 type NextUpProps = {
-  url: string
+  data: {
+    list: {
+      lessons: Types.Lesson[]
```
Comment (DrShpongle) [inline src/pages/lessons/[slug].tsx:61]: Created the `Lesson` type and stored it in `types.ts` - I think we will use this type on other pages also?

src/components/PlayerControls.tsx
```diff
@@ -0,0 +1,77 @@
...
+              <ListboxPopover>
+                <ListboxList>
+                  {availableSpeeds.map((option) => (
+                    <ListboxOption key={option} value={'' + option}>
+                      {`x${option}`}
+                    </ListboxOption>
+                  ))}
+                </ListboxList>
+              </ListboxPopover>
+            </ListboxInput> */}
```
DrShpongle, line 61: I've lost in `@reach/listbox` API, not sure how to handle controlled listbox.

DrShpongle:
Few more notions:
1) I noticed that data, gotten with `const {data} = useSWR(url, fetcher)` doesn't get updated when we switch the lessons in `NextUp` component, it gets updated after page refresh only. So it's impossible to get actual `next_lesson_url`. Because of SWR memoization? Not sure.
2) `next-svgr` breakes some icons (from https://heroicons.dev/?search=check&query=download at least):
Image 2020-09-17 at 3 55 40 PM
it supposed to be:
Image 2020-09-17 at 3 56 22 PM

src/pages/_app.tsx
joelhooks: Ya it doesn’t allow global style import at the component t level.

src/pages/lessons/[slug].tsx
Comment (joelhooks) [reply src/pages/lessons/[slug].tsx:57]: I think this should probably be named like a React Hook useCollectionData, but looks good really. That’s just a convention

src/pages/lessons/[slug].tsx
Comment (joelhooks) [reply src/pages/lessons/[slug].tsx:18]: I prefer to import them Individually and keep the code below less  wrbose and having a few extra lines of imports

src/pages/lessons/[slug].tsx
Comment (joelhooks) [reply src/pages/lessons/[slug].tsx:61]: I would prefer not to have a lesson type currently or consider a more generic resource type. We’ve discussed generating these types from our GraphQL schema on the egghead side but haven’t come to a resolution

Probably fine for now, but that’s why there wasn’t a Lesson type.

src/pages/lessons/[slug].tsx
Comment (DrShpongle) [reply src/pages/lessons/[slug].tsx:121]: Got it, will turn it back

src/components/PlayerControls.tsx
```diff
@@ -29,7 +29,7 @@ const PlayerControls: FunctionComponent<PlayerControlsProps> = ({
   isPro,
 }: PlayerControlsProps) => {
   console.log('nextLessonUrl', nextLessonUrl)
-  const [speed, setSpeed] = useState(1)
+  const [speed, setSpeed] = useState<string>('1')
```
joelhooks, line 32: the React hooks use some format of generic types and this has been my approach. If you use a custom style it would go here.

src/components/PlayerControls.tsx
```diff
@@ -45,20 +45,20 @@ const PlayerControls: FunctionComponent<PlayerControlsProps> = ({
             )}
           </div>
           <div className="flex items-center">
-            {/* <ListboxInput value={'' + speed} onChange={setSpeed}>
+            <ListboxInput value={speed} onChange={setSpeed}>
```
joelhooks, line 48: I typed everything above instead of this "casting" hack. I'd recommend using `speed as string` in this circumstance for casting

https://basarat.gitbook.io/typescript/type-system/type-assertion

src/components/PlayerControls.tsx
joelhooks: https://fettblog.eu/typescript-react/hooks/ this is a great bookmark

src/components/PlayerControls.tsx
```diff
@@ -45,20 +45,20 @@ const PlayerControls: FunctionComponent<PlayerControlsProps> = ({
...
               <ListboxButton className="w-20 bg-gray-300 rounded">
                 <IconChevronRight className="w-4" /> x{speed}
               </ListboxButton>
               <ListboxPopover>
                 <ListboxList>
                   {availableSpeeds.map((option) => (
-                    <ListboxOption key={option} value={'' + option}>
+                    <ListboxOption key={option} value={option}>
```
joelhooks, line 55: this appears to work as expected

I think we are going to move this control back into the player. Talking to @vojtaholik about it.

src/types.ts
```diff
@@ -12,3 +12,10 @@ export type LessonResource = Resource & {
   id: string | number
   tags: any[]
 }
+
+export type Lesson = {
```
joelhooks, line 16: We will chat about this, and what gets added to types in general. You can tell from the signature below that something isn't coded properly because of its complexity.

DrShpongle:
closed in favor of https://github.com/eggheadio/egghead-next/pull/90

PR #90, opened 2020-09-17
Jh/player state machine

joelhooks:


src/pages/lessons/[slug].tsx
```diff
@@ -49,44 +50,42 @@ const lessonQuery = /* GraphQL */ `
...
+            <div className="flex">
+              <div>
+                {index + 1}{' '}
+                <input type="checkbox" checked={lesson.completed} readOnly />
               </div>
-            </li>
-          )
-        },
-      )}
+              <Link href={`/lessons/[id]`} as={lesson.path}>
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:80]: @DrShpongle note this format for `Link` when linking to internal pages!

src/pages/lessons/[slug].tsx
```diff
@@ -182,7 +113,22 @@ const lessonLoader = (slug: any, token: any) => (query: string) => {
       ...authorizationHeader,
     },
   })
-  return graphQLClient.request(query, variables)
+  return graphQLClient.request(lessonQuery, variables)
+}
+
+const NextResourceButton: FunctionComponent<{
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:119]: I made this component as a small example of composition to consider. If the component has children, they are delivered and displayed, if not is uses default text `Next Lesson`

src/pages/lessons/[slug].tsx
```diff
@@ -195,28 +141,22 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
   const {authToken, logout} = useViewer()
   const [playerState, send] = useMachine(playerMachine)
 
+  console.log({playerState})
+
   const {data = {}, error} = useSWR(
-    lessonQuery,
-    lessonLoader(initialLesson.slug, authToken),
+    [initialLesson.slug, authToken],
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:147]: this was the refactor we did on the screen.

`useSWR` first argument is the **key** so the query was causing this to fail. We want to reload for a new lesson and/or authToken

src/pages/lessons/[slug].tsx
```diff
@@ -228,10 +168,39 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
...
+          send('PLAY')
+        } else {
+          send('SUBSCRIBE')
+        }
+      case 'completed':
+        send('NEXT')
+      default:
+        break
+    }
+  }, [currentState, data.lesson])
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:193]: if the current state changes OR the current lesson changes, re-run this effect.

src/pages/lessons/[slug].tsx
```diff
@@ -228,10 +168,39 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
...
 
-  console.log(lesson)
+  const currentState = playerState.value
+
+  console.log(`The current player state: ${currentState}`)
+
+  React.useEffect(() => {
+    switch (currentState) {
+      case 'loading':
+        if (!isEmpty(data.lesson)) {
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:178]: once we have `data.lesson` we are done loading and can do stuff

src/pages/lessons/[slug].tsx
```diff
@@ -228,10 +168,39 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
...
+          send('LOADED')
+        }
+        break
+      case 'loaded':
+        if (hls_url || dash_url) {
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:183]: if we have urls we can display the video player

note: we are missing the "ask user for email" state here that would check to see how many free lessons they have watched and push them to create an account to continue.

src/pages/lessons/[slug].tsx
```diff
@@ -228,10 +168,39 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
...
+          send('PLAY')
+        } else {
+          send('SUBSCRIBE')
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:186]: if those URLs are NOT present they need to pay up!

src/pages/lessons/[slug].tsx
```diff
@@ -228,10 +168,39 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
...
+        }
+      case 'completed':
+        send('NEXT')
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:189]: when the lesson finishes we decide what overlay to display. This would also be where we would transition to a quiz, review, etc when a video completes. Currently, it's just the next lesson overlay.

src/pages/lessons/[slug].tsx
```diff
@@ -228,10 +168,39 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
...
+  const nextUpData = useNextUpData(next_up_url)
+  const nextUpPath = get(nextUpData, 'next_lesson')
+  const playerVisible: boolean =
+    playerState.value === 'playing' || playerState.value === 'paused'
+
+  console.log(nextUpPath)
 
   return (
-    <div className="max-w-none">
+    <div className="max-w-none" key={lesson.slug}>
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:203]: this is a generally good practice at the top level of a very dynamic component.

src/pages/lessons/[slug].tsx
```diff
@@ -242,22 +211,51 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
...
+                </NextResourceButton>
+              </div>
+            )}
+          </div>
         </div>
+        <PlayerControls
+          handlerDownload={() => console.log('handlerDownload')}
+          isPro={true}
+        >
+          <NextResourceButton path={nextUpPath} onClick={() => send('LOAD')} />
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:257]: here's an example of using component composition

https://www.robinwieruch.de/react-component-composition

Instead of burying the next button in this component we can pass it in as a child. This means we don't have to pass in the `send` method to change state.

src/pages/lessons/[slug].tsx
```diff
@@ -242,22 +211,51 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
...
+                  <a>Get Access to This Video</a>
+                </Link>
+              </div>
+            )}
+
+            {playerState.value === 'showingNext' && (
+              <div className="flex justify-center items-center h-full">
+                <NextResourceButton
+                  path={nextUpPath}
+                  onClick={() => send('LOAD')}
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:245]: because of the dynamic nature of this page and the router we need to set the state machine!

src/pages/lessons/components/PlayerControls.tsx
```diff
@@ -0,0 +1,50 @@
...
+const availableSpeeds: string[] = ['0.85', '1', '1.25', '1.5', '1.75', '2']
+
+const PlayerControls: FunctionComponent<PlayerControlsProps> = ({
+  children,
+  handlerDownload,
+  isPro,
+}) => {
+  return (
+    <div className="flex items-center justify-between mt-4">
+      {children}
```
joelhooks, line 18: we can get the button or anything else now like this.

src/pages/lessons/lesson-player-machine.ts
```diff
@@ -0,0 +1,109 @@
+import {Machine} from 'xstate'
+
+interface PlayerStateSchema {
```
joelhooks, line 3: felt like this state machine was related to the lesson page more than the player specifically so I moved it here. We might consider consolidating all of this and the overlays into the player later?

DrShpongle (approved):

PR #95, opened 2020-09-22
fix useEffect

DrShpongle:


src/pages/lessons/[slug].tsx
```diff
@@ -199,6 +199,8 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
       case 'completed':
         send('NEXT')
         break
+      case 'showingNext':
```
Comment (DrShpongle) [inline src/pages/lessons/[slug].tsx:202]: it switches to default "loading" case without this option

joelhooks (commented):
I think we want to remove that default send instead.

joelhooks (approved):
Sorry, the default+break needs to stay. Just empty. Should solve problem. Maybe it’s fine like this. 😅

PR #99, opened 2020-09-23
Handling search url with instructor, but no tags

johnlindquist:
`tagsForPath` was incorrectly parsing "lessons-by" as a tag.

Added a test for "only instructor, no tag" and implemented.

### DIFF
--- a/src/lib/search-url-builder.ts
+++ b/src/lib/search-url-builder.ts
@@ -4,7 +4,7 @@ import get from 'lodash/get'
 import qs from 'query-string'
 import nameToSlug from './name-to-slug'
 import humanize from 'humanize-list'
-import {first, pickBy, isEmpty} from 'lodash'
+import {first, pickBy, isEmpty, isUndefined} from 'lodash'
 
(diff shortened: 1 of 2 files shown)


src/lib/search-url-builder.ts
```diff
@@ -31,11 +31,12 @@ const toTitleCase = (name: string) => {
 }
 
 const tagsForPath = (path: string) => {
-  const tagsSplit = path?.split('-lessons-by-') || []
+  const [tagsString] = path?.split('-lessons-by-') ?? []
 
-  const tags = tagsSplit.length >= 1 ? tagsSplit[0].split('-and-').sort() : []
+  if (isUndefined(tagsString)) return []
+  if (tagsString.startsWith('lessons-by')) return []
```
johnlindquist, line 37: This `startsWith` fixes the main issue. 

The other changes make sure we return an empty Array instead of `undefined` so the return type is always an Array of Strings.

joelhooks (approved):

PR #110, opened 2020-10-07
Formatted guide layouts and typography with tailwinds prose styling

MaggieAppleton:
Initial typography and layout styling setup for the ultimate guides and articles

image

## Changes

- Added next-images to help process images in articles
- Added gsap for scrolling animations (not used in any articles in this PR - coming next)

Setup css and tailwinds files with prose styling. Adjusted layout width, typography sizes, and leading on all pages.

Added a "Fancy Guide" layout (not in use yet, will add articles that use it in a separate PR)


src/styles/index.css
```diff
@@ -24,8 +24,28 @@ h1 {
...
+}
+
+pre {
+  @apply mx-2 px-8 py-6 bg-gray-800 text-gray-200 rounded-md;
+}
+
+/* Contain typography to constrained width */
+.prose h1, h2, h3, h4, h5, p, ul, ol, li, span {
+  @apply max-w-2xl mx-auto;
 }
```
vojtaholik, line 39: **suggestion:** Each prose size modifier already comes with a [baked in max-width](https://github.com/tailwindlabs/tailwindcss-typography#overriding-max-width). Setting max-width on higher level, e.g. [in templates](https://github.com/eggheadio/egghead-next/blob/ma/prose-typography-styling/src/layouts/ultimate-guide.tsx#L34) would be my preferred way of doing it. More importantly, this change would disallow us from setting different max-width on all prose classes. If we want to modify prose, we should do so for each size modifier in tailwind config.

src/styles/index.css
```diff
@@ -24,8 +24,28 @@ h1 {
...
+
+h1, h2, h3, h4, h5, p, ul, ol, li, span {
+  @apply max-w-2xl mx-auto;
+}
```
vojtaholik, line 43: **suggestion:** I would prefer setting max width and margin on a container wrapping these general elements (usually layout or template component) rather than global level.

src/utils/breakpoints.js
```diff
@@ -0,0 +1,14 @@
+const minSM = 640
+const minMD = 768
+const minLG = 1024
+const minXL = 1280
```
vojtaholik, line 4: **question**: What are these for? I didn't see them being used.

src/styles/index.css
```diff
@@ -15,7 +15,7 @@
 }
 
 h1 {
-  @apply text-lg;
+  @apply text-xl;
 }
```
vojtaholik, line 19: **thought:** I actually like tailwind's hard reset on everything and don't mind setting things like this on component level. I've noticed you're already modifying font size in prose config.

tailwind.config.js
```diff
@@ -18,6 +18,33 @@ module.exports = {
...
+        css: {
+          h1: {
+            lineHeight: '1.25',
+          },
+          h2: {
+            lineHeight: '1.33',
+          },
+          h3: {
+            lineHeight: '1.4',
+          },
```
vojtaholik, line 47: **suggestion:** reuse line heights from theme object: `lineHeight: theme('lineHeight.relaxed')` ([see available options](https://tailwindcss.com/docs/line-height))

src/styles/index.css
MaggieAppleton: Ah didn't realise the max-width changed between prose sizes. Can easily switch to modifying these in templates - will remove these.

dealingwith:
You had me at "leading".

src/utils/breakpoints.js
MaggieAppleton: Ah, they're used in a set of scrollytelling components I built for individual articles that I haven't PR-ed in yet. Will add them after the components and styling are all cleared. Happy to swap them for another system though. Hadn't seen facepaint before.

src/styles/index.css
MaggieAppleton: Is there a way for us to bump up the typographic scale across the whole site?
The default prose one doesn't have enough size distinction between h1, h2, h3. Want the h1's across the whole site to be bigger than prose sets by default and seemed like the easiest way to do it?

src/styles/index.css
MaggieAppleton: Yeah, we can do it at the template/component level - just removed the `.prose` size constraints. Should I nix the non-prose ones too?

tailwind.config.js
MaggieAppleton: Yup can swap it for the theme object

We should probably pick a consistent typeface(s) for the egghead-next site sooner rather than later so we can design around it and not worry about these changing based on the typeface. Squishy line heights are kind of a killer for readability for articles so would love to get these looking good early on.

src/utils/breakpoints.js
MaggieAppleton: Just swapped this for facepaint 🤡 493a456

tailwind.config.js
MaggieAppleton: Swapped to the theme objects :) ec24e33

src/styles/index.css
vojtaholik: I'd prefer tailwind's reset applied to everything by default. utility-first approach. h1's usually aren't part of prose wrapper and can be styled individually as we see fit.

tailwind.config.js
vojtaholik: I find prose defaults to have very relaxed line heights and read well. system fonts are fine for the foreseeable future imho.

src/layouts/ultimate-guide.tsx
```diff
@@ -30,8 +31,8 @@ const UltimateGuideLayout: FunctionComponent<LayoutProps> = ({
         }}
         canonical={url}
       />
-      <div className="prose max-w-none">
-        <h1>{title}</h1>
+      <div className="prose md:prose-xl max-w-2xl mx-auto leading-6">
+        <h1 className="text-xl">{title}</h1>
```
vojtaholik, line 35: nitpick: h1 should move above div for its classes to take effect.

src/styles/index.css
MaggieAppleton: Yup no worries that works, updated with 5bf101c and be8378a. Moved the size adjustments to the prose config so it'll only apply to articles and won't mess up font sizes across the site as a whole.

src/layouts/ultimate-guide.tsx
MaggieAppleton: Oh yeah didn't realise this was redundant. Removed it 72e2bac

src/styles/index.css
MaggieAppleton: Nixed: f409cdb

tailwind.config.js
```diff
@@ -14,10 +14,53 @@ module.exports = {
...
+      '2xl': '1.953em',
+      '3xl': '2.441em',
+      '4xl': '3.052em',
+      '5xl': '3.815em',
+    },
     typography: (theme) => ({
       default: {
         css: {
           color: theme('colors.gray.800'),
+          marginTop: '3em',
```
vojtaholik, line 31: might be better to set margin individually rather than on all prose classes

tailwind.config.js
```diff
@@ -14,10 +14,53 @@ module.exports = {
...
     },
+    fontSize: {
+      // Set in Major Third typescale (1.25)
+      base: '1em',
+      lg: '1.25em',
+      xl: '1.563em',
+      '2xl': '1.953em',
+      '3xl': '2.441em',
+      '4xl': '3.052em',
+      '5xl': '3.815em',
```
vojtaholik, line 25: can we use `rem`s instead of `em`s please? :)

joelhooks (approved):

PR #115, opened 2020-10-08
First pass at an embedded player

johnlindquist:
The `_app.js` check for an `/embed` at the end of the route and removes all layout/chrome if it exists.

This currently only matches a `/lessons/[slug]/embed.txt`, but maybe there are other scenarios where people would want to iframe some content from egghead? 🤔


src/pages/_app.tsx
```diff
@@ -25,6 +25,12 @@ Router.events.on('routeChangeComplete', () => window.ahoy.trackView())
 export default class App extends NextApp {
   render() {
     const {Component, pageProps} = this.props
+    const {pathname} = this.props.router
+
+    if (pathname.endsWith(`/embed`)) {
+      return <Component {...pageProps}></Component>
```
johnlindquist, line 31: Removing all wrapping components.

We probably still want some Head/Metadata/Tracking?

src/pages/lessons/[slug]/embed.tsx
```diff
@@ -0,0 +1,151 @@
+/** @jsx jsx */
```
Comment (johnlindquist) [inline src/pages/lessons/[slug]/embed.tsx:1]: I copy/pasted the lessons page and removed non-video player stuff

joelhooks:
the egghead embed player introduces a couple of things. Mainly it's using PReact and has several essential query params. I'd probably be inclined to deprecate it entirely tbh and just have folks link to lessons.

joelhooks:
I'm gonna close this for now.

PR #117, opened 2020-10-09
migrate back to basics with mdx

joelhooks:
I stripped out `next-mdx-enhanced`. The other community mdx solution is `next-mdx-remote` which is fine and good, but the tradeoff is that you can't import in mdx and frankly I think that sort of overhead long term is probably not great.

But, there is the official Vercel backed `next-mdx` which is more basic and doesn't support frontmatter in the same fashion (you can see here it uses more js/ts syntax) and doesn't support templates as simply (you have to export them from the mdx)

That said, it's also going to be the most maintainable long term and relies on Vercel core implementation instead of community churn at the expense of mdx files being more programmery.


src/pages/index.mdx
```diff
@@ -105,3 +105,9 @@ When you commit to leveling up with a year of egghead, you will get instant acce
 You can use your new skills to ship a new project, negotiate a raise, or just for fun. But whatever you do, don't wait until this deal's over.
 
 [Click here to become an egghead member!](https://egghead.io/pricing)
+
+import DefaultLayout from '../layouts/index'
```
joelhooks, line 109: my feeling generally is that this is a fine way to handle templates

struggled to avoid needing an `import` for the default template. That'd make it a lot nicer.

src/pages/learn/react/accessibility/index.mdx
```diff
@@ -1,6 +1,11 @@
----
-layout: 'ultimate-guide'
-title: 'React Accessibility Ultimate Guide'
----
+export const meta = {
```
joelhooks, line 1: for me, this approach to frontmatter is fine and actually preferred because it doesn't require an oddball mix of YAML syntax and we just stick to typescript FTW.

johnlindquist: Yah, yaml is 💩

src/pages/index.mdx
johnlindquist: I took a stab at it, but failed. I think we'd have to customize babel in next.config.js `withMDX` to accept `https://babeljs.io/docs/en/babel-plugin-proposal-export-default-from`, but my effort was in vain, so 🤷‍♂️

johnlindquist (approved):

src/pages/index.mdx
joelhooks: ya that was my observation too and concluded that staying "stock" is solid goal where possible

leerob:
If this is any help, here's how Vercel's docs are set up using `@next/mdx`.

```js
import Guide from '~/components/layout/guide'

export const meta = {
  title: `Using Express.js with Vercel`,
  description: `Learn how to use Express.js in a Serverless environment.`,
  published: '2020-10-02T17:00:00.000Z',
  authors: ['okbel', 'leerob'],
  url: '/guides/using-express-with-vercel',
  image: `https://og-image.now.sh/Using%20**Express.js**%20with%20Vercel.png?theme=light&md=1&fontSize=100px&images=https%3A%2F%2Fassets.vercel.com%2Fimage%2Fupload%2Ffront%2Fassets%2Fdesign%2Fvercel-triangle-black.svg`,
  editUrl: 'pages/guides/using-express-with-vercel.mdx',
  lastEdited: '2020-10-02T17:00:00.000Z'
}

# Hello

This is Markdown.

export default ({ children }) => <Guide meta={meta}>{children}</Guide>

export const config = {
  amp: 'hybrid',
}
```

joelhooks:
Thanks @leerob that looks like where we landed here and it's a very nice KISS solution imo!

PR #123, opened 2020-10-16
Sitemap generation based on tags

johnlindquist:
- Grab all the tags from the egghead v1 api (currently 158 tags)
- Generate sitemaps based on combinations of 3 or fewer tags
- Sitemaps have 10,000 entries each
- It only takes ~3 seconds to run

I think this is a simple and solid approach. It's fun to click on all the links in the sitemaps and see them work.


joelhooks (approved):
@johnlindquist neat! Is the next step to add instructor names? That is probably only interesting if the instructor has lessons for a given tag?

PR #133, opened 2020-10-27
fixing search bug

johnlindquist:
This was a three-part bug:

1. `react-instantsearch-dom` 6.8 has breaking changes
[react-instantsearch 6.8 breaking changes](https://github.com/algolia/react-instantsearch/issues/2978)

We settled on updating to 6.8.2. I didn't update the package.json in this commit, since ^6.7 assumes 6.8. But the next depfu updates contains the hard-coded version.

2. 6.8 requires metadata
We were stripping `metadata` from the "resultsState" when working with "rawResults", I added them back in.

3. Next.js JSON validation error
[Next.js getServerSideProps serialization validation](https://github.com/vercel/next.js/issues/17215)
The metadata has a function which breaks the Next.js validation JSON logic. I need to talk to Tim about best practices here, but seems [we're not alone](https://github.com/vercel/next.js/issues/17215#issuecomment-695174835)


src/pages/s/[[...all]].tsx
```diff
@@ -86,7 +86,7 @@ export const getServerSideProps: GetServerSideProps = async function ({
   const {all, ...rest} = query
   const initialSearchState = parseUrl(query)
   const pageTitle = titleFromPath(all as string[])
-  const {rawResults, state} = await findResultsState(Search, {
+  const {rawResults, metadata, state} = await findResultsState(Search, {
```
Comment (johnlindquist) [inline src/pages/s/[[...all]].tsx:89]: Adding metadata back in

src/pages/s/[[...all]].tsx
```diff
@@ -98,11 +98,13 @@ export const getServerSideProps: GetServerSideProps = async function ({
   const noIndexInitial = queryParamsPresent || noHits || userQueryPresent
 
   return {
-    props: {
-      resultsState: {rawResults},
-      initialSearchState,
-      pageTitle,
-      noIndexInitial,
-    },
+    props: JSON.parse(
```
Comment (johnlindquist) [inline src/pages/s/[[...all]].tsx:101]: forcing JSON :/

Comment (joelhooks) [reply src/pages/s/[[...all]].tsx:101]: parsing json doesn't result in json, technically speaking, it results in a JS object.

another consideration here would be to stop destructuring the resultsState and then using the prop shorthand? vI'll pull the branch and have a peek.

Comment (joelhooks) [reply src/pages/s/[[...all]].tsx:101]: ah, I see, some weird ass deep object from Algolia doesn't play nice with Next. Cool lol

src/pages/s/[[...all]].tsx
```diff
@@ -99,7 +101,7 @@ export const getServerSideProps: GetServerSideProps = async function ({
 
   return {
     props: {
-      resultsState: {rawResults},
+      resultsState: JSON.parse(JSON.stringify(resultsState)),
```
Comment (joelhooks) [inline src/pages/s/[[...all]].tsx:104]: I changed this a bit so that instead of destructuring the resultsState that algolia feeds us we pass along the full object and do the destructuring in a separate step.

joelhooks (approved):

src/pages/s/[[...all]].tsx
Comment (ItsWendell) [reply src/pages/s/[[...all]].tsx:104]: Nice, encountered the same issue a while back, this solved it for me too!

PR #135, opened 2020-10-28
Adds Cypress tests on the lesson page

dealingwith:
This is a first pass at creating a test that confirms a lesson page(s) load and the video loads (or at least the preview image does).

There is also a test to make sure the "code" tab exists and is clickable. This may be unnecessary.

tenor (1)


cypress/integration/lesson_page_spec.js
```diff
@@ -0,0 +1,16 @@
+describe('lesson page', () => {
+  it('/lesson loads', () => {
+    cy.visit(
+      '/lessons/react-using-react-error-boundaries-to-handle-errors-in-react-components',
+    )
+    cy.expect('#bmpui-id-204 > .bmpui-image').to.exist
```
dealingwith, line 6: I'm not crazy about the selectors available to us on this page. Even though they're IDs, they feel fragile, but maybe I'm overthinking it?

cypress/integration/lesson_page_spec.js
```diff
@@ -0,0 +1,16 @@
...
+
+    // the code tab is there and clickable
```
dealingwith, line 8: Is this really necessary? How much of this page do we want tests around?

cypress/integration/lesson_page_spec.js
```diff
@@ -0,0 +1,16 @@
...
+    cy.expect('#tabs--1--tab--0').to.exist
+    cy.get('#tabs--1--tab--1').click().get('#tabs--1--panel--1 > p')
```
dealingwith, line 10: Is this correct? Does it need an `expect`?

cypress/integration/lesson_page_spec.js
```diff
@@ -0,0 +1,16 @@
...
+
+    // another free video page
```
dealingwith, line 12: Similarly, testing a 2nd page might not be necessary.

dealingwith:
...still playing with this...update forthcoming

cypress/integration/lesson_page_spec.js
```diff
@@ -3,8 +3,7 @@ describe('lesson page', () => {
     cy.visit(
       '/lessons/react-using-react-error-boundaries-to-handle-errors-in-react-components',
     )
-    // failing
-    cy.get('#bmpui-id-204').should('be.visible')
+    cy.get('#egghead-player', {timeout: 10000}).should('exist')
```
lsminter, line 6: Changed this to look for the player itself. The reason this was failing was that it was timing out before it could find the element. I extended the timeout to be ten seconds and it worked. I'm unsure what would be the best practice here because just extending the timeout feels wrong.

johnlindquist: I need to look into getting the video player to load faster.

joelhooks: It won’t load unless it has the media URLs which load async. We could potentially load them in SSR because we have access to the token. This would make loads of sense, but we would want to make sure that caching doesn’t distribute protected media URLs Willy nilly

joelhooks: I also think waiting 10 seconds is totally fine.

joelhooks:
Is this ready to go? It is labeled `help wanted` but I don't see a specific question. What sort of help is wanted?

dealingwith:
@joelhooks I think this is good for the first pass at tests for the lesson page. Suggestions on how to improve existing tests, or additional tests to add to this page, are welcome, however.

johnlindquist (approved):

PR #147, opened 2020-11-09
removes instructors spec

dealingwith:
https://github.com/eggheadio/egghead-next/commit/4ac6ccf3140be6f0455d0427cff11bf4968979cc removed the instructor page, which caused the Cypress tests to fail. This removes the instructors spec.

image


dealingwith:
Looking to add some tests to the `/s/` endpoint but this keeps the tests from failing constantly.

johnlindquist (approved):

PR #150, opened 2020-11-10
feat: Integrate basic story for Podcast Card

jh3y:
- Basic `Storybook` integration
- Promotes splitting of logic from UI
- Converted `postcss.config.js` to keep everything happy

~~⚠️  Does not work with `next/image` from what I can see~~

- Works with `next/image` by adjusting the `webpack` config to substitute for native `img` element in `Storybook` land.
- We may want to define size in `tailwind` rems that we'd like `next/image` to use for the Podcast cards, etc. Or is there one that's already being used globally? Else, I've configured a `height` and `width` to match the `height` and `width` used in the component for the `Image` 👍 

ezgif com-gif-maker (5)

ezgif com-gif-maker (6)

This PR had me feeling like this as I battled `webpack`/`postcss` 😅  

"Oh, `Storybook`! That's a good idea 💡 "
...
"I despise Webpack && PostCSS" 😅 
...
"I love Webpack && PostCSS" ❤️ 

cat chasing tail


joelhooks:
@jh3y hilarious gif is a hard requirement for PR descriptions

You'll also want to pick a reviewer!

jh3y:
I'm still playing around with this one a little. 

An appropriate GIF will be attached 💯

src/pages/podcasts/[slug].tsx
```diff
@@ -46,7 +25,13 @@ const MorePodcasts: FunctionComponent<MorePodcastProps> = ({podcasts}) => (
       </h3>
       <ul className="mb-10 justify-items-center grid gap-4 grid-cols-1 sm:grid-cols-2 md:grid-cols-3">
         {podcasts.map((podcast: PodcastResource) => (
-          <PodcastCard podcast={podcast} key={podcast.id} />
+          <PodcastCard
```
Comment (joelhooks) [inline src/pages/podcasts/[slug].tsx:28]: Should this just take a `PodcastResource` as a prop instead of breaking it apart like this in the parent? Then the card can make its own life choices about what it uses.

src/pages/podcasts/[slug].tsx
```diff
@@ -4,6 +4,7 @@ import {loadPodcast, loadPodcasts} from 'lib/podcasts'
 import Rss from '../../components/images/rss.svg'
 import Spotify from '../../components/images/spotify.svg'
 import ApplePodcasts from '../../components/images/apple-podcasts.svg'
+import PodcastCard from '../../components/podcasts/card'
```
Comment (joelhooks) [inline src/pages/podcasts/[slug].tsx:7]: you can use `components/podcasts/card` here as all top level `src/*` folders are aliased.

src/pages/podcasts/[slug].tsx
Comment (jh3y) [reply src/pages/podcasts/[slug].tsx:28]: I'm cool either way on this 👍 

I think because of `TypeScript`, it's more forgiving because that will enforce what we're up to.

Happy to revert it back. The only downside is that it's not overly informative if someone views the code in the Storybook. However, that's easily mitigated by documenting things if needed in the MDX.

src/pages/podcasts/[slug].tsx
Comment (jh3y) [reply src/pages/podcasts/[slug].tsx:7]: Rad 😎 🚀

jh3y:
Resolved 👍

src/components/podcasts/card.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+const PodcastCard: FunctionComponent<PodcastCardProps> = ({
+  title,
+  path,
+  imageUrl,
+  contributors,
+}) => (
+  <li className="list-none bg-white p-3 text-center max-w-xs shadow-md rounded-md transform transition-transform duration-300 hover:scale-105">
+    <a title="View podcast" href={path} className="flex flex-col h-full">
+      <div className="mb-0 h-cardimage w-cardimage flex-grow-0">
```
joelhooks, line 14: instead of globally applying a size in the tailwind config, we might consider using emotion for this singular case of a specified size. It would look like:

```
css={{
  width: '200px',
  height: '200px'
}}
```

I believe you need:

```
import {jsx} from '@emotion/core'```

up top as well for the `css` prop to function, but that's its intent and why emotion is in the codebase.

We generally want to keep tailwind config very non-specific and reserved for truly global concerns

Comment (jh3y) [reply src/components/podcasts/card.tsx:14]: Yeah, I wasn't sure how to do this in an elegant way. Either, using a predefined size from tailwind like `h-64` && `w-64` and then setting the `width` and `height` as `64 x 16` on the `Image` component.

Is there a nice way in the repo to create a shared value? 

Essentially, a `size` that is an Integer that can also be used as a `px` value by this `div`?

Comment (jh3y) [reply src/components/podcasts/card.tsx:14]: Ahh right, I see 👍

File: .storybook/main.js [hunk 4]
```diff
@@ -0,0 +1,21 @@
+const path = require('path')
+const webpack = require('webpack')
+module.exports = {
+  stories: ['../src/**/*.stories.mdx', '../src/**/*.stories.@(js|jsx|ts|tsx)'],
+  addons: ['@storybook/addon-links', '@storybook/addon-essentials'],
+  webpackFinal: (config) => {
+    // Necessary to "mock" next/image in Storybook land
+    config.plugins.push(
+      new webpack.DefinePlugin({
```
joelhooks, line 9: 😱

Is this brittle? I panicked a bit when I first saw it but it looks contained specifically to storyboard.

jh3y: Yeah, this is specific to the `Storybook` environment 👍 

`next/image` won't work in the `Storybook` environment. 

`Storybook` use highlights some places where we are coupling the data requests to the UI. For example, `search/stack`. That component will not work as is in `Storybook` as it relies on functions like `loadLesson`, etc. which are reliant on `next`.

But yeah, anything inside the `.storybook` directory won't have an effect on things outside of it 👍

joelhooks (approved):

PR #153, opened 2020-11-11
checkout and pricing

joelhooks:
This adds checkout flow.

* load price from stripe
* pick a plan (only 1 currently)
* get email if not logged in
* redirect to stripe checkout
* bring them back after 

Requires this to be merged on the backend: https://github.com/eggheadio/egghead-rails/pull/4155

scrooge mcduck counting his money


src/api/stripe/stripe-checkout-redirect.ts
```diff
@@ -0,0 +1,36 @@
...
+if (!process.env.NEXT_PUBLIC_STRIPE_PUBLIC_KEY) {
+  throw new Error('no Stripe public key found')
+}
+
+const stripePromise = loadStripe(process.env.NEXT_PUBLIC_STRIPE_PUBLIC_KEY)
+
+const stripeCheckoutRedirect = async (priceId: string, email: string) => {
+  return await axios
+    .post(
+      `${process.env.NEXT_PUBLIC_AUTH_DOMAIN}/api/v1/stripe/subscription`,
```
joelhooks, line 14: loads a checkout session from backend with the configured price ID

src/api/stripe/stripe-checkout-redirect.ts
```diff
@@ -0,0 +1,36 @@
...
+        client_id: process.env.NEXT_PUBLIC_CLIENT_ID,
+        success_url: `${process.env.NEXT_PUBLIC_AUTH_DOMAIN}?session_id={CHECKOUT_SESSION_ID}`,
+        cancel_url: `${process.env.NEXT_PUBLIC_AUTH_DOMAIN}/pricing`,
+      }),
+    )
+    .then(({data}) => {
+      stripePromise.then((stripe: any) => {
+        if (!stripe) throw new Error('Stripe not loaded 😭')
+        stripe
+          .redirectToCheckout({
```
joelhooks, line 28: send us to Stripe if all looks good!

src/components/pricing/email-form.tsx
```diff
@@ -0,0 +1,128 @@
...
+  const validateEmail = async (email: string) => {
+    const isPro = await axios
+      .post(`/api/users/check-pro-status`, {
+        email,
+      })
+      .then(({data}) => data)
+
+    console.log(isPro)
+
+    if (isPro) {
```
joelhooks, line 29: this is a quick and dirty lookup to see if we already have a subscription on this particular email

not great UX

src/components/pricing/email-form.tsx
```diff
@@ -0,0 +1,128 @@
...
+      setIsError(
+        `You've already got a pro account at ${email}. [Please login](/login).`,
+      )
+    } else {
+      setIsError(false)
+      stripeCheckoutRedirect(priceId, email)
```
joelhooks, line 35: no sub? let's fix that quick.

src/components/pricing/email-form.tsx
```diff
@@ -0,0 +1,128 @@
...
+    }
+  }
+
+  return (
+    <div className="text-text w-full  mx-auto flex flex-col justify-center sm:mt-24 mt-5">
```
joelhooks, line 40: I just reused the login form here 😅

src/lib/stripe/price.ts
```diff
@@ -0,0 +1,15 @@
...
+}
+
+const stripe = new Stripe(process.env.STRIPE_SECRET_KEY, {
+  apiVersion: '2020-08-27',
+})
+
+export async function loadPrice(priceId: string) {
+  const price = await stripe.prices.retrieve(priceId)
+
+  return price
```
joelhooks, line 14: this is a Stripe `Price` object

src/lib/stripe/price.ts
```diff
@@ -0,0 +1,15 @@
+import Stripe from 'stripe'
+
+if (!process.env.STRIPE_SECRET_KEY) {
```
joelhooks, line 3: this is a concern with this because technically it could be imported into a frontend file and exposed, where currently it's being used in a SSR method, and doesn't expose it.

Alternative would be to be more explicit and migrate to a `/src/pages/api` function where it might be more clear it's not meant to be used in the UI. 🤔

src/pages/pricing/index.tsx
```diff
@@ -1,3 +1,81 @@
...
+  annualPrice: {
+    id: string
+    recurring: {
+      interval: 'year'
+    }
+    unit_amount: number
+  }
+}
+
+const Pricing: FunctionComponent<PricingProps> = ({annualPrice}) => {
```
joelhooks, line 40: modestly sloppy UX that could probably use a little state machine...

src/pages/api/users/check-pro-status.ts
```diff
@@ -0,0 +1,32 @@
...
+
+      const eggheadUser = await axios
+        .get(userUrl, {
+          headers: {
+            Authorization: `Bearer ${process.env.EGGHEAD_SUPPORT_BOT_TOKEN}`,
+          },
+        })
+        .then(({data}) => data)
+
+      res.status(200).json(!isEmpty(eggheadUser.subscription))
```
joelhooks, line 26: we are checking if the user has a subscription, which is probably not a sufficient check but will at least prevent a double subscription for now

joelhooks (commented):
This is "safe" in that it can be live and if it gets used people should be fine and get a membership, but there are a lot of rough UX edges.

src/api/stripe/stripe-checkout-redirect.ts
```diff
@@ -0,0 +1,36 @@
...
+const stripeCheckoutRedirect = async (priceId: string, email: string) => {
+  return await axios
+    .post(
+      `${process.env.NEXT_PUBLIC_AUTH_DOMAIN}/api/v1/stripe/subscription`,
+      pickBy({
+        price_id: priceId,
+        site: 'egghead.io',
+        email: email,
+        client_id: process.env.NEXT_PUBLIC_CLIENT_ID,
+        success_url: `${process.env.NEXT_PUBLIC_AUTH_DOMAIN}?session_id={CHECKOUT_SESSION_ID}`,
```
theianjones, line 20: should this be:

```suggestion
        success_url: `${process.env.NEXT_PUBLIC_REDIRECT_URI}?session_id={CHECKOUT_SESSION_ID}`,
```

So the user comes back to next.egghead.io?

src/api/stripe/stripe-checkout-redirect.ts
```diff
@@ -0,0 +1,36 @@
...
+        cancel_url: `${process.env.NEXT_PUBLIC_AUTH_DOMAIN}/pricing`,
```
theianjones, line 21: same here 
```suggestion
        cancel_url: `${process.env.NEXT_PUBLIC_REDIRECT_URI}/pricing`,
```

theianjones (approved):

PR #148, opened 2020-11-10
Timecodes for Transcript + UI tweaks on the Lesson page

DrShpongle:
Screen Recording 2020-11-10 at 07 27 59 PM

Note: I was forced to downgrade `react-markdown` from 5.0.2 --> 4.3.1, because I was not able to get `renderers` prop working. They have [braking changes](https://github.com/remarkjs/react-markdown/blob/main/changelog.md#new-serializer-property-node) in version 5.x.x but even example from the docs didn't work for me (probably I did it wrong).

thanks @vojtaholik for the reference 🙌

PR #154, opened 2020-11-11
Minor fix for the state

DrShpongle:
It fixes state changing on the Lesson page (missed line in [previous PR](https://github.com/eggheadio/egghead-next/pull/148))


src/pages/lessons/[slug].tsx
```diff
@@ -218,7 +218,10 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
                   pip="true"
                   controls
                   onPlay={() => send('PLAY')}
-                  onPause={() => send('PAUSE')}
+                  onPause={() => {
+                    send('PAUSE')
+                    setIsPlaying(false)
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:223]: technically we should know "is playing" because the state is currently "playing" and there is no need for this extra `useState`

what do you think? why do we need `isPlaying` in addition to `playing` state?

Comment (DrShpongle) [reply src/pages/lessons/[slug].tsx]: pushed the fixes

src/components/pages/lessons/Transcript.tsx
```diff
@@ -61,7 +61,7 @@ const Transcript: FunctionComponent<TranscriptProps> = ({
   }
 
   useEffect(() => {
-    if (dataText) {
+    if (dataText && player.current) {
```
joelhooks, line 64: if we don't have a player (overlay) we don't want to have timestamp links

src/pages/lessons/[slug].tsx
```diff
@@ -303,7 +303,7 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
                         player={playerRef}
                         url={transcript_url}
                         fetcher={fetcher}
-                        send={send}
+                        playVideo={() => send('PLAY')}
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:306]: since we don't need to control the entire state machine, send in a specific method to do just what we want

joelhooks (approved):

PR #157, opened 2020-11-11
actually make linter work

joelhooks:
whew!

So I turned on ES Lint, which in particular spots some issues with React hooks. I actually ONLY turned on React Hooks and a11y linters for now.

Now common error that it picked up was using hooks conditionally.

@vojtaholik @DrShpongle you can't use React hooks conditionally. That's a hard rule, but I'll comment inline about where it was an issue.

wrath of the lint god


next.config.js
```diff
@@ -13,6 +13,9 @@ checkEnv({
 
 const nextConfig = {
   reactStrictMode: true,
+  experimental: {
+    reactMode: 'concurrent',
```
joelhooks, line 17: this is what I set out to do 😅

src/components/login-required.tsx
```diff
@@ -14,7 +14,7 @@ const LoginRequired: FunctionComponent<LoginRequiredParams> = ({
     if (loginRequired) {
       router.push('/login')
     }
-  }, [])
+  }, [loginRequired])
```
joelhooks, line 17: this is a fairly common and difficult issue

when the boolean changes, we want them to login, so it should re-run

src/components/pages/lessons/LessonInfo.tsx
```diff
@@ -84,10 +84,9 @@ const CopyToClipboard: FunctionComponent<{
   stringToCopy: string
   className?: string
 }> = ({stringToCopy, className = ''}) => {
+  const [copied, setCopied] = useState(false)
+  const [state, copyToClipboard] = useCopyToClipboard()
   if (stringToCopy) {
-    const [copied, setCopied] = useState(false)
```
joelhooks, line 88: @DrShpongle no hooks can run inside an `if` or any other conditional. This is tricky, but now the linter should break if it happens.

src/components/search/components/stack.tsx
```diff
@@ -6,63 +6,62 @@ import {loadCourse} from 'lib/courses'
...
 type StackProps = {
-  resources: any[]
+  resources: StackResource[]
 }
 
-const Stack: FunctionComponent<StackProps> = ({resources}) => {
-  let res: any[] = []
-  resources.forEach((resource) => {
-    if (resource.type === 'lesson') {
-      const {data: lesson} = useSWR(resource.slug, loadLesson)
```
joelhooks, line 17: @vojtaholik this was a prime example of a conditional hook use. Instead, I broke the element out into individual components.

src/components/search/components/stack.tsx
```diff
@@ -6,63 +6,62 @@ import {loadCourse} from 'lib/courses'
...
+    case 'course':
+      loader = loadCourse
+      break
+    case 'playlist':
+      loader = loadPlaylist
+      break
+    default:
+      break
+  }
+  const {data: resource} = useSWR(slug, loader)
```
joelhooks, line 30: I'm not stoked on how this functions on the page because of the loading causing visible shifts. I'd prefer that it was fully static not only for visible aesthetics, but the overall SEO of the page. Since we are opting for convenience here, the good SEO friendly text will be ignored since it is async.

src/components/search/index.tsx
```diff
@@ -223,12 +222,12 @@ const Search: FunctionComponent<SearchProps> = ({
                     <RefinementList isShown={isFilterShown} attribute="type" />
                   </div>
                   {isRefinementOn && (
-                    <div
+                    <button
```
joelhooks, line 225: a11y linter really hates onclick when it's not a button and demands all sorts of changes/additions so I just made it a button

src/pages/_app.tsx
```diff
@@ -32,7 +33,7 @@ export default class App extends NextApp {
         <SocialProfileJsonLd
           type="Organization"
           name="egghead.io"
-          url={useURL()}
+          url={`${process.env.NEXT_PUBLIC_DEPLOYMENT_URL}`}
```
joelhooks, line 36: this was what React deemed a hook in a class component, though the actual useUrl isn't really a hook hook, but it is based on a naming convention.

src/pages/lessons/[slug].tsx
```diff
@@ -79,14 +79,11 @@ const lessonLoader = (slug: string, token: string) => {
...
-  onClick: () => void
   className: string
-}> = ({children, path, onClick, className = ''}) => {
+}> = ({children, path, className = ''}) => {
   return (
     <Link href={path || '#'}>
-      <a className={className} onClick={onClick}>
-        {children || 'Next Lesson'}
-      </a>
+      <a className={className}>{children || 'Next Lesson'}</a>
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:86]: this was another case where `onclick` is very demanding, I went for another option instead and monitor for a route change to set the state.

jh3y (approved):
Looks good to me! 🚀

src/components/search/components/stack.tsx
vojtaholik: Static version of the same thing is here: https://github.com/eggheadio/egghead-next/blob/main/src/components/search/curated/react.tsx#L100

src/components/search/components/stack.tsx
vojtaholik: Noted!

src/components/search/components/stack.tsx
joelhooks: So can this be deleted entirely then?

src/components/pages/lessons/LessonInfo.tsx
DrShpongle: Didn't know it, thanks!

PR #159, opened 2020-11-12
feat: refactor podcast related components + search hits into stories

jh3y:
- Decoupled some more UI pieces from logic and broke into stories so we can see what we're working with 👍 
- As you correctly stated, not a lot comes back for a `podcast` resource 🤔  I've attached an Object to each story for each resource type which gives a bit of transparency without having to go and grab it from the API
- Wanted to switch the main podcast image to use `next/image`. However, this posed a problem with `tailwind`'s `.prose img` styling. Something to be wary of if you're going to take advantage of `next/image` within `.prose`. It needs to be moved out so you don't get unwanted `margin` in places.

Feels like I've been doing some "Marie Kondo"

Marie shuffling stuff


joelhooks:
image

I'm seeing a double footer here.

I'm also getting significant layout shift

jh3y:
Yeah, I'm seeing this now 👍 

I'm struggling to see why it's happening though 🤔  I haven't changed anything on the layouts. I've taken the UI code and extracted it into components.

jh3y:
OK. Had a look. Unless my version is cached, this is actually happening on the `main` branch also 👎 

Something is causing a layout shift and double footer. The route perhaps?

jh3y:
Can confirm this is on `main` not related to this PR 👍 
https://next.egghead.io/s?type=podcast

vojtaholik (approved):

PR #161, opened 2020-11-13
Handle traffic to legacy urls

jbranchaud:
In order for `next.egghead.io` to live at `egghead.io` with the Rails app moving over to `app.egghead.io`, traffic hitting _legacy_ URLs will need to be directed to wherever makes most sense.

To that end, this PR does two things:

1. Introduce a custom 404 page that suggests visiting `app.` for any unrecognized URLs. This acts as a catch-all for any routes that don't get handled with explicit redirects and acts as a 404 page for any paths that we'd not expect to be found. I didn't attempt any styling on this component for the time being.

2. Add some additional redirects to the `vercel.json` file (renamed from `now.json`) for paths that seemed like necessary and obvious wins. I added a couple internal redirects for things like `/browse` that should go to `/s`. I also introduced some external redirects to `app.egghead.io` with globbing for groups of routes like `/admin` and `/users`.

There are a lot more routes to potentially handle, but it is going to take a bit of time to parse through all of them. This is an initial effort to get that ball rolling.


jbranchaud:
It looks like there are also redirects defined in the `next.config.js` file.

```javascript
  async redirects() {
    return [
      {
        source: '/instructors/:instructor',
        destination: '/s/content-by-:instructor',
        permanent: true,
      },
      {
        source: '/podcasts',
        destination: '/s?type=podcast',
        permanent: true,
      },
...
```

From what I can glean from the docs ([Vercel Redirects](https://vercel.com/docs/configuration#project/redirects) and [Next.js Redirects](https://nextjs.org/docs/api-reference/next.config.js/redirects)), they both behave the same way, acting at the web server routing layer.

We can put them either place, but for consistency sake, should pick one or the other.

joelhooks:
since the now|vercel.json ONLY has redirects we could reduce to a single file by using the config.

I don't really feel strongly either way.

jbranchaud:
Vercel and Next Config both support Redirects, Rewrites, altering headers. Vercel's config also lets you drop down an abstraction layer with Routes configuration -- though they kinda suggest only using that if you really need it. I don't think this app will need it.

That said, I just discovered that the Next Config redirects work for local development, whereas the Vercel config can only be tested in a deployment. That tips the scales to using the Next config for me.

joelhooks (approved):

PR #162, opened 2020-11-13
Add redirects for egghead RSS feeds

jbranchaud:
The goal of this PR is to redirect RSS feed requests to `app.egghead.io`.

I'm not certain if it will work as is or if it will require some customization of the headers, since the format is rss. Because Vercel does preview deployments, I can test it out before merging.


joelhooks:
Looks like it works as expected!

joelhooks (approved):
🦀

PR #165, opened 2020-11-17
Set dynamic maxWidth to the video on the Lesson page

DrShpongle:
preview: https://share.getcloudapp.com/E0urD8pB


src/pages/lessons/[slug].tsx
```diff
@@ -138,7 +144,7 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
     free_forever,
   } = lesson
 
-  React.useEffect(() => {
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:141]: @DrShpongle please use the `React.useState` syntax. This is to distinguish clearly from React internat hooks and user hooks.

joelhooks (approved):

Issue #171, opened 2020-11-19
New redirect configs are overriding existing next.egghead.io pages


lsminter:
This [PR](https://github.com/eggheadio/egghead-next/pull/161) added redirects but it appears that it is being too aggressive and redirecting from existing pages to legacy pages on app.egghead.io. For example, navigating to https://next.egghead.io/courses/record-badass-screencasts-for-egghead-io through next.egghead.io (eg. from the search) does not redirect but pasting the link into the browser redirects.

joelhooks:
looks like this was fixed!

PR #172, opened 2020-11-19
Prevent redirect from wiping out courses pages

jbranchaud:
The wildcard redirect for `/courses/:title/:all*` was globbing in URLs
like /courses/some-course-title when that should be served directly from
the Next.js pages directory. Using `.+` regex syntax ensures that the
redirect rule only matches for courses routes that have additional URL
structure after the title.

Addresses #171

With this PR:

- `/courses/some-course-title` should resolve as a 200 request to next.egghead.io
- `/courses/some-course-title/edit` should resolve as a 308 redirect to app.egghead.io

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -49,8 +49,8 @@ const nextConfig = {
         permanent: true,
       },
       {
-        source: '/courses/:title/:all*',
-        destination: 'https://app.egghead.io/courses/:title/:all*',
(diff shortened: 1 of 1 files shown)


lsminter (approved):
Looks good to me! Just tested with Daniel and it all worked perfectly.

PR #175, opened 2020-11-25
Link to code on the Lesson page

DrShpongle:
thanks @theianjones for the help 🙌


src/components/pages/lessons/LessonInfo.tsx
```diff
@@ -153,6 +216,7 @@ const LessonInfo: FunctionComponent<LessonInfoProps> = ({
   lesson,
   ...restProps
 }) => {
+  const hasCode = !Object.values(lesson.code || {}).every((x) => x === null)
```
DrShpongle, line 219: Here I check if `lesson.code` fields values are not `null` (Lodash's `isEmpty` is not useful here)

joelhooks: @theianjones I think the graphql endpoint could probably make this **way** simpler by just returning a `code_url` instead of forcing all this hoop jumping into the client.

joelhooks: suggestion: @DrShpongle i think we should fix the api because all we need is a url and all of this work is messy and not the UI's business

what do you think?

DrShpongle: I totally agree, just tried to make best with what I have. But yes - getting just the url will make things cleaner.

theianjones: I added a `code_url` field and a `repo_url` field to the lesson graphql type here https://github.com/eggheadio/egghead-rails/pull/4175/

DrShpongle: refactored due to the latest changes in API

src/components/pages/lessons/LessonInfo.tsx
```diff
@@ -156,6 +156,8 @@ const LessonInfo: FunctionComponent<LessonInfoProps> = ({
   lesson,
   ...restProps
 }) => {
+  const code_url = get(lesson, 'code_url')
```
joelhooks, line 159: you could eliminate these lines by using `if(lesson?.code_url)` below

DrShpongle: oh, not completely

DrShpongle: pushed, does it what you mean?

src/components/pages/lessons/LessonInfo.tsx
```diff
@@ -60,22 +60,24 @@ const TweetLink: FunctionComponent<{
   }
   className?: string
 }> = ({lesson, instructor, className = ''}) => {
+  const encodeTweetUrl = () => {
```
joelhooks, line 63: made this so it is parseable by human, for some reason it was affecting my syntax highlighting in VS Code

src/components/pages/lessons/LessonInfo.tsx
```diff
@@ -122,6 +124,21 @@ const CopyToClipboard: FunctionComponent<{
   return null
 }
 
+const CodeLink: FunctionComponent<{url: string}> = ({url, children}) => {
```
joelhooks, line 127: made this a simple component

joelhooks (approved):

PR #178, opened 2020-11-27
fix Transcript render

DrShpongle:
It renders Transcription only when there is transcription exists (https://next.egghead.io/lessons/bash-interactively-find-anything-by-piping-it-to-fzf-in-the-terminal?pl=upgrade-your-terminal-and-make-it-a-joy-to-work-with-13f1 - an example of the lesson with no transcription)

Image 2020-11-27 at 7 19 36 PM


joelhooks:
future: @lsminter @zacjones93 would be interesting I think if this sent an automated message to us somehow so that we'd see lessons that were missing transcripts and could deal with them accordingly.

I think most of these are the "double video latest transcript order was cancelled" bug that we've had for a long time. The backend could also intelligently handle that when a transcript is missing I suspect 🤔

DrShpongle:
@joelhooks if you think it should be handled on the front-end could you advice the direction where I should to look for solution?

joelhooks:
```bash
07:11:06.792 | ./src/pages/lessons/[slug].tsx:350:24
07:11:06.792 | Type error: Type '{ player: MutableRefObject<null>; playVideo: () => State<PlayerContext, PlayerStateEvent, any, { value: any; context: PlayerContext; }>; transcriptText: any; }' is missing the following properties from type 'TranscriptProps': fetcher, url
```

Looks like some properties are missing in the types

joelhooks (approved):

DrShpongle:
fixed!

PR #183, opened 2020-12-01
Handle access token cookie for shared auth

jbranchaud:
This PR is the companion of the authentication changes introduced server-side in the Rails app. It allows a user who has already authenticated through egghead-rails to be automatically signed in to egghead-next.

The way this works is:

1. egghead-rails adds a cookie with a fresh oauth access token when a user authenticates
2. egghead-next detects if there is an oauth access token available in a `.egghead.io` cookie
3. it calls `handleCookieBasedAccessTokenAuthentication` with the access token which in turn sets up the session and fetches fresh user data (in the same way as the standard auth flow)

### DIFF
--- a/src/context/viewer-context.tsx
+++ b/src/context/viewer-context.tsx
@@ -5,6 +5,7 @@ import get from 'lodash/get'
 import isEqual from 'lodash/isEqual'
 import isEmpty from 'lodash/isEmpty'
 import {useRouter} from 'next/router'
+import getAccessTokenFromCookie from '../utils/getAccessTokenFromCookie'
 
 export const auth = new Auth()
 
@@ -31,6 +32,7 @@ export const ViewerContext = React.createContext(defaultViewerContext)
 function useAuthedViewer() {
   const router = useRouter()
   const [viewer, setViewer] = React.useState()
+  const viewerId = get(viewer, 'id', null)
   const [loading, setLoading] = React.useState(true)
   const previousViewer = React.useRef(viewer)
 
@@ -43,17 +45,18 @@ function useAuthedViewer() {
   })
 
   React.useEffect(() => {
+    const authToken = getAccessTokenFromCookie()
     const queryHash = queryString.parse(window.location.hash)
     const accessToken = get(queryHash, 'access_token')
     const noAccessTokenFound = isEmpty(accessToken)
-    const viewerIsPresent = !isEmpty(viewer)
+    const viewerIsPresent = !isEmpty(viewerId)
 
     let viewerMonitorIntervalId: number | undefined
 
     const loadViewerFromStorage = async () => {
       console.log(`loading viewer from storage`)
       const newViewer: any = await auth.refreshUser()
-      if (!isEqual(newViewer, viewer)) {
+      if (!isEqual(newViewer.id, viewerId)) {
         setViewer(newViewer)
       }
       setLoading(() => false)
@@ -79,7 +82,19 @@ function useAuthedViewer() {
       }
     }
 
-    if (viewerIsPresent) {
+    if (authToken) {
+      auth
+        .refreshUser()
+        .then((refreshedViewer: any) => {
+          setViewer(refreshedViewer)
+          console.log(`refresh user from cookie auth token`)
+          setLoading(() => false)
+        })
+        .catch((error) => {
+          // if it is a 403, clear out the token and then call auth.monitor?
+          console.log('Catching this error: ', error.message)
+        })
+    } else if (viewerIsPresent) {
       loadViewerFromStorage()
       clearAccessToken()
     } else if (noAccessTokenFound) {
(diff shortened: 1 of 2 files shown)


src/utils/auth.ts
```diff
@@ -188,26 +210,24 @@ export default class Auth {
...
       const now: number = new Date().getTime()
 
-      const expiresInDays = Math.floor(
-        (Number(expiresAt) - now) / (60 * 60 * 24 * 1000),
-      )
+      const millisecondsInASecond = 1000
+      const expiresAt = Number(expiresInSeconds) * millisecondsInASecond + now
+
+      const millisecondsInADay = 60 * 60 * 24 * 1000
+      const expiresInDays = Math.floor((expiresAt - now) / millisecondsInADay)
```
jbranchaud, line 224: I struggled to figure out what was going on in this function and what some of the numbers were for. Once I made sense of it, I encoded my findings in variable names.

src/utils/auth.ts
```diff
@@ -104,6 +107,32 @@ export default class Auth {
     }
   }
 
+  handleCookieBasedAccessTokenAuthentication(
+    accessToken: string,
+    expiresInSeconds: string,
+  ) {
+    // handle any previous location redirects here
```
jbranchaud, line 114: On second thought, this is probably not where previous URL redirects will happen.

src/context/viewer-context.tsx
```diff
@@ -95,7 +114,7 @@ function useAuthedViewer() {
     }
 
     return clearUserMonitorInterval
-  }, [viewer])
+  }, [viewerId])
```
jbranchaud, line 117: `useEffect` doesn't do deep object comparison, so instead using the user's id in the dependency array.

src/context/viewer-context.tsx
```diff
@@ -79,7 +82,23 @@ function useAuthedViewer() {
       }
     }
 
-    if (viewerIsPresent) {
+    if (authToken) {
```
jbranchaud, line 85: This new case at the front of the if/then/else checks if there is an authToken (from a cookie) and then attempts to set up the session with it.

src/utils/auth.ts
```diff
@@ -104,6 +107,32 @@ export default class Auth {
...
+          resolve(user)
+        },
+        (error) => {
+          console.error(error)
+          this.logout()
+          reject(error)
+        },
+      )
+    })
+  }
```
jbranchaud, line 134: Extracted this bit of session set up logic from `handleAuthentication` so that either auth approach can call it.

jbranchaud: Also updated it so that `resolve` gets called with the fetched `user` instead of with the oauth `Token` object.

src/context/viewer-context.tsx
```diff
@@ -35,6 +35,7 @@ function useAuthedViewer() {
   const viewerId = get(viewer, 'id', null)
   const [loading, setLoading] = React.useState(true)
   const previousViewer = React.useRef(viewer)
+  const authToken = getAccessTokenFromCookie()
```
joelhooks, line 38: moved this out of the `useEffect` because if the token changes we definitely want to re-run the effect, eh?

src/context/viewer-context.tsx
```diff
@@ -82,22 +82,23 @@ function useAuthedViewer() {
...
     }
 
-    if (authToken) {
-      const sixtyDaysInSeconds = JSON.stringify(60 * 24 * 60 * 60)
-      auth
-        .handleCookieBasedAccessTokenAuthentication(
-          authToken,
-          sixtyDaysInSeconds,
-        )
+    const loadViewerFromToken = async () => {
```
joelhooks, line 85: the similar methods here are async where their underlying actions are async so I made it async here as well

src/context/viewer-context.tsx
```diff
@@ -82,22 +82,23 @@ function useAuthedViewer() {
...
+      await auth
+        .handleCookieBasedAccessTokenAuthentication(authToken)
         .then((refreshedViewer: any) => {
           setViewer(refreshedViewer)
-          console.log(`refresh user from cookie auth token`)
           setLoading(() => false)
         })
         .catch((error) => {
           // if it is a 403, clear out the token and then call auth.monitor?
+          // auth does a logout on any error currently -jh
```
joelhooks, line 94: I think `auth` does this under the hood so we should be good?

src/utils/auth.ts
```diff
@@ -23,6 +23,8 @@ export function getTokenFromCookieHeaders(serverCookies = '') {
   return {eggheadToken, loginRequired: eggheadToken.length <= 0}
 }
 
+const SIXTY_DAYS_IN_SECONDS = JSON.stringify(60 * 24 * 60 * 60)
```
joelhooks, line 26: seemed like a sensible default

src/context/viewer-context.tsx
jbranchaud: yep, I removed the catch and flattened this out https://github.com/eggheadio/egghead-next/pull/183/commits/a69c120641cad9feed568a4dfd94066483924326

joelhooks (approved):

PR #184, opened 2020-12-02
URL Swap: Update prod env vars to point to apex domain

jbranchaud:
This commit is to be merged after the DNS gets flipped, such that
egghead-next is running at egghead.io.


joelhooks:
i think we can close this in favor of the sitemaps PR

PR #185, opened 2020-12-02
Allow robots, update URLs to apex domain, and rebuild sitemaps

jbranchaud:
Another PR to be merged after this app is living at egghead.io.

Open the SEO floodgates


public/robots.txt
```diff
@@ -5,14 +5,12 @@ Allow: /
 Disallow: /discord
 Disallow: /_next
 Disallow: /purchase
-User-agent: *
-Disallow: /
```
jbranchaud, line 9: Allow robots to crawl the site

public/robots.txt
```diff
@@ -5,14 +5,12 @@ Allow: /
...
-Sitemap: https://next.egghead.io/tags-sitemap-6.xml
+Host: https://egghead.io
+Sitemap: https://egghead.io/sitemap.xml
+Sitemap: https://egghead.io/tags-sitemap-0.xml
+Sitemap: https://egghead.io/tags-sitemap-1.xml
+Sitemap: https://egghead.io/tags-sitemap-2.xml
+Sitemap: https://egghead.io/tags-sitemap-3.xml
+Sitemap: https://egghead.io/tags-sitemap-4.xml
+Sitemap: https://egghead.io/tags-sitemap-5.xml
+Sitemap: https://egghead.io/tags-sitemap-6.xml
```
jbranchaud, line 16: Update the Host and Sitemap directives to use `egghead.io`

public/robots.txt
joelhooks: this is actually generated!

next-sitemap.js
```diff
@@ -31,6 +27,7 @@ module.exports = {
       'https://next.egghead.io/tags-sitemap-4.xml',
       'https://next.egghead.io/tags-sitemap-5.xml',
       'https://next.egghead.io/tags-sitemap-6.xml',
+      'http://egghead-sitemaps.s3.amazonaws.com/sitemaps/sitemap.xml.gz',
```
joelhooks, line 30: egghead-rails will still generate the content sitemap, we should make sure it generates it with the appropriate URL

joelhooks: https://github.com/eggheadio/egghead-rails/blob/8419d61081cc5305be69285c165213ddd0a87174/config%2Fsitemap.rb

verified that the root url is configured in the sitemap config

joelhooks: plugin link for convenience: https://www.npmjs.com/package/next-sitemap#full-configuration-example

joelhooks:
I did this in another PR

PR #191, opened 2020-12-06
optimize lesson page loading

joelhooks:
There is a noticeable "lag" when you load a video resource:

https://next.egghead.io/lessons/react-implement-a-custom-material-ui-theme-on-the-server-using-a-higher-order-component

image

There's a clear gap in the waterfall chart (chrome dev tools, network tab) and a kind of dead zone before everything kicks in.

The initial lesson data is loaded during SSR, and then we load the media urls async with auth. It also refreshes the "current viewer" in the background.

dj at the records

### DIFF
--- a/src/components/pages/lessons/LessonInfo.tsx
+++ b/src/components/pages/lessons/LessonInfo.tsx
@@ -5,46 +5,45 @@ import {isEmpty, get} from 'lodash'
 import Markdown from 'react-markdown'
 import useCopyToClipboard from 'react-use/lib/useCopyToClipboard'
 import Eggo from '../../../components/images/eggo.svg'
-import {LessonResource} from 'types'
+import {useNextUpData} from 'hooks/use-next-up-data'
 
 type NextUpProps = {
   currentLessonSlug: string
-  data: {
-    list: {
-      lessons: LessonResource[]
-    }
-  }
+  url: string
 }
 
-const NextUp: FunctionComponent<NextUpProps> = ({data, currentLessonSlug}) => {
-  return data ? (
+const NextUp: FunctionComponent<NextUpProps> = ({url, currentLessonSlug}) => {
+  const {nextUpData} = useNextUpData(url)
+  return nextUpData ? (
     <ol>
       <span className="font-semibold">Lessons</span>
-      {data.list.lessons.map((lesson, index = 0) => {
-        return (
-          <li key={lesson.slug} className="py-2 pr-3">
-            <div className="flex">
-              <div className="flex items-center mr-2">
-                <div className="mr-1 text-xs text-cool-gray-400">
-                  {index + 1}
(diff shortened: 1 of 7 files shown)


src/context/viewer-context.tsx
```diff
@@ -56,6 +56,15 @@ function useAuthedViewer() {
     const loadViewerFromStorage = async () => {
       console.log(`loading viewer from storage`)
 
+      const localViewer = auth.getLocalUser()
+
+      if (localViewer) {
+        if (!isEqual(localViewer.id, viewerId)) {
+          setViewer(localViewer)
+        }
+        setLoading(() => false)
```
joelhooks, line 65: this was causing a delay, but if we have a viewer in localstorage might as well call it loaded optimistically and use that while a refresh happens in the background

src/pages/lessons/[slug].tsx
```diff
@@ -203,7 +203,10 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
         }}
       />
       <Head>
-        <script src="//cdn.bitmovin.com/player/web/8/bitmovinplayer.js" />
+        <script
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:206]: since the player component isn't onscreen it won't load the script, so I thought that this would load the bitmovin js earlier

src/utils/auth.ts
```diff
@@ -194,13 +194,13 @@ export default class Auth {
     return !expired
   }
 
-  refreshUser(loadFullUser = false) {
+  refreshUser(minimalUser = true) {
```
joelhooks, line 197: this logic was flipped

src/components/pages/lessons/LessonInfo.tsx
```diff
@@ -5,46 +5,45 @@ import {isEmpty, get} from 'lodash'
...
-      lessons: LessonResource[]
-    }
-  }
+  url: string
 }
 
-const NextUp: FunctionComponent<NextUpProps> = ({data, currentLessonSlug}) => {
-  return data ? (
+const NextUp: FunctionComponent<NextUpProps> = ({url, currentLessonSlug}) => {
+  const {nextUpData} = useNextUpData(url)
```
joelhooks, line 16: moving network calls into the components they are used in so we aren't loading everything up front on load by default

src/components/pages/lessons/LessonInfo.tsx
```diff
@@ -318,9 +320,9 @@ const LessonInfo: FunctionComponent<LessonInfoProps> = ({
           </div>
         </div>
       </div>
-      {nextUpData && (
+      {!playerState.matches('loading') && nextUpUrl && (
```
joelhooks, line 323: if the player is loading, hold off on the next up list. Hoping to provide the player with priority.

jbranchaud (approved):

PR #193, opened 2020-12-06
playlists use items

joelhooks:
nothing visual but sets up data for playlists to actually display properly

### DIFF
--- a/src/lib/playlists.ts
+++ b/src/lib/playlists.ts
@@ -12,6 +12,45 @@ export async function loadPlaylist(slug: string, token: string) {
         description
         image_thumb_url
         path
+        items {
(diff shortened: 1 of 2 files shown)


src/lib/playlists.ts
```diff
@@ -12,6 +12,45 @@ export async function loadPlaylist(slug: string, token: string) {
         description
         image_thumb_url
         path
+        items {
```
joelhooks, line 15: items is a "union type" so you need to specify types ad use fragments

https://github.com/eggheadio/egghead-rails/commit/6b105015eeb4134b49bb1549ecabf840f47af9ea

Made some backend changes to accomodate this on the api side

PR #195, opened 2020-12-07
store speed rate in cookies

DrShpongle:
https://share.getcloudapp.com/d5uPEvWA


src/components/EggheadPlayer/use-egghead-player.ts
```diff
@@ -225,12 +225,22 @@ function onVideoQualityChanged() {}
 
 function onSubtitleChange() {}
 
+function onPlaybackRateChange(rate: number) {
+  cookies.set('egghead-playback-rate', rate)
```
joelhooks, line 229: a cookie like this should probably be a `const` and perhaps stored in config

joelhooks (approved):

PR #202, opened 2020-12-09
Video player tweaks

DrShpongle:
- spinner loader added
- mobile styles added to the NextUp overlay

Screen Recording 2020-12-09 at 08 49 05 PM


joelhooks (approved):
looks great!

PR #204, opened 2020-12-10
Store cookie for referral when present in URL

jbranchaud:
Grab the rc value out of the URL, request that egghead-rails sign the
value, and then add that value to a permanent cookie for a later
checkout session.

### DIFF
--- a/src/hooks/use-token-signer.tsx
+++ b/src/hooks/use-token-signer.tsx
@@ -41,6 +41,13 @@ function useTokenSigner(): void {
     router.query.af,
   )
 
+  const removeQueryFromUrl = (paramName: string) => {
+    const {[paramName]: _paramToRemove, ...updatedQuery} = router.query
+    router.push({pathname: router.pathname, query: updatedQuery}, undefined, {
+      shallow: true,
+    })
+  }
+
   React.useEffect(() => {
     const requestSignedReferralToken = async (referralToken: string) => {
       const existingReferralToken = cookie.get('rc')
@@ -59,7 +66,7 @@ function useTokenSigner(): void {
 
             createPermanentCookie('rc', newReferralToken)
 
-            // TODO: Remove `rc` from URL query params, similar to `af`.
+            removeQueryFromUrl('rc')
 
             return newReferralToken
           })
@@ -74,7 +81,7 @@ function useTokenSigner(): void {
     }
 
     return () => {}
-  }, [referralQueryParam])
+  }, [referralQueryParam, removeQueryFromUrl])
 
   React.useEffect(() => {
     const requestSignedAffiliateToken = async (affiliateToken: string) => {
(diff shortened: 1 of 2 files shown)


src/context/viewer-context.tsx
```diff
@@ -37,6 +38,8 @@ function useAuthedViewer() {
   const previousViewer = React.useRef(viewer)
   const authToken = getAccessTokenFromCookie()
 
+  useTokenSigner()
```
jbranchaud, line 41: I figured the viewer context was as good of place as any to handle a referral token that eventually gets tied to a user's (viewer's) subscription.

src/hooks/use-token-signer.tsx
```diff
@@ -0,0 +1,65 @@
...
+  param: undefined | string | string[],
+): string | undefined {
+  if (Array.isArray(param)) {
+    return param[0]
+  } else {
+    return param
+  }
+}
+
+function useTokenSigner(): void {
```
jbranchaud, line 19: If the referral token is in the URL's query params, grab it, request that the server sign it, then stick the signed value in a cookie.

If a signed referral token is already present, don't do any of this.

src/hooks/use-token-signer.tsx
```diff
@@ -0,0 +1,65 @@
...
+        return http
+          .post(
+            `${AUTH_DOMAIN}/api/v1/sign_referral_token`,
+            {rc: referralToken},
+            {},
+          )
+          .then(({data}) => {
+            const newReferralToken = data.signed_referral_token
+            const twenty_years = 365 * 20
+            const permanent = twenty_years // this is what permanent means for Rails cookies
```
jbranchaud, line 42: When referral token cookies get set on the Rails side, they are _permanent_ cookies, which for Rails means 20 years. This emulates that expiration timeframe.

src/hooks/use-token-signer.tsx
joelhooks: nitpick (typing): return `void` is redundant and always (safely) inferred

src/hooks/use-token-signer.tsx
```diff
@@ -0,0 +1,65 @@
...
+            return null
+          })
+      }
+    }
+
+    if (!!referralQueryParam) {
+      requestSignedReferralToken(referralQueryParam)
+    }
+
+    return () => {}
```
joelhooks, line 61: nitpick (react): there's no need to return a `noop` - but if you are going to we have a `noop` you can import

src/hooks/use-token-signer.tsx
```diff
@@ -0,0 +1,65 @@
...
+            return newReferralToken
+          })
+          .catch((error) => {
+            return null
+          })
+      }
+    }
+
+    if (!!referralQueryParam) {
+      requestSignedReferralToken(referralQueryParam)
```
joelhooks, line 58: can we also clear the query string?

jbranchaud: For sure, was just about to do that on another PR, but I'll add it here.

joelhooks (approved):

PR #211, opened 2020-12-15
polish final paragraphs

filiphric:


joelhooks:
@filiphric we require a CLA to be signed for merging PRs. It's pretty easy if you follow the instructions above!

PR #214, opened 2020-12-16
merge course and playlist layouts

joelhooks:
we had a lot of duplication and missing functionality in the course and playlist page components so this merges them together and brings playlists in line with courses.

Required some graphql api additions as well to capture the missing data for playlists.


src/components/pages/courses/course-page-layout.tsx
```diff
@@ -0,0 +1,220 @@
...
+import removeMarkdown from 'remove-markdown'
+
+type CoursePageLayoutProps = {
+  lessons: any
+  course: any
+  ogImageUrl: string
+}
+
+const CoursePageLayout: React.FunctionComponent<CoursePageLayoutProps> = ({
+  lessons,
```
joelhooks, line 20: lessons get passed in because with playlists we are loading `items` and not `lessons` specifically.

src/components/pages/courses/instructor-profile.tsx
```diff
@@ -0,0 +1,38 @@
+import Link from 'next/link'
+import * as React from 'react'
+import Markdown from 'react-markdown'
+
+const InstructorProfile: React.FunctionComponent<{
```
joelhooks, line 5: originally just started breaking out the shared components and then did the big one above. these ar5e still nicer here...

src/data/courseDependencies.ts
```diff
@@ -0,0 +1,6 @@
+import dependencies from './courseDependenciesData'
```
joelhooks, line 1: moved this into the `src` folder and copied over the `js` file instead of using JSON so it is the same and "just works" with a cut and paste.

src/pages/courses/[slug].tsx
```diff
@@ -285,11 +30,9 @@ export default Course
 export const getServerSideProps: GetServerSideProps = async ({res, params}) => {
   res.setHeader('Cache-Control', 's-maxage=120, stale-while-revalidate')
   const course = params && (await loadCourse(params.slug as string))
-  const dependencies = getDependencies(course.slug)
```
Comment (joelhooks) [inline src/pages/courses/[slug].tsx:288]: there was no compelling reason to do this here since it is "compiled" data anyway so I moved it into the client side render method

src/pages/playlists/[slug].tsx
```diff
@@ -178,9 +35,8 @@ export const getServerSideProps: GetServerSideProps = async ({
   req,
   params,
 }) => {
-  const {eggheadToken} = getTokenFromCookieHeaders(req.headers.cookie as string)
-  const playlist =
-    params && (await loadPlaylist(params.slug as string, eggheadToken))
+  res.setHeader('Cache-Control', 's-maxage=1, stale-while-revalidate')
```
Comment (joelhooks) [inline src/pages/playlists/[slug].tsx:38]: caching good.

PR #216, opened 2020-12-16
refactor Transcript stuff

DrShpongle:
Fixes the Transcript tab render condition so that such situations are avoided now:

Image 2020-12-16 at 7 58 07 PM


src/hooks/use-enhanced-transcript.ts
```diff
@@ -0,0 +1,8 @@
+import fetcher from 'utils/fetcher'
+import useSWR from 'swr'
+
+export const useEnhancedTranscript = (url: string) => {
+  const {data: transcriptData} = useSWR(url, fetcher)
+  const enhancedTranscript = transcriptData?.text
+  return {enhancedTranscript}
```
joelhooks, line 7:
```suggestion
  return enhancedTranscript
```

src/pages/lessons/[slug].tsx
```diff
@@ -95,6 +96,9 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
     free_forever,
   } = lesson
 
+  const {enhancedTranscript} = useEnhancedTranscript(transcript_url)
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:99]:
```suggestion
  const enhancedTranscript = useEnhancedTranscript(transcript_url)
```

src/hooks/use-enhanced-transcript.ts
```diff
@@ -0,0 +1,8 @@
+import fetcher from 'utils/fetcher'
+import useSWR from 'swr'
+
+export const useEnhancedTranscript = (url: string) => {
+  const {data: transcriptData} = useSWR(url, fetcher)
+  const enhancedTranscript = transcriptData?.text
+  return enhancedTranscript
```
joelhooks, line 7: @DrShpongle I removed the `{}` from this, no need to return an object if we are only interested in the transcript. I think technically this could just return `data?.text`

src/pages/talks/[slug].tsx
```diff
@@ -119,6 +121,9 @@ const Talk: FunctionComponent<LessonProps> = ({initialLesson}) => {
     slug,
   } = lesson
 
+  const {enhancedTranscript} = useEnhancedTranscript(transcript_url)
```
Comment (joelhooks) [inline src/pages/talks/[slug].tsx:124]:
```suggestion
  const enhancedTranscript = useEnhancedTranscript(transcript_url)
```

joelhooks (approved):

PR #222, opened 2020-12-18
Fetch pro plan from Rails API, display as pricing

jbranchaud:


src/pages/pricing/index.tsx
```diff
@@ -12,10 +13,17 @@ import emailIsValid from 'utils/email-is-valid'
...
 
-  if (!process.env.NEXT_PUBLIC_STRIPE_ANNUAL_PRICE_ID)
-    throw new Error('no annual price to load 😭')
+  const {data: pricingData} = await axios.get(
+    `${process.env.NEXT_PUBLIC_AUTH_DOMAIN}/api/v1/pricing`,
+  )
+  // TODO: Is finding the annual plan by name reliable?
+  const proYearly: {price_id: string} | undefined = pricingData.plans.find(
+    (plan: any) => plan.name === 'Pro Yearly',
+  )
```
jbranchaud, line 22: Not sure how standard the naming is

joelhooks: no, using lodash syntax:

```js
find(plans, {interval: 'year', type: 'individual', tier: 'pro'})
```

tested here just to verify: https://codesandbox.io/s/beautiful-lamport-4twub?file=/src/App.js

src/pages/pricing/index.tsx
```diff
@@ -12,10 +13,17 @@ import emailIsValid from 'utils/email-is-valid'
 export const getServerSideProps: GetServerSideProps = async function ({res}) {
   res.setHeader('Cache-Control', 's-maxage=1, stale-while-revalidate')
```
joelhooks, line 14: we should probably remove this since it will edge cache dynamic prices and we probably want a "no cache" situation

src/pages/pricing/index.tsx
```diff
@@ -12,10 +13,17 @@ import emailIsValid from 'utils/email-is-valid'
...
+  )
+  // TODO: Is finding the annual plan by name reliable?
+  const proYearly: {price_id: string} | undefined = pricingData.plans.find(
+    (plan: any) => plan.name === 'Pro Yearly',
+  )
+
+  if (!proYearly?.price_id) throw new Error('no annual price to load 😭')
 
-  const price = await loadPrice(process.env.NEXT_PUBLIC_STRIPE_ANNUAL_PRICE_ID)
+  const price = await loadPrice(proYearly.price_id)
```
joelhooks, line 26: imo the above should be moved into `loadPrice`

technically speaking we don't really need to load both from egghead api and the stripe api in this situation since the data we need is in the egghead response, **but** the data in the egghead response is all hand typed where Stripe is the legit "source of truth" so doing the double hop might make sense.

trade-off is speed to load.

src/pages/pricing/index.tsx
joelhooks: I went ahead and updated this to just load from the egghead api

src/components/pricing/select-plan.tsx
```diff
@@ -79,7 +79,11 @@ const SelectPlan: React.FunctionComponent<SelectPlanProps> = ({
                 One low price...
               </p>
               <div className="mt-4 flex items-center justify-center text-5xl leading-none font-extrabold text-gray-900">
-                <span>${price}</span>
+                {prices ? (
+                  <span>${prices.annualPrice.price}</span>
+                ) : (
+                  <span>$ ---</span>
```
joelhooks, line 85: this makes the layout shift, which isn't super nice, but probably fixable

src/pages/pricing/index.tsx
```diff
@@ -2,28 +2,13 @@ import * as React from 'react'
...
-import {GetServerSideProps} from 'next'
-import {loadPrice} from 'lib/stripe/price'
+import {loadPrices} from 'lib/prices'
 import stripeCheckoutRedirect from 'api/stripe/stripe-checkout-redirect'
 import SelectPlan from 'components/pricing/select-plan'
 import EmailForm from 'components/pricing/email-form'
 import emailIsValid from 'utils/email-is-valid'
 import {track} from 'utils/analytics'
-
-export const getServerSideProps: GetServerSideProps = async function ({res}) {
```
joelhooks, line 13: dump this altogether in favor of a faster paint and JIT price loading

joelhooks (approved):

PR #223, opened 2020-12-19
forward query params for discounts

joelhooks:
the pricing API accepts `en` and `dc` params for loading coupons and applying them to the display price.

### DIFF
--- a/src/components/facebook-pixel.tsx
+++ b/src/components/facebook-pixel.tsx
@@ -10,17 +10,17 @@ export const FacebookPixel = function () {
     let fb: any
 
     function onRouteChange(url: string) {
-      fb.pageView()
+      fb?.pageView()
     }
 
     import('react-facebook-pixel')
       .then((module) => (fb = module.default))
       .then(() => {
-        fb.init(FACEBOOK_PIXEL_ID, {
+        fb?.init(FACEBOOK_PIXEL_ID, {
           autoConfig: true,
           debug: true,
         })
-        fb.pageView()
+        fb?.pageView()
       })
 
     router.events.on('routeChangeComplete', onRouteChange)
--- a/src/components/pricing/select-plan.tsx
+++ b/src/components/pricing/select-plan.tsx
@@ -79,7 +79,7 @@ const SelectPlan: React.FunctionComponent<SelectPlanProps> = ({
                 One low price...
               </p>
               <div className="mt-4 flex items-center justify-center text-5xl leading-none font-extrabold text-gray-900">
-                {prices ? (
+                {prices.annualPrice ? (
                   <span>${prices.annualPrice.price}</span>
                 ) : (
                   <span>$ ---</span>
@@ -93,14 +93,14 @@ const SelectPlan: React.FunctionComponent<SelectPlanProps> = ({
                 <div className="rounded-md shadow">
                   <button
                     onClick={onClickCheckout}
-                    disabled={!prices || viewer?.is_pro}
+                    disabled={!prices.annualPrice || viewer?.is_pro}
                     className={`${
                       !prices || loading || viewer?.is_pro
                         ? 'opacity-40'
                         : 'opacity-100'
                     } w-full flex items-center justify-center px-5 py-3 border border-transparent text-base leading-6 font-medium rounded-md text-white bg-gray-900 hover:bg-gray-800 focus:outline-none focus:shadow-outline transition duration-150 ease-in-out`}
                   >
-                    {prices && !loading
+                    {prices.annualPrice && !loading
                       ? viewer?.is_pro
                         ? `Already a Member!`
                         : `Get Access`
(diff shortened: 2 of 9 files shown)


src/components/facebook-pixel.tsx
```diff
@@ -10,17 +10,17 @@ export const FacebookPixel = function () {
     let fb: any
 
     function onRouteChange(url: string) {
-      fb.pageView()
+      fb?.pageView()
```
joelhooks, line 13: this was throwing an error sometimes

src/components/pricing/select-plan.tsx
```diff
@@ -79,7 +79,7 @@ const SelectPlan: React.FunctionComponent<SelectPlanProps> = ({
                 One low price...
               </p>
               <div className="mt-4 flex items-center justify-center text-5xl leading-none font-extrabold text-gray-900">
-                {prices ? (
+                {prices.annualPrice ? (
```
joelhooks, line 82: we always have prices, but sometimes it's empty

src/lib/prices.ts
```diff
@@ -1,5 +1,5 @@
 import {find} from 'lodash'
-import axios from 'axios'
+import axios from 'utils/configured-axios'
```
joelhooks, line 2: `configured-axios` reads the auth token from the cookie and sends it in the header

src/lib/prices.ts
```diff
@@ -8,9 +8,17 @@ type Prices = {
...
+export async function loadPrices(
+  params: {
+    en?: string
+    dc?: string
+  } = {},
+): Promise<Prices> {
   const {data: pricingData} = await axios.get(
     `${process.env.NEXT_PUBLIC_AUTH_DOMAIN}/api/v1/pricing`,
+    {
+      params,
```
joelhooks, line 20: forward the params, the api already expected them

src/hooks/use-pricing.ts
```diff
@@ -0,0 +1,22 @@
+import * as React from 'react'
+import {loadPrices} from 'lib/prices'
+import queryString from 'query-string'
+
+export type Prices = {
+  annualPrice?: any
+}
+
+export const usePricing = () => {
```
joelhooks, line 9: this hook loads prices from the egghead api

src/components/pricing/select-plan.tsx
```diff
@@ -74,41 +64,7 @@ const SelectPlan: React.FunctionComponent<SelectPlanProps> = ({
...
-                    {prices && !loading
-                      ? viewer?.is_pro
-                        ? `Already a Member!`
-                        : `Get Access`
-                      : '--'}
-                  </button>
-                </div>
-              </div>
-            </div>
+            {children}
```
joelhooks, line 67: felt like this section would probably be fine composed in instead of passing down props

src/pages/pricing/index.tsx
```diff
@@ -80,10 +72,46 @@ const Pricing: FunctionComponent<PricingProps> = () => {
           </div>
         )}
         {!needsEmail && (
-          <SelectPlan prices={data} onClickCheckout={onClickCheckout} />
+          <SelectPlan>
+            <div className="py-8 px-6 text-center bg-gray-50 lg:flex-shrink-0 lg:flex lg:flex-col lg:justify-center lg:p-12">
```
joelhooks, line 76: the composition makes this files a little chunkier, but also makes it easier to work with since there's not a lot of reason to visit the `SelectPrice` component

src/pages/pricing/index.tsx
```diff
@@ -80,10 +72,46 @@ const Pricing: FunctionComponent<PricingProps> = () => {
...
+                  </button>
+                </div>
+              </div>
+            </div>
+          </SelectPlan>
         )}
-        {data && needsEmail && (
-          <EmailForm priceId={data.annualPrice.price_id} />
+        {prices.annualPrice && needsEmail && (
+          <EmailForm priceId={prices.annualPrice.price_id} />
```
joelhooks, line 114: unlike `SelectPrice` the form is more stateful and only take a string value as a prop so it didn't get broken down into composed components (yet!)

src/utils/auth.ts
```diff
@@ -17,6 +17,19 @@ export const ACCESS_TOKEN_KEY = 'eh_token_2020_11_22'
 export const EXPIRES_AT_KEY = 'eh_token_expiration'
 export const VIEWING_AS_USER_KEY = 'eh_viewing_as_user_2020_11_22'
 
+type AuthorizationHeader = {
+  Authorization: string
+}
+
+export const getAuthorizationHeader = () => {
```
joelhooks, line 24: we are using this pattern around the app so centralized it into a method

src/utils/configured-axios.ts
```diff
@@ -1,8 +1,9 @@
 import axios from 'axios'
+import {getAuthorizationHeader} from './auth'
 
 axios.interceptors.request.use(
```
joelhooks, line 4: preconfigured `axios` that sends auth headers if they are available

PR #224, opened 2020-12-19
fixes #206 handle nested collections

### LINKED ISSUE #206: Wrong link on "Start watching" button
On article https://next.egghead.io/playlists/composing-closures-and-callbacks-in-javascript-1223 , "Start watching" href leads to https://next.egghead.io/lessons/undefined

joelhooks:
with a "collection of collections" we weren't displaying them appropriately

This looks to see if we have inner collections and use those to present each inner collection in "sections"

### DIFF
--- a/src/pages/playlists/[slug].tsx
+++ b/src/pages/playlists/[slug].tsx
@@ -13,16 +13,18 @@ type PlaylistProps = {
 
 const Playlist: FunctionComponent<PlaylistProps> = ({playlist}) => {
   const initialData = playlist
-  const {data} = useSWR(playlist.url, fetcher, {initialData})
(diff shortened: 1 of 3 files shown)


src/pages/playlists/[slug].tsx
```diff
@@ -13,16 +13,18 @@ type PlaylistProps = {
 
 const Playlist: FunctionComponent<PlaylistProps> = ({playlist}) => {
   const initialData = playlist
-  const {data} = useSWR(playlist.url, fetcher, {initialData})
+  const {data} = useSWR(playlist.url, fetcher)
```
Comment (joelhooks) [inline src/pages/playlists/[slug].tsx:16]: this "initial data" call made it work very strangely and links to other playlists would load the first playlist. Removing this and combining them like below fixed it.

PR #230, opened 2020-12-22
WIP - attempted fix of the double-logout bug

jbranchaud:
This commit doesn't fix the issue. I'm stuck on where to go from here.

The viewer-context involves a mechanism that tries to resolve a missing
user session. If the authToken is present when it mounts or during any
rerender of the viewer-context, it will request fresh user data.

What I'm noticing when hitting the logout endpoint, is that while a
request to `DELETE /api/users/session` goes out to clear the authToken
cookie, the viewer-context useEffect gets triggered with that authToken,
so by the time the cookie gets deleted and the local storage gets
cleared out, the refreshUser call returns with data to repopulate local
storage.

I tried introducing some state to represent when a logout is being
performed to avoid this timing issue, but the useEffect is still
triggering the refreshUser.

### DIFF
--- a/src/context/viewer-context.tsx
+++ b/src/context/viewer-context.tsx
@@ -39,6 +39,7 @@ function useAuthedViewer() {
   const [loading, setLoading] = React.useState(true)
   const previousViewer = React.useRef(viewer)
   const authToken = getAccessTokenFromCookie()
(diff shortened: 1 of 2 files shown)


joelhooks:
https://github.com/eggheadio/egghead-next/pull/231 I think this does the trick.

PR #231, opened 2020-12-22
breaking up is hard to do (fix /logout)

joelhooks:
We've had a fun issue where clicking logout didn't actually log you out so this is just an exploration of fixing that with the "messy" version committed.

Will clean up, but it was interesting for discussion.


src/context/viewer-context.tsx
```diff
@@ -56,6 +56,21 @@ function useAuthedViewer() {
     const accessToken = get(queryHash, 'access_token')
     const noAccessTokenFound = isEmpty(accessToken)
     const viewerIsPresent = !isEmpty(viewerId)
+    const authToken = getAccessTokenFromCookie()
+
+    console.log({loggingOut})
+
+    if (loggingOut) {
```
joelhooks, line 63: since `loggingOut` is a state, we can capture it in the effect and make sure we aren't running anything else if that's what we are doing right now

src/context/viewer-context.tsx
```diff
@@ -56,6 +56,21 @@ function useAuthedViewer() {
     const accessToken = get(queryHash, 'access_token')
     const noAccessTokenFound = isEmpty(accessToken)
     const viewerIsPresent = !isEmpty(viewerId)
+    const authToken = getAccessTokenFromCookie()
```
joelhooks, line 59: the auth token doesn't need to trigger a rerun of this effect?

src/context/viewer-context.tsx
```diff
@@ -135,21 +150,14 @@ function useAuthedViewer() {
...
-            router.reload()
-          }
-        }
-        const logoutCallback = callback || defaultCallback
-
-        auth.logout()
-        logoutCallback()
+      logout: async () => {
+        console.log('LOGOUT REQUESTED')
+        setLoggingOut(true)
```
joelhooks, line 160: sets the state and bails, we perform the side effects in the effect above

src/context/viewer-context.tsx
```diff
@@ -135,21 +150,14 @@ function useAuthedViewer() {
     }
 
     return clearUserMonitorInterval
-  }, [viewerId, authToken])
+  }, [viewerId, loggingOut])
```
joelhooks, line 153: if we are `loggingOut` let's rerun the effect.

src/pages/logout.tsx
```diff
@@ -1,20 +1,12 @@
...
 
   React.useEffect(() => {
-    const redirectToIndex = () => {
-      if (typeof window !== 'undefined') {
-        router.replace('/')
-      }
-    }
-
-    logout(redirectToIndex)
+    logout()
```
joelhooks, line 9: default behavior is the redirect to index

src/utils/auth.ts
```diff
@@ -40,6 +40,7 @@ const SIXTY_DAYS_IN_SECONDS = JSON.stringify(60 * 24 * 60 * 60)
 
 export default class Auth {
   eggheadAuth: OAuthClient
+  loggingOut: boolean = false
```
joelhooks, line 43: this can probably go, no need for this internal state I don't think

src/utils/auth.ts
```diff
@@ -112,8 +113,15 @@ export default class Auth {
   }
 
   logout() {
-    track('logged out')
-    return this.clearLocalStorage()
+    this.loggingOut = true
+    return new Promise((resolve) => {
```
joelhooks, line 117: made everything promises

src/utils/auth.ts
```diff
@@ -176,27 +183,42 @@ export default class Auth {
...
-    if (typeof localStorage !== 'undefined') {
-      localStorage.removeItem(ACCESS_TOKEN_KEY)
-      localStorage.removeItem(EXPIRES_AT_KEY)
-      localStorage.removeItem(USER_KEY)
-      localStorage.removeItem(VIEWING_AS_USER_KEY)
-    }
+    return new Promise((resolve) => {
+      const removeLocalStorage = async () => {
+        cookie.remove(ACCESS_TOKEN_KEY, {
+          domain: process.env.NEXT_PUBLIC_AUTH_COOKIE_DOMAIN,
```
joelhooks, line 197: this one is a big **gotcha** because it has to have the domain or it doesn't actually clear it :trollface:

src/context/viewer-context.tsx
```diff
@@ -56,6 +56,21 @@ function useAuthedViewer() {
...
+    const authToken = getAccessTokenFromCookie()
+
+    console.log({loggingOut})
+
+    if (loggingOut) {
+      const doLogout = async () => {
+        console.log('do the actual logout')
+        await auth.logout()
+        if (typeof window !== 'undefined') {
+          window.location.href = '/'
```
joelhooks, line 68: we want a hard refresh not a router change

joelhooks (commented):
fundamentally this should be a state machine and not a wacky mess IMO

src/context/viewer-context.tsx
```diff
@@ -37,11 +37,11 @@ function useAuthedViewer() {
   const [viewer, setViewer] = React.useState()
   const viewerId = get(viewer, 'id', null)
   const [loading, setLoading] = React.useState(true)
+  const [loggingOut, setLoggingOut] = React.useState(false)
```
joelhooks, line 40: this is a state of the viewer context, so treat it accordingly

src/context/viewer-context.tsx
```diff
@@ -37,11 +37,11 @@ function useAuthedViewer() {
...
   const previousViewer = React.useRef(viewer)
-  const authToken = getAccessTokenFromCookie()
 
   useTokenSigner()
-  useAffiliateAssigner(viewerId, authToken)
+  useAffiliateAssigner(viewerId, getAccessTokenFromCookie())
```
joelhooks, line 44: really the auth token lookup could be service locator style inside of the hook instead of passed in, but I wanted to make the auth token a non-dependency for the useEffect below

src/utils/auth.ts
joelhooks: I think the way this cookie was being removed async was causing heartache and doing it this way is better.

jbranchaud: I thought it was a cool use of a serverless function, but yeah, it sure seemed like it was at least in part to blame for the timing issues.

src/utils/auth.ts
```diff
@@ -176,27 +177,32 @@ export default class Auth {
...
-            this.logout()
-            reject(error)
+            this.logout().then(() => reject(error))
           },
         )
       }
     })
   }
 
   clearLocalStorage() {
```
jbranchaud, line 187: What about calling this `clearSessionData` since it is wiping out session data from both a cookie and local storage?

joelhooks: I'm all for improvements, but would probably defer them for now. I think this needs to be deeply reconsidered and probably deserves some xstate.

src/context/viewer-context.tsx
```diff
@@ -56,6 +56,18 @@ function useAuthedViewer() {
...
+
+    if (loggingOut) {
+      const doLogout = async () => {
+        await auth.logout()
+        if (typeof window !== 'undefined') {
+          window.location.href = '/'
+        }
+      }
+      doLogout()
+      return
```
jbranchaud, line 69: How does this return impact the unmounting of the useEffect? I guess this is overriding the bottom of the effect where it:

```javascript
return clearUserMonitorInterval
```

joelhooks: it won't be an issue because the interval is never created to be clear since we bail early on this run. Any previous instances will be cleared when its component instance unmounts and before this effect is executed.

jbranchaud:
This feels good enough for now, especially if we can revisit it with a state machine makeover later.

I'll still point out that there appears to be a double-render happening, where the first render still has session state.

joelhooks:
@jbranchaud I think you'd solve that by reconsidering the layout of the `/logout` route. It's gonna display, show that, logout, and redirect (if I'm seeing it correctly)

jbranchaud (approved):

joelhooks:
(the header is in `AppLayout` which is applied by `_app`)

PR #233, opened 2020-12-24
Volume rate persists in cookie now

DrShpongle:


src/components/EggheadPlayer/use-egghead-player.ts
```diff
@@ -229,11 +229,19 @@ function onPlaybackRateChange(rate: number) {
   cookies.set('egghead-playback-rate', rate)
 }
 
+function onVolumeChange(targetVolume: number) {
+  cookies.set('egghead-volume-rate', targetVolume)
```
joelhooks, line 233: these could all be stored in a single egg-player-prefs cookie

joelhooks (approved):

PR #245, opened 2020-12-29
Include redirect URL in stripe checkout metadata

jbranchaud:
This will allow the post-checkout magic link auth flow to know where to
redirect the user so that they can get back to where they left off (e.g.
the video they were about to watch).


src/pages/pricing/index.tsx
```diff
@@ -117,11 +119,30 @@ const Pricing: FunctionComponent<PricingProps> = () => {
...
+export async function getServerSideProps(context: any) {
+  let redirectURL = null
+  try {
+    const referer = context.req.headers.referer
+    const refererURL = new URL(referer)
+    if (refererURL.origin === process.env.NEXT_PUBLIC_REDIRECT_URI)
+      redirectURL = referer
+  } catch (_e) {
+    // no redirect URL from referer
+  }
```
jbranchaud, line 141: In the context of `getServerSideProps` we can get the referer URL and if it is an egghead-next URL, then it will be treated as a redirect URL to be passed to `stripeCheckoutRedirect`.

joelhooks:
this might be redundant:

https://github.com/eggheadio/egghead-next/blob/main/src/pages/redirect.tsx

I started storing a "last resource" cookie and changed the redirect_uri to this page which looks at that cookie or shifts to root.

Currently there is only 1 situation where a viewer would need to redirect and it's if they were watching a video. 

`useLastResource` will also display "opt in" links in key spots if the viewer does other things

I know we'd talked about this and using the checkout meta, but I wanted to stop the sadness and the cookie approach was easy. It fails in the "uses a different browser" scenario...

jbranchaud:
> I know we'd talked about this and using the checkout meta, but I wanted to stop the sadness and the cookie approach was easy. It fails in the "uses a different browser" scenario...

An option for this would be for the server to set the cookie if it knows about a redirect uri. That'd be compatible with what you already have working for this, but still allow the magic link to tell other browsers and computers where to redirect to.

If you're up for this addition, this PR could go in as is and it'd require pretty minimal work on the server (ie. creating an all-domain cookie).

src/pages/pricing/index.tsx
joelhooks: this is the only part of this I'm not content with. I think we could consider using `useLastResource` to get the redirectURL if it exists.

The problem with this approach is that it is indiscriminate, and in my experience users don't want to be redirected to the last referer and instead the last "point of action" which in our case is always some resource.

I think if we went that route, I'd be fine with the approach described in the [other comment](https://github.com/eggheadio/egghead-next/pull/245#issuecomment-752706204)

joelhooks (approved):

PR #251, opened 2021-01-03
makes uidevtools optional

joelhooks:
This adds `@ui-devtools/tailwind` as an `optinalDependencies` item with an optional/dynamic import

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,8 @@
     "make-tags-sitemap": "NODE_ENV=production node tags-sitemap.js",
     "storybook": "start-storybook -p 6006",
     "build-storybook": "build-storybook",
-    "lint": "eslint --ext ts --ext tsx --fix src/"
+    "lint": "eslint --ext ts --ext tsx --fix src/",
+    "uidevtools": "devtools-server -c ./tailwind.config.js"
   },
   "dependencies": {
     "@emotion/babel-preset-css-prop": "^10.2.1",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -173,6 +174,9 @@
     "ts-node": "^9.1.1",
     "typescript": "^4.1.3"
   },
+  "optionalDependencies": {
```
joelhooks, line 177: this can fail and it's fine, so if you have access to the repo you get the sauce, and if not, no biggie

src/pages/_app.tsx
```diff
@@ -18,6 +18,14 @@ import {FacebookPixel} from 'components/facebook-pixel'
 import {Ahoy} from 'components/ahoy'
 import {CioProvider} from 'hooks/use-cio'
 
+let UIDevtools: React.FunctionComponent = ({children}) => <>{children}</>
+
+if (process.env.NEXT_PUBLIC_USE_UIDEVTOOLS === 'true') {
```
joelhooks, line 23: we need to explicitly turn ui-devtools on in `.env.local`

src/pages/_app.tsx
```diff
@@ -18,6 +18,14 @@ import {FacebookPixel} from 'components/facebook-pixel'
...
+  import('@ui-devtools/tailwind').then((module: any) => {
```
joelhooks, line 24: sprinkle on some dynamic imports and there ya go!

joelhooks:
This isn't working. `yarn install` on vercel still breaks (wtf?) with missing optional dependencies and running `yarn install --ignore-optional` breaks because some optional dependencies aren't in fact optional (deeper internally, it fails on `sharp`)

joelhooks:
pipe dreams?

PR #253, opened 2021-01-03
load the schedule from airtable

joelhooks:
instead of events being hard coded we can load them from a database stored in Airtable


src/pages/api/schedule.ts
```diff
@@ -0,0 +1,48 @@
...
+
+      const events = records.reduce((acc: any, potentialEvent: any) => {
+        const {expirationDate, ...rawEvent} = potentialEvent.fields
+        const now = new Date().getTime()
+        const eventExpiration = new Date(expirationDate).getTime()
+        const expired = now > eventExpiration
+
+        if (expired) return acc
+
+        return [
```
joelhooks, line 28: if it's not expired add the event

PR #258, opened 2021-01-05
stop autoplaying video on home page

vojtaholik:
this PR stops video on home page from autoplaying even if autoplay cookie is set to true.


joelhooks:
I was actually thinking about autoplay and muted by default for these 😅

PR #264, opened 2021-01-07
course page tuning and ratings

joelhooks:
image


src/components/icons/star.jsx
```diff
@@ -0,0 +1,51 @@
+import React from 'react'
+import SvgIcon from './svg-icon'
+import {css} from 'emotion'
+
+class Star extends React.Component {
```
joelhooks, line 5: copy=pasta from egghead-rails

PR #267, opened 2021-01-07
Include last resource in stripe checkout metadata

jbranchaud:
Per the conversation in https://github.com/eggheadio/egghead-next/pull/245, this PR replaces the `redirectUrl` with serialized lastResource as the value that gets passed to Stripe Checkout metadata.

This, in conjunction with some server-side changes, allows the magic link to encode a redirect to the last used resource cross-browser and cross-machine.


jbranchaud:
hmm, only passing the title and path (which I thought would be enough to get the user where they need to go) leads to not being able to properly render the "Keep Watching" block:

Screen Shot 2021-01-07 at 2 10 00 PM

That is, the image and author is missing.

The limitation that led to only passing the title and path was running up against Stripe's metadata restriction of 500 characters (or something around there). Perhaps the resource type and ID can be passed to Stripe instead and then the server can reconstitute that later after checkout and auth.

joelhooks:
@jbranchaud ya you can pass `resource_slug` and `resource_type` which should let you load it back up.

joelhooks:
I feel like the cookie approach is adequate for now so I'm going to close this.

PR #276, opened 2021-01-10
course in search results

joelhooks:
image


src/components/pages/lessons/overlay/recommend-next-step-overlay/lesson-01.js
```diff
@@ -0,0 +1,120 @@
+export const cloudflareLesson = {
```
joelhooks, line 1: this can be used anywhere you need static data for a video resource (lesson) that is associated with a collection

src/components/pages/lessons/overlay/recommend-next-step-overlay/recommend-next-step-overlay.stories.mdx
```diff
@@ -0,0 +1,18 @@
+import {Meta, Story, Canvas} from '@storybook/addon-docs/blocks'
```
joelhooks, line 1: Storybook makes this super easy to work on. We should definitely be using it.

PR #278, opened 2021-01-10
stabilizes video progress

joelhooks:
There were inconsistent results when watching collection videos and the appropriate overlays weren't functioning consistently.

### DIFF
--- a/src/components/EggheadPlayer/use-egghead-player.ts
+++ b/src/components/EggheadPlayer/use-egghead-player.ts
@@ -70,9 +70,12 @@ const storeProgress = (
   })
 }
 
-const onProgress = (lesson: LessonResource) => async (progress: {
-  playedSeconds: any
-}) => {
+const onProgress = async (
+  progress: {
+    playedSeconds: any
+  },
+  lesson: any,
+) => {
   const {playedSeconds} = progress
   const roundedProgress = Math.ceil(playedSeconds)
   const isSegment = roundedProgress % 30 === 0
@@ -199,11 +202,11 @@ const onComplete = (lesson: any, collection?: any) => {
     .then(trackPercentComplete)
 }
 
-const onEnded = (lesson: {
+const onEnded = async (lesson: {
   lesson_view_url: any
   slug: any
   tags: any[]
-}) => async () => {
+}) => {
   setEmailCaptureCookie({
     ...emailCaptureCookie,
     watchCount: emailCaptureCookie.watchCount + 1,
@@ -237,7 +240,7 @@ const defaultPlayerPreferences = {
     lang: null,
   },
   muted: false,
-  theater: false
+  theater: false,
 }
 
 export const getPlayerPrefs = () => {
@@ -282,17 +285,13 @@ export default function useEggheadPlayer(lesson: LessonResource) {
     setPlayerPrefs(getPlayerPrefs())
   }, [lesson.slug])
 
-  const onProgressCallback = React.useCallback(onProgress(lesson), [lesson])
-
-  const onEndedCallback = React.useCallback(onEnded(lesson), [lesson])
-
   const setPlayerPrefsCallback = React.useCallback((options: any) => {
     setPlayerPrefs(savePlayerPrefs(options))
   }, [])
 
   return {
-    onProgress: onProgressCallback,
-    onEnded: onEndedCallback,
+    onProgress,
+    onEnded,
     onError,
     setPlayerPrefs: setPlayerPrefsCallback,
     ...playerPrefs,
--- a/src/components/pages/lessons/collection-lessons-list.tsx
+++ b/src/components/pages/lessons/collection-lessons-list.tsx
@@ -17,7 +17,7 @@ const CollectionLessonsList: FunctionComponent<NextUpListProps> = ({
   course,
   currentLessonSlug,
   progress,
-  nextToVideo
+  nextToVideo,
 }) => {
   const {lessons} = course
   const [activeElement, setActiveElement] = React.useState(currentLessonSlug)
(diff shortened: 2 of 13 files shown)


src/components/EggheadPlayer/use-egghead-player.ts
```diff
@@ -70,9 +70,12 @@ const storeProgress = (
   })
 }
 
-const onProgress = (lesson: LessonResource) => async (progress: {
-  playedSeconds: any
-}) => {
+const onProgress = async (
```
joelhooks, line 73: instead of trying to do a "double jump" we require the lesson to be passed in explicitly

src/components/EggheadPlayer/use-egghead-player.ts
```diff
@@ -199,11 +202,11 @@ const onComplete = (lesson: any, collection?: any) => {
     .then(trackPercentComplete)
 }
 
-const onEnded = (lesson: {
+const onEnded = async (lesson: {
```
joelhooks, line 205: similar here, this kept having a "stale" video and I couldn't figure out why

src/components/pages/lessons/collection-lessons-list.tsx
```diff
@@ -37,21 +37,28 @@ const CollectionLessonsList: FunctionComponent<NextUpListProps> = ({
...
             maxHeight: 300,
           }}
         >
           {lessons.map((lesson: LessonResource, index = 0) => {
             const completedLessons = get(progress, 'completed_lessons', []).map(
               (lesson: LessonResource) => lesson.slug,
             )
-            const completed = completedLessons.includes(lesson.slug)
+            const completed =
+              lesson.completed || completedLessons.includes(lesson.slug)
```
joelhooks, line 61: `lesson.completed` is available much more often (now)

src/components/pages/lessons/overlay/next-up-overlay.tsx
```diff
@@ -34,7 +35,7 @@ const NextUpOverlay: React.FunctionComponent<{
             track('clicked rewatch video', {
               lesson: lesson.slug,
             })
-            send('LOAD')
+            onClickRewatch()
```
joelhooks, line 38: this wasn't working at all

src/lib/lessons.ts
```diff
@@ -55,6 +55,7 @@ export async function loadLesson(slug: string, token?: string) {
               type
               path
               title
+              completed
```
joelhooks, line 58: the api already had this data

src/machines/lesson-player-machine.ts
```diff
@@ -164,6 +164,7 @@ const playerMachine = Machine<
       showingNext: {
         on: {
           LOAD: 'loading',
+          VIEW: 'viewing',
```
joelhooks, line 167: we want to be able to transition to `viewing` and don't need to load anything.

src/pages/lessons/[slug].tsx
```diff
@@ -146,6 +146,8 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
   const [watchCount, setWatchCount] = React.useState<number>(0)
   const currentPlayerState = playerState.value as string
 
+  const actualPlayerRef = React.useRef<any>(null)
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:149]: the `playerRef` above is the wrapper and not the player player, so here we are...

src/pages/lessons/[slug].tsx
```diff
@@ -168,6 +170,11 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
     if (media_url) {
       axios.get(media_url).then(({data}) => {
         setMedia(data)
+        const event: PlayerStateEvent = {
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:173]: instead of doing oddball stuff in a switch statement, we will jump to `loaded` here when we have media

src/pages/lessons/[slug].tsx
```diff
@@ -187,7 +194,7 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
   })
 
   const primary_tag = get(first(get(lesson, 'tags')), 'name', 'javascript')
-  const getProgress = () => {
+  const getProgress = (lessonView: any) => {
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:197]: we require a lesson view to be passed in instead of using the state var

src/pages/lessons/[slug].tsx
```diff
@@ -222,16 +229,17 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
...
   }
 
-  const completeVideo = () => {
+  const completeVideo = (lessonView: any) => {
     if (lessonView) {
-      const progress = getProgress()
-      if (progress?.rate_url) {
+      const hasNextLesson = nextLesson || nextUp.nextUpPath
+      const progress = getProgress(lessonView)
+      if (!hasNextLesson && progress?.rate_url) {
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:236]: we are only gonna pop the rating if it's the last lesson and there is no place else to go

src/pages/lessons/[slug].tsx
```diff
@@ -276,13 +273,18 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
         if (!media) {
           send('LOAD')
         }
-
         break
       case 'completed':
-        completeVideo()
+        console.debug('handling a change to completed', {lesson})
+        onEnded(lesson).then((lessonView: any) => {
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:279]: when this callback was passed into the player event handler below it was always using a different instance of `lesson` which leads me to believe there might be multiple player instances?

this was very unpleasant to debug.

src/pages/lessons/[slug].tsx
```diff
@@ -297,7 +299,17 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
   React.useEffect(() => {
     send('LOAD')
     setLesson(initialLesson)
-    loadLesson(initialLesson.slug, getAccessTokenFromCookie()).then(setLesson)
+
+    async function run() {
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:303]: the `then` call was acting extremely strangely and I didn't figure out a technical reason why, but this works as expected

src/pages/lessons/[slug].tsx
```diff
@@ -416,19 +431,17 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
                         send('PAUSE')
                       }}
                       onProgress={({...progress}) => {
-                        onProgress(progress).then((lessonView: any) => {
+                        onProgress(progress, lesson).then((lessonView: any) => {
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:434]: but, unlike `onEnded` this one has a reference to the correct `lesson` instance 😭

PR #283, opened 2021-01-11
transfers work to display comments

joelhooks:

### DIFF
--- a/src/components/EggheadPlayer/use-egghead-player.ts
+++ b/src/components/EggheadPlayer/use-egghead-player.ts
@@ -241,6 +241,7 @@ const defaultPlayerPreferences = {
   },
   muted: false,
   theater: false,
+  defaultView: 'transcript',
 }
 
 export const getPlayerPrefs = () => {
--- a/src/components/pages/lessons/comment.tsx
+++ b/src/components/pages/lessons/comment.tsx
@@ -1,0 +1,64 @@
+import * as React from 'react'
+import ReactMarkdown from 'react-markdown'
+import friendlyTime from 'friendly-time'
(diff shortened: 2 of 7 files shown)


src/components/EggheadPlayer/use-egghead-player.ts
```diff
@@ -241,6 +241,7 @@ const defaultPlayerPreferences = {
   },
   muted: false,
   theater: false,
+  defaultView: 'transcript',
```
joelhooks, line 244: this stretches "player prefs" but it was right there and EZ

src/components/pages/lessons/comment.tsx
```diff
@@ -0,0 +1,64 @@
...
+            }`}
+          >
+            <span className={isCommentableOwner ? 'font-medium' : ''}>
+              {user.full_name}
+            </span>
+            {isCommentableOwner && (
+              <span className="ml-3 text-blue-600 text-sm">(instructor)</span>
+            )}
+          </div>
+          <div className="ml-3">~ {friendlyTime(new Date(createdAt))}</div>
```
joelhooks, line 54: found a lib that formats time

src/pages/lessons/[slug].tsx
```diff
@@ -683,8 +689,19 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
                   />
                 </>
               )}
-
-              <Tabs>
+              <Tabs
+                defaultIndex={
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:693]: thought we can be smart when people switch and show them that view as default

PR #286, opened 2021-01-12
adds fullscreen autoplay

joelhooks:
Let's people watch videos in fullscreen and autoplay with only minor glitchiness between videos


src/components/EggheadPlayer/players/Bitmovin.js
```diff
@@ -228,9 +228,14 @@ export default class Bitmovin extends Base {
       onSubtitleChange,
       onVideoQualityChanged,
       onMuted,
+      onViewModeChanged,
     } = this.props
 
     this.player.on(this.player.exports.PlayerEvent.Play, this.onPlay)
+    this.player.on(
+      this.player.exports.PlayerEvent.ViewModeChanged,
```
joelhooks, line 236: we needed a way to monitor when the view mode changed to `fullscreen`

src/components/EggheadPlayer/players/Bitmovin.js
```diff
@@ -329,6 +339,11 @@ export default class Bitmovin extends Base {
     this.player && this.player.pause()
   }
 
+  isFullscreen() {
```
joelhooks, line 342: this is a façade over the bitmovin player so we need to wrap underlying methods

fun fact, you forget this every couple of years

src/pages/lessons/[slug].tsx
```diff
@@ -221,13 +222,40 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
...
-  const checkAutoPlay = () => {
+  const checkAutoPlay = async () => {
     if (autoplay && nextLesson) {
       console.debug('autoplaying next lesson', {nextLesson})
       track('autoplaying next video', {
         video: nextLesson.slug,
       })
-      router.push(nextLesson.path)
+
+      if (isFullscreen) {
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:232]: if we are fullscreenin we want to just move to the next video and make sure the player instance doesn't curn

src/pages/lessons/[slug].tsx
```diff
@@ -221,13 +222,40 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
...
+
+        send({
+          type: 'VIEW',
+          lesson: loadedLesson,
+        })
+
+        setMedia(mediaUrls)
+        setLesson(loadedLesson)
+
+        router.push(loadedLesson.path, undefined, {
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:253]: this causes a flash of the whole UI, which isn't nice, but also it keeps playing so that's cool

src/pages/lessons/[slug].tsx
```diff
@@ -221,13 +222,40 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
...
+          shallow: true,
+        })
+      } else {
+        router.push(nextLesson.path)
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:257]: otherwise, autoplay like normal by pushing the path

src/pages/lessons/[slug].tsx
```diff
@@ -236,9 +264,16 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
   }
 
   const completeVideo = (lessonView: any) => {
+    console.debug('completed video', {lessonView, video: lesson})
     if (lessonView) {
       const hasNextLesson = nextLesson
       const progress = getProgress(lessonView)
+
+      if (!hasNextLesson && isFullscreen) {
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:272]: if we are fullscreenin and there's no more videos to watch, jump out of fullscreen

theianjones (approved):

PR #289, opened 2021-01-13
use xstate instead of usestate for lesson state management 

joelhooks:
Was having odd issues with setState and xState and the lesson is already in xState context so let's just use that instead of storing lesson also in useState.


src/components/pages/lessons/overlay/next-up-overlay.tsx
```diff
@@ -11,7 +11,7 @@ const NextUpOverlay: React.FunctionComponent<{
 }> = ({lesson, nextLesson, onClickRewatch = noop}) => {
   const courseImage = lesson?.collection?.square_cover_480_url
   return (
-    <>
+    <div className="bg-gray-800 bg-opacity-90 grid place-items-center w-full h-full absolute z-10 top-0 left-0">
```
joelhooks, line 14: let this exist as an overaly on top of the player

@DrShpongle this made it look kind of bad.

src/pages/lessons/[slug].tsx
```diff
@@ -116,44 +116,41 @@ type LessonProps = {
...
-  }
-
-  const [lesson, setLesson] = React.useState<any>(initialLesson)
-  const router = useRouter()
   const playerRef = React.useRef(null)
-  const {viewer} = useViewer()
-  const [playerState, send] = useMachine(playerMachine)
+  const actualPlayerRef = React.useRef<any>(null)
+
+  const [playerState, send] = useMachine(playerMachine, {
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:142]: with `useMachine` the second argument is `options` which let's us send in an initial context. In this case, we can use the initial lesson.

src/pages/lessons/[slug].tsx
```diff
@@ -116,44 +116,41 @@ type LessonProps = {
...
+    context: {lesson: initialLesson},
+  })
+
+  const lesson: any = get(playerState, 'context.lesson', initialLesson)
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:146]: now we can use a `lesson` variable that is retrieved from the context. The `initialLesson` fallback is likely not strictly required here.

src/pages/lessons/[slug].tsx
```diff
@@ -199,13 +197,15 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
...
   const primary_tag = get(first(get(lesson, 'tags')), 'name', 'javascript')
+
   const getProgress = (lessonView: any) => {
     if (lessonView?.collection_progress) {
       return lessonView.collection_progress
     }
   }
 
-  const loaderVisible = playerState.matches('loading')
+  const loaderVisible = ['loading', 'completed'].includes(currentPlayerState)
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:207]: when we are in `completed` there can be a pause so we want to show the spinner

src/pages/lessons/[slug].tsx
```diff
@@ -217,7 +217,7 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
         'loaded',
         'viewing',
         'completed',
-        'autoPlaying',
+        'showingNext',
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:220]: I'm trying to "churn" the player less and reuse the instance instead of hiding it and bringing it back

still not "just right" but better

src/pages/lessons/[slug].tsx
```diff
@@ -317,13 +316,19 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
...
+        console.debug('handling a change to completed', {lesson, lessonView})
         onEnded(lesson).then((lessonView: any) => {
           if (lessonView) {
             setLessonView(lessonView)
           }
-          completeVideo(lessonView)
+          if (!completed) completeVideo(lessonView)
         })
+
+        if (lessonView) {
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:328]: if we have a `lessonView` in state, we can "complete" instead of waiting for the async saving of the lesson view. The makes the wait a lot less at the end of a lesson, but will also capture a lesson view if needed (if the user skipped to the end before we were able to get the lesson view, for instance)

src/pages/lessons/[slug].tsx
```diff
@@ -383,10 +385,6 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
     setLessonMaxWidth(Math.round((height - HEIGHT_OFFSET) * 1.77))
   }, [height])
 
-  React.useEffect(() => {
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:386]: I didn't understand this and don't think it is needed as `useEggheadPlayer` prefs manage this state so no need to duplicate it.

PR #291, opened 2021-01-14
initial value of viewer from local storage

joelhooks:
We are loading the `viewer` async and sometimes this would mean we don't have a viewer even though the lesson has been loaded.


src/context/viewer-context.tsx
```diff
@@ -34,8 +34,7 @@ export function useViewer() {
 export const ViewerContext = React.createContext(defaultViewerContext)
 
 function useAuthedViewer() {
-  const router = useRouter()
-  const [viewer, setViewer] = React.useState<any>()
+  const [viewer, setViewer] = React.useState<any>(auth.getLocalUser())
```
joelhooks, line 37: if we have a viewer in `localStorage` we can be optimistic about it?

src/machines/lesson-player-machine.ts
```diff
@@ -20,11 +20,11 @@ interface PlayerStateSchema {
 
 export type PlayerStateEvent =
   | {type: 'VIEW'}
-  | {type: 'VIEW'; lesson: any}
+  | {type: 'VIEW'; lesson: any; viewer: any}
```
joelhooks, line 23: I'd like to move more of the transition logic between states into the machine and I noticed that the viewer was `undefined` when these transitions occurred because the viewer was still loading in the background

src/pages/lessons/[slug].tsx
```diff
@@ -167,24 +169,9 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
     collection,
     free_forever,
     slug,
-    media_url,
     comments,
   } = lesson
 
-  React.useEffect(() => {
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:174]: not really using this anymore so no need to load

src/utils/get-access-token-from-cookie.ts
```diff
@@ -1,11 +1,13 @@
...
 import cookies from './cookies'
 
 const getAccessTokenFromCookie = () => {
   if (!ACCESS_TOKEN_KEY) return false
-  const token = cookies.get(ACCESS_TOKEN_KEY)
+  let token = cookies.get(ACCESS_TOKEN_KEY)
   if (token) {
     return token
+  } else if (typeof localStorage !== 'undefined') {
+    return localStorage.getItem(ACCESS_TOKEN_KEY) ?? false
```
joelhooks, line 10: this is a fallback since the key is here as well...

joelhooks (commented):
Trying to figure out why [an instructor couldn't view pro lessons](https://discord.com/channels/365559264850477066/715365813816786995/799310855165444127).

When "viewing as" the user, I was able to see lessons, so latency is my number one culprit.

PR #292, opened 2021-01-15
Hide Comments tab if no comments

DrShpongle:
1. `Comments` tab is hidden now if there are no comments
2. Fixed default tab index we store in cookies

vs

### DIFF
--- a/src/pages/lessons/[slug].tsx
+++ b/src/pages/lessons/[slug].tsx
@@ -740,9 +740,7 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
                 </>
               )}
               <Tabs
-                defaultIndex={
-                  defaultView === 'comments' && commentsAvailable ? 1 : 0
-                }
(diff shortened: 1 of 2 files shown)


src/pages/lessons/[slug].tsx
```diff
@@ -754,7 +752,7 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
               >
                 <TabList className="md:text-lg text-normal font-semibold bg-transparent space-x-1">
                   {transcriptAvailable && <Tab>Transcript</Tab>}
-                  <Tab>Comments</Tab>
+                  {commentsAvailable && <Tab>Comments</Tab>}
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:755]: this doesn't advance the feature. We need to show a comment box if they are able to comment and a CTA if they are not able to comment. If this PR included that markup, then I can wire it to the backend and advance the functionality. 

I feel like this is a regression.

Comment (DrShpongle) [reply src/pages/lessons/[slug].tsx:755]: hmmm, really 🤔
let me refactor it with markup for the comment field

Issue #275, opened 2021-01-10
useBreakpoint value names are hard to comprehend


joelhooks:
These are confusing when in encountered in JSX and we should consider using verbose longform variable names that fully describe what it is/does.

https://github.com/eggheadio/egghead-next/blob/9536ccb34ba3a080f6c703c7fc9f440587b9499b/src/utils/breakpoints.ts#L13-L21

This longer form we are exporting are much easier to comprehend, so something similar for this purpose would be very appreciated. 

The names should express what the value does.

Additionally this utility is prefixed with `use` and *is not a React Hook*. `use*` should only be applied to React Hooks.

For "custom hooks" this typically means the custom hook will call one of the built-in hooks like `React.useState` or `React.useEffect` and not simply provide a return value in the style of a React hook. 

![image](https://user-images.githubusercontent.com/86834/104131156-811f2a80-5329-11eb-8902-2fe612292a17.png)

This utility could use some documentation as well.

Discussion welcome!

vojtaholik:
breakpoints | descriptive names | alt | device names
-- | -- | -- | --
375 | bpMinXS | isSizeMinXS | mobileSmall
640 | bpMinSM | isSizeMinSM | mobile
768 | bpMinMD | isSizeMinMD | tablet
1024 | bpMinLG | isSizeMinLG | laptop
1280 | bpMinXL | isSizeMinXL | desktop
1440 | bpMinXL2 | isSizeMinXL2 | desktopLarge
1920 | bpMinXL3 | isSizeMinXL3 | desktopWide


I think I like `bpMinSM` (or some form of it) approach the most — it’s descriptive thanks to the `bp` keyword, extendable with `max-width` breakpoints (if we ever need them), while still resembling Tailwind's responsive classes. 

Device names would need a separate function exported for `max-width` breakpoints. Tailwind does support [custom breakpoint names](https://tailwindcss.com/docs/breakpoints) which allows for consistency across the board.

I also find idea of [ergonomic breakpoint variables](http://alexbea.com/blog/ergonomic-breakpoint-variables/) interesting.

joelhooks:
i wouldn’t recommend (or even consider) patching tailwind to use a different system, and i understand “md” etc were probably trying to match tailwind, but it’s two different contexts IMO

joelhooks:
i really like the “ergonomic” names!

--- original message ---
On January 11, 2021, 09:43 PST joel@egghead.io wrote:
--- end of original message ---

[...]

PR #307, opened 2021-01-19
Refactored breakpoints

DrShpongle:
Hopefully fixes this: https://github.com/eggheadio/egghead-next/issues/275

Used "mobile-first" approach to define breakpoints.
For CSS breakpoints (which we strive to use whenever it's possible) we use `bpMinXX` naming and I used `isMinXX` for the "js-defined" breakpoints.

Demo of how this approach works: https://share.getcloudapp.com/NQuKyD0r


DrShpongle:
in process of refactoring...

src/utils/breakpoints.ts
```diff
@@ -20,20 +21,16 @@ export const bpMaxSM = `@media (max-width: ${bp.md - 1}px)`
...
-  return {xs, sm, md, lg, xl, xl2, xl3}
+const breakpoints = () => {
+  const breakpoint = getBreakpoint()
+  const keys = Object.keys(bp)
+  const foundInd = keys.indexOf(breakpoint)
+  return keys
+    .map((key, ind) => ({['isMin' + key.toUpperCase()]: ind <= foundInd}))
+    .filter((_, ind) => ind)
+    .reduce((acc, cur) => ({...acc, ...cur}), {})
 }
```
DrShpongle, line 34: Eventually we get the object like:
```
{
  isMin2XL: false
  isMin3XL: false
  isMinLG: true
  isMinMD: true
  isMinSM: true
  isMinXL: true
}
```

only "mobile view" (i.e. `xs`) could be considered like `!isMinSM`

src/utils/breakpoints.ts
```diff
@@ -20,20 +21,16 @@ export const bpMaxSM = `@media (max-width: ${bp.md - 1}px)`
...
-  const xl3 = breakpoint == '3xl' || xl2
-
-  return {xs, sm, md, lg, xl, xl2, xl3}
+const breakpoints = () => {
+  const breakpoint = getBreakpoint()
+  const keys = Object.keys(bp)
+  const foundInd = keys.indexOf(breakpoint)
+  return keys
+    .map((key, ind) => ({['isMin' + key.toUpperCase()]: ind <= foundInd}))
+    .filter((_, ind) => ind)
```
DrShpongle, line 32: Idea here is to filter out `XS` value, because it's like `default` class in tailwindcss (`flex-row` vs `sm:flex-row`)

src/utils/breakpoints.ts
DrShpongle: Screen+Recording+2021-01-20+at+06 05 57+PM

PR #322, opened 2021-01-24
adds a little block about the freshness of a course

joelhooks:
As far as design goes this is just slapped in there, but conceptually I think this has a lot of merit so I wanted to put it on the page and see how it felt.

I just used our trusty data buddy `courseDependencies`

image


src/data/courseDependenciesData.js
```diff
@@ -367,8 +369,17 @@ const courseDependencyData = (courseSlug) =>
       },
       {
         slug: 'the-beginner-s-guide-to-react',
+        freshness: {
```
joelhooks, line 372: here's the ✨

joelhooks:
image

PR #325, opened 2021-01-25
Comments styles tweaks

DrShpongle:
before:
before

after:
after

### DIFF
--- a/src/components/pages/lessons/comment.tsx
+++ b/src/components/pages/lessons/comment.tsx
@@ -29,7 +29,7 @@ const Comment: React.FunctionComponent<CommentProps> = ({
   return (
     <div className="flex">
       {user.avatar_url && (
-        <div className="flex-shrink-0 mr-4">
+        <div className="flex-shrink-0 mr-3 sm:mr-4">
           <img
             src={user.avatar_url}
             alt={user.full_name}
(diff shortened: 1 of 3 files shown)


src/pages/lessons/[slug].tsx
```diff
@@ -781,7 +781,10 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
                   )}
                   {commentsAvailable && (
                     <TabPanel>
-                      <div className="space-y-10">
+                      <div
+                        className="space-y-8 sm:space-y-10"
```
Comment (vojtaholik) [inline src/pages/lessons/[slug].tsx:785]: what do you think about using `prose` class for comments?

src/components/pages/lessons/comment.tsx
```diff
@@ -38,20 +38,20 @@ const Comment: React.FunctionComponent<CommentProps> = ({
...
               <span className="ml-3 text-blue-600 text-sm">(instructor)</span>
             )}
           </div>
-          <div className="ml-3">~ {friendlyTime(new Date(createdAt))}</div>
+          <div className="ml-2 sm:ml-3">
+            ~ {friendlyTime(new Date(createdAt))}
+          </div>
         </div>
         <div className="mt-3">
           <ReactMarkdown>{comment}</ReactMarkdown>
```
vojtaholik, line 57: what do you think about using `prose` class for comments?

joelhooks (approved):

PR #326, opened 2021-01-25
add a stale course

joelhooks:


dealingwith:
Thought for a follow-up: find a spot on the child lesson pages for this info

PR #333, opened 2021-01-26
Zj/more javascript landing page

zacjones93:
This PR adds the layout for the JavaScript landing page:

CleanShot 2021-01-29 at 16 44 28@2x

dog i cant handle this - WZP3qaxYj10gU


vojtaholik:
the whole "Composing Closures and Callbacks in JavaScript" section has a darker background, which is meant to group things under one topic visually. (see state management on react page) But it looks like Debugging and favorites don't quite belong under it.

MaggieAppleton:
Yeah I think this page needs a different layout to the React one based around the content. I'm going to layout some ideas in Figma then rearrange

zacjones93:
That makes sense. I used the React page components to get the data up and running with as little friction as possible so hopefully rearranging the layout won't be an issue once we've landed on a solution

joelhooks:
I added the `RFC` tag, which we use as a "request for comments" - it also blocks the auto-merge robot so should be removed when this is ready

zacjones93:
I added the async card which looks great on desktop but can't figure out why the styling gets cramped on mobile

## Desktop
In-Depth Up-to-Date JavaScript Tutorials for 2021  egghead io

## Mobile
In-Depth Up-to-Date JavaScript Tutorials for 2021  egghead io

src/components/search/curated/javascript/index.tsx
```diff
@@ -108,21 +109,32 @@ When someone comes to you and asks "Hey! Can you build this?", you'll be able to
...
         </Card>
-        <Card
-          className="md:col-span-8"
-          resource={composingCallbacks}
-          location={location}
-        >
-          <Collection />
-        </Card>
+        <div className="md:col-span-8 min-w-full gap-5">
+          <div className="md:grid md:grid-cols-8 flex flex-col w-max items-start gap-5">
```
zacjones93, line 121: It has something to do with how I'm applying the nested grid here

zacjones93:
slight margin issue at the bottom but other than that this is gtg
In-Depth Up-to-Date JavaScript Tutorials for 2021  egghead io

laurosilvacom (approved):

PR #334, opened 2021-01-27
Ma/design improvements

MaggieAppleton:
Added a tiny bit more breathing room to the homepage and React landing page – added "mt-20" between distinct sections, and adjusted the padding on the main feature box.

At the moment all our layouts are very dense. We need to start adding more whitespace between sections to create distinct clusters of related information. This will help breakup the 'wall of content' people are sad about in our feedbox section.

These are quick, minor adjustments within the current limitations of our tailwind setup.
Even when I add "mt-20" to an element, it only adds about 100px of space. Anything above "mt-20" doesn't register. I can't figure out how I would add proper spacing between elements (eg. anything in the 100-500px range.)

I also was trying to switch the dark mode colour palette from 'trueGray' to 'blueGray' (that's the palette we use across the rest of the site so it'll feel more consistent), but Tailwind wasn't having it. When I would swap the colour it would just turn the dark mode background to white  ¯\_(ツ)_/¯


MaggieAppleton:
Previews of the extra spaces in between sections:

image

Screenshot 2021-01-27 at 10 51 29

Not a perfect solution, but want to explore more ways of creating sections of content going forward. We could also do it with divider lines, or different patterns/colours on the backgrounds of sections

joelhooks:
this PR needs a "hilarious gif" in the description :)

joelhooks:
We've got `purging` turned on locally and it can cause a frustrating experience when you add a new class that wasn't in the code at build time. I chose this because otherwise the build time is very slow and also crashes if you don't five node additional memory

wrote some about it here -> https://egghead.io/learn/next/tailwindcss-dark-mode-nextjs-typography-prose

joelhooks (approved):

PR #336, opened 2021-01-27
adds a generic topic/tag header

joelhooks:
This is low effort, but adds a generic topic header for all filtered landing pages:

image


src/components/search/generic-topic.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+type TopicProps = {
+  title: string
+  children?: string
+  imageUrl?: string
+  className?: string
+  description?: string
+}
+
+const GenericTopic: FunctionComponent<TopicProps> = ({
```
joelhooks, line 14: I cut and pasted `Topic` component for this with some specific changes like seo tags etc

PR #340, opened 2021-01-28
updated benefits on pricing page

MaggieAppleton:
Added extra benefits to the pricing page


MaggieAppleton:
@vojtaholik Daniel and I were addding a few extra points to the pricing page and got hit with a weird vercel error.
Can't debug it locally without stripe public keys.

Can you give the changes a look over and merge if it's all okay?

vojtaholik:
@MaggieAppleton  looks like API requests are timing out, redeploy usually fixes it. 
cd474423-88b0-40b9-bc6b-cb708d6996d2

MaggieAppleton:
@vojtaholik Nice, that fixed it. Thanks!

Preview:
image

src/components/pricing/select-plan.tsx
```diff
@@ -42,7 +42,83 @@ const SelectPlan: React.FunctionComponent = ({children}) => {
...
+                      >
+                        <path
+                          fillRule="evenodd"
+                          d="M10 18a8 8 0 100-16 8 8 0 000 16zm3.707-9.293a1 1 0 00-1.414-1.414L9 10.586 7.707 9.293a1 1 0 00-1.414 1.414l2 2a1 1 0 001.414 0l4-4z"
+                          clipRule="evenodd"
+                        />
+                      </svg>
+                    </div>
+                    <p className="ml-3 text-sm leading-5 text-gray-700 dark:text-gray-100">
+                      Access to all 192 premium courses
```
joelhooks, line 61: we've got considerably more than this

dealingwith: Yeah, we were still crunching the numbers. Does this look right?

Paid courses: 214
Free courses: 44

(I just realized that's not "considerably more")

We also considered using # of lessons instead as it would be BIGGER.

joelhooks: I dunno, I don't think it's gonna make a difference wrt to conversions or anything. You could probably just say "over 200" or whatever and it'd be fine instead of hyper-specific which is gonna be wrong quickly unless loaded dynamically at build time.

If "premium" in this instance means "paywalled" 212 sounds about right, not including any courses that are actually playlists.

joelhooks (approved):
approving this, but let's not quit **improving** this 😂

@vojtaholik @MaggieAppleton would love to see this get the full boat UX+design overhaul! I'm at your disposal.

dealingwith:
This was def a minimal first pass. I wanted to get that "over 200" change in there, too :D

PR #344, opened 2021-01-28
Get stripe_customer_id from viewer.subscriptions

jbranchaud:
It looks like the server is passing `stripe_customer_id` in the
`viewer.subscriptions` object instead of `stripe_subscription_id`.

I'm not sure how user's with existing customer_ids were ended up with an
actual value passed to Checkout with this as is.


src/pages/pricing/index.tsx
```diff
@@ -45,7 +45,7 @@ const Pricing: FunctionComponent<PricingProps> = ({redirectURL}) => {
       stripeCheckoutRedirect(
         annualPrice.price_id,
         viewer.email,
-        viewer.subscription?.stripe_subscription_id,
+        viewer.subscription?.stripe_customer_id,
```
joelhooks, line 48: I think this should look at account now, this was pre-account (should hold up for awhile)

joelhooks (approved):

PR #350, opened 2021-01-30
add sanity

joelhooks:
this adds sanity.io to the project and defines an initial schema to replace our [`courseDependencies`](https://github.com/eggheadio/egghead-next/blob/b487eade265b0f0a3fe40dbef1ac4672c6095c91/src/data/courseDependenciesData.js) with a structured schema we can query.


studio/schemas/collaborator.js
```diff
@@ -0,0 +1,50 @@
+export default {
+  name: 'collaborator',
```
joelhooks, line 2: can be anybody for anything really, currently we use this for illustrators

studio/schemas/npmDependency.js
```diff
@@ -0,0 +1,30 @@
+export default {
+  name: 'npmDependency',
+  title: 'npm Dependency',
```
joelhooks, line 3: our "dependencies" are mostly npm libraries so this is specific, we could expand to other fields

studio/schemas/resource.js
```diff
@@ -0,0 +1,249 @@
+import {MdKitchen as icon} from 'react-icons/md'
+
+export default {
+  name: 'resource',
```
joelhooks, line 4: this is the main thing, it's a resource and can be used to describe a lot of different content type things that you'd find on egghead etc

studio/schemas/resource.js
```diff
@@ -0,0 +1,249 @@
...
+  title: 'Resource',
+  type: 'document',
+  icon,
+  fields: [
+    {
+      name: 'type',
```
joelhooks, line 10: type primarily describes how we handle it in the UI and what we expect in that regard.

joelhooks:
Deploying the graphql api requires a second command `sanity graphql deploy`

https://graphqlbin.com/v2/nGNBsx

image

PR #353, opened 2021-01-30
Add id to playlists and courses

zacjones93:
<details>
<summary>Developer Log</summary>
<br>

- Assign course/playlist id's to courseDependency items
    - Problem:
        - We are storing course metadata in a file as an array of JSON that we look up by slug. 
        - The issue that arises is that the slug can change depending on  the current course title. 
            - This whole file is hardcoded data so it needs to manually be updated every time titles are updated. 
            - This is prone to break. 
            - Initially, it was a decent solution as the slugs are readable.
    - Solution
        - Use the ID! 
        - We serve the id through the API 
        - Will need to zip the id to the current courses we have in the courseDependency 
        - Step 1:
            - Separate playlists from courses, did this manually since we have a handful of playlists in the file
            - Blazer query to get id and slug
            - create hash of `id: slug`
            - map over the keys and set `id: key` and spread the rest of the obj       
```javascript
let array = []
Object.keys(playlistID).map(async (key) => {
  let tempObj = _.cloneDeep(playlistDependencyData(playlistID[key]))
  let newObj = {
    id: key,
    ...tempObj
    }

  array.push(newObj)
})
```
        - Step 2: Courses
            - Because not all courses are in the dependency list, need to guard against that
            - processing with the guard:
```javascript
let courseArray = []
let notFoundArray = []
Object.keys(courseID).map((key) => {
  if(!courseDependencyData(courseID[key])) {
    notFoundArray.push({[key]: courseID[key]})
    return
  }

  let tempObj = _.cloneDeep(courseDependencyData(courseID[key]))
  let newObj = {
    id: key,
    ...tempObj
  }

  courseArray.push(newObj)

})
```

- I posted the courses-not-in-list hash above if we ever need it. 

</details>

- [Gist of full script](https://gist.github.com/zacjones93/7f9977c61f1f2886c5ddac4bf6150e13)
- [List of courses not in dependencies json](https://gist.github.com/zacjones93/e921f301e4e299017217414d13a3be5f)

*edit - Gists are hard to navigate, the logic is [here](https://gist.github.com/zacjones93/7f9977c61f1f2886c5ddac4bf6150e13#file-get-collection-ids-js-L323-L335) for playlists and [here](https://gist.github.com/zacjones93/7f9977c61f1f2886c5ddac4bf6150e13#file-get-collection-ids-js-L3751-L3771) for courses

Because I was looping through a hash of IDs -> slugs, the resulting courseDependenciesData file is not sorted based on creation data since we just increment the id on creation.


zacjones93:
The diff is a little jarring since the objects got re-sorted

joelhooks:
@zacjones93 I think this will need a `type` field for the ids to be useful. Also interesting to consider the migration to #350

src/data/courseDependenciesData.js
```diff
@@ -4,20 +4,22 @@ const courseDependencyData = (courseSlug) =>
   find(
     [
       {
+        id: '418892',
+        type: 'playlist',
```
zacjones93, line 8: Type and id!

src/data/courseDependenciesData.js
```diff
@@ -36,6 +38,9 @@ const courseDependencyData = (courseSlug) =>
         ],
         pairWithResources: [
           {
+            id: '414202',
+            type: 'playlist',
+            guid: '553c',
```
zacjones93, line 43: not sure if guid is right here.. I grabbed the end of the playlist slug. `id` probably works?

joelhooks (approved):

PR #357, opened 2021-01-31
Curated Instructor Page

laurosilvacom:
This PR adds 2 things:  

1. A reusable header component for instructor pages. 
2. A curated instructor page for [Dan Abramov](https://egghead-io-nextjs-git-ls-instructor-curated-page.eggheadio1.vercel.app/q/resources-by-dan-abramov).

This component is used for the non-curated instructor pages, but it can also be composed for more complex uses when curating instructor pages.

## Curated Instructor Page

CleanShot 2021-01-31 at 00 24 44@2x

## Default Instructor Page

CleanShot 2021-01-31 at 00 23 23@2x

dog head slant - 16StbF7VCSw6Y

PR #358, opened 2021-01-31
Proof of Concept: Customization for instructor's header

colbyfayock:
Note: Not intended to be merged

Put together a PoC with what could be a way for instructors to customize their page. It's pretty much based on the curated topics, but tried to make it in a way where instructors could add their own header without having to touch the core search page component.

I'm not entirely familiar with Typescript so please ignore my ignorance with a few things (particularly where I put "any" to make things work".

Curious on thoughts about it, but thought it could be a fun way to expose it for others to hop in and make something easily.

### DIFF
--- a/src/components/search/components/topic.tsx
+++ b/src/components/search/components/topic.tsx
@@ -15,8 +15,6 @@ const Topic: FunctionComponent<TopicProps> = ({
   imageUrl,
 }) => {
   const isRenderProp = typeof children === 'function'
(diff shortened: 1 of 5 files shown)


joelhooks:
this looks great. we merged this, which I think makes it simpler to apply custom pages for instructors

joelhooks:
https://github.com/eggheadio/egghead-next/pull/357

PR #362, opened 2021-01-31
tweaking the schema

joelhooks:


studio/schemas/bigIdea.js
```diff
@@ -0,0 +1,35 @@
+import {MdLightbulbOutline as icon} from 'react-icons/md'
+
+export default {
+  name: 'bigIdea',
```
zacjones93, line 4: Oh I like this as a content type to sprinkle in across resources

studio/schemas/cta.js
```diff
@@ -0,0 +1,73 @@
+export default {
+  title: 'Call to action',
```
zacjones93, line 2: This is going to be super helpful to have and be able to drop on different pages

studio/schemas/npmDependency.js
```diff
@@ -19,8 +17,23 @@ export default {
...
+      title: 'Description',
+      type: 'markdown',
+      options: {
+        maxLength: 180,
+      },
+    },
+    {
+      name: 'path',
+      description: 'Path on egghead.io (if applicable)',
+      title: 'egghead.io/ Path',
```
zacjones93, line 34: Will this be all the places that a single dependency is linked to?

studio/schemas/resource.js
```diff
@@ -41,61 +53,167 @@ export default {
       },
     },
     {
-      name: 'title',
-      title: 'Title',
-      type: 'string',
+      name: 'challengeRating',
```
zacjones93, line 56: I like this

studio/schemas/resource.js
```diff
@@ -126,54 +259,92 @@ export default {
       of: [{type: 'staffReview'}],
     },
     {
-      name: 'bullets',
-      title: 'Bullets',
-      description: 'Topics, value, etc. A list.',
+      name: 'features',
```
zacjones93, line 262: I like this name better than bullets

zacjones93 (approved):
This makes sense to me! I need to wrap my head around cta <-> ctaPlug. Especially like how were going to be able to surface Big Ideas and Essential Questions through the CMS

Is ctaPlug a common term?

PR #368, opened 2021-02-01
pricing components design

vojtaholik:
This PR is adding new pricing, select-plan, and testimonials components to StoryBook.

pricing-screenshots_shrink


src/components/pricing/index.tsx
```diff
@@ -0,0 +1,84 @@
+import * as React from 'react'
+import {FunctionComponent} from 'react'
+import SelectPlanNew from 'components/pricing/select-plan-new'
+import Testimonials from 'components/pricing/testimonials'
+import testimonialsData from './testimonials/data'
+import PoweredByStripe from './powered-by-stripe'
+
+const QUANTITY_PRICING_ENABLED =
```
joelhooks, line 8: I introduced this as a feature flag.

src/components/pricing/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+            interval: 'year'
+          }
+          unit_amount: number
+        }
+      }
+  pricesLoading: boolean
+  needsEmail: boolean
+  setNeedsEmail: React.Dispatch<React.SetStateAction<boolean>>
+  redirectURL?: string
+  dark?: boolean
```
joelhooks, line 29: suggestion: passing `dark` in should probably be removed in favor of a global storybook level toggle imo

src/components/pricing/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+            as a Web Developer
+          </h1>
+          <h2 className="text-lg font-light max-w-md pt-8 leading-tight dark:text-gray-200 text-gray-700">
+            learn the skills you need to advance your career and build
+            real-world business focused professional projects on the job and for
+            your portfolio
+          </h2>
+        </header>
+        <main className="flex flex-col items-center">
+          <div className="p-2 relative dark:bg-gray-800 bg-gray-100 rounded-md dark:shadow-none shadow-lg">
```
joelhooks, line 65: I moved this from `SelectPlanNew` to give the parent control of the basic layout of the component instead of specifying it internally and having to "dig" in.

src/components/pricing/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+            <SelectPlanNew
+              prices={prices}
+              pricesLoading={pricesLoading}
+              handleClickGetAccess={() => {}}
+              quantityAvailable={QUANTITY_PRICING_ENABLED && quantityAvailable}
```
joelhooks, line 70: `QUANTITY_PRICING_ENABLED` is false in prod, this feature is gonna lag the rest so I wanted to get this in.

src/components/pricing/select-plan-new/index.tsx
```diff
@@ -0,0 +1,242 @@
...
+        <div className={quantityAvailable ? '' : 'pb-4'}>
+          <PlanIntervalsSwitch
+            disabled={forTeams}
+            currentPlan={currentPlan}
+            setCurrentPlan={setCurrentPlan}
+            planTypes={individualPlans}
+          />
+        </div>
+        {quantityAvailable && (
+          <div className="py-4">
```
joelhooks, line 222: suggestion: spacing and basic structural layout is likely better suited for the "parent" where the component itself only defines essential styles. It's more verbose but also means the components are more flexible and you don't have to dig around sub-components for simple layout manipulation

joelhooks (commented):
added a feature flag for `quantity`

src/components/pricing/pricing.stories.mdx
```diff
@@ -0,0 +1,115 @@
...
+<Meta
+  title="Components/Pricing"
+  component={PricingPage}
+  argTypes={{pricesLoading: {control: 'boolean'}}}
+/>
+
+import usePricing from '../../hooks/use-pricing'
+
+export const Template = (args) => {
+  const {prices} = usePricing({mock: true})
```
joelhooks, line 91: added a `mock` mode to prices just to "keep it real"

src/components/pricing/select-plan-new/index.tsx
```diff
@@ -0,0 +1,248 @@
...
+  pricesLoading,
+  prices,
+}) => {
+  const individualPlans = filter(
+    prices,
+    (plan: any) => plan.type === 'individual',
+  )
+  const teamPlan = find(prices, {type: 'team'})
+
+  const annualPlan = get(prices, 'annualPrice', {
```
joelhooks, line 176: our service throws an error if these don't exist, and the page would be broken so since it's the default we can pass in some info.

src/components/pricing/select-plan-new/index.tsx
```diff
@@ -0,0 +1,248 @@
...
+  const monthlyPlan = get(prices, 'monthlyPrice')
+  const quarterlyPlan = get(prices, 'quarterlyPrice')
+
+  const pricesForInterval = (interval: any) => {
+    switch (interval) {
+      case 'year':
+        return annualPlan
+      case 'month':
+        return monthlyPlan
+      case 'three months':
```
joelhooks, line 189:
```suggestion
      case 'quarter':
```

src/components/pricing/select-plan-new/index.tsx
```diff
@@ -0,0 +1,248 @@
...
+  React.useEffect(() => {
+    forTeams ? setCurrentPlan(teamPlan) : setCurrentPlan(annualPlan)
+  }, [forTeams, annualPlan])
+
+  return (
+    <>
+      <div className="dark:text-white text-gray-900 dark:bg-gray-900 bg-white sm:p-10 p-5 flex flex-col items-center max-w-sm relative z-10 rounded-sm">
+        <PlanTitle>{currentPlan?.name}</PlanTitle>
+        <PlanPrice pricesLoading={pricesLoading} plan={currentPlan} />
+        {keys(prices).length > 1 && (
```
joelhooks, line 217: if we don't have more than one price there's nothing to switch on.

This also "jumps" on page load, so we probably want to stabilize it somehow?

src/hooks/use-pricing.ts
```diff
@@ -1,22 +1,19 @@
 import * as React from 'react'
-import {loadPrices} from 'lib/prices'
+import {loadPrices, Prices} from 'lib/prices'
 import queryString from 'query-string'
 
-export type Prices = {
```
joelhooks, line 5: this was defined twice

joelhooks (approved):

src/components/pricing/index.tsx
vojtaholik: done that in this PR: https://github.com/eggheadio/egghead-next/pull/381

PR #369, opened 2021-02-01
Add system preference default for dark mode

zacjones93:
- Checks system preferences on the machine and sets the default

https://user-images.githubusercontent.com/6188161/106482991-0a82d200-647c-11eb-8042-5d01735523a6.mp4

### DIFF
--- a/src/components/app/header.tsx
+++ b/src/components/app/header.tsx
@@ -223,6 +223,13 @@ const DarkModeToggle = () => {
   const {theme, setTheme} = useTheme()
   React.useEffect(() => setMounted(true), [])
 
+  React.useLayoutEffect(() => {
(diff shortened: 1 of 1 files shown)


zacjones93:
Ah, with the `<DarkModeToggle />` being re-mounted on page transitions this is jarring if you switch to the light/dark mode that isn't your system preference

vojtaholik:
tailwind can check for this too https://tailwindcss.com/docs/dark-mode

zacjones93:
@vojtaholik are there reasons why we don't do that? happy to put that PR in if you think it's an upgrade

vojtaholik:
@zacjones93 currently it doesn't support combination of manual toggle and system preference at the same time and can only handle one of them. ;(

PR #379, opened 2021-02-02
prepare banner image for new course

vojtaholik:
this PR adds hero image for @signalnerve's  course 🌩

image


src/components/pages/home/jumbotron/index.tsx
```diff
@@ -88,123 +88,30 @@ const Jumbotron: FunctionComponent<JumbotronProps> = ({resource}) => {
...
-                  className="mt-4 text-gray-200 text-base"
+                  className="mt-4 text-gray-200 text-base max-w-screen-sm"
                 />
               )}
             </div>
           </div>
         </div>
       </div>
-      <UniqueBackground className="absolute left-0 top-0 w-full h-full z-0" />
+      <UniqueBackground className="absolute left-0 top-0 w-full h-full z-0 object-cover" />
```
vojtaholik, line 98: thanks @laurosilvacom 🙏

src/components/pages/home/jumbotron/index.tsx
```diff
@@ -22,10 +22,10 @@ const Jumbotron: FunctionComponent<JumbotronProps> = ({resource}) => {
         },
       }}
     >
-      <div className="absolute top-0 left-0 bg-gradient-to-r from-blue-500 to-blue-400 h-2 w-full" />
-      <div className="relative z-10 px-5 sm:py-16 py-10 sm:text-left text-center">
-        <div className="space-y-5 mx-auto flex items-center justify-center max-w-screen-xl lg:px-8">
-          <div className="flex lg:flex-row flex-col items-center justify-center sm:space-x-5 sm:space-y-0 space-y-5">
+      <div className="absolute top-0 left-0 bg-gradient-to-r from-blue-500 to-blue-400 h-2 " />
+      <div className="relative z-10 px-5 sm:py-16 py-10 sm:text-left text-center w-full">
```
zacjones93, line 26: learned that `w-full` needs to be 'passed' down if you want the UI to be more dynamic as setting a hard value somewhere in the structure will affect the `w-full` elements nested beneath

src/components/pages/home/jumbotron/index.tsx
```diff
@@ -88,123 +88,30 @@ const Jumbotron: FunctionComponent<JumbotronProps> = ({resource}) => {
...
     </div>
   )
 }
 
 const UniqueBackground = ({className}: any) => {
   return (
-    <svg
-      preserveAspectRatio="xMidYMid slice"
+    <Image
+      priority={true}
```
zacjones93, line 106: `priority={true}` makes the image load first!

MaggieAppleton:
Can we nix the bubbles/circles on the orange bit of the background? They feel a bit distracting and pull you away from the content on the right

jh3y:
I didn't find the ones on the orange as distracting as the ones on the shield. Because of the high contrast and the way they are laid out perhaps.

Not sure if there's a thought here perhaps that if the hero styles are kept to a consistent style using vector where possible (_even if on the backdrop only for now_) that gives more potential for making something that can draw the user in or a speaking point. For example, the previous Docker hero had waves and stars. Here, could be vector clouds and sun rays or something like that to communicate "flare".

jh3y:
My 30-second take in Procreate 😂 

Untitled_Artwork

MaggieAppleton:
Yeah I think generally what's happening is there's just too visual noise much going on.
Competing elements that don't necessarily match each other - photo overlay + background pattern + visually dense course illustration. The colours and tones between the background elements and the foreground illustration are different.

I'm going to start having illustrators make the banners at the same time as making the course illustrations so that everything matches. Banners are a very easy expansion once you already have an illustration made and using the same source file will make sure all the elements jive.

jh3y:
Yeah, they're good points! 💯  The orange is quite strong. The photo backdrop details/noise detracts and then it's also inconsistent with previous banners.

Sounds like a plan! 😎

MaggieAppleton:
Quickie idea - something along these lines. Elements that expand out from the course illustration.

Screenshot - 2021-02-03 13 56 05

jh3y:
Is the shield piece something that lives on consistently with the other banners? I like the idea of the dots spanning out. But, the ones in the shield look funny to me. The one "white" one, then the ones below but inside look like rain drops. Thinking out loud 😅

MaggieAppleton:
The 'shield' is the main course illustration

EGH_IntroCloudFlareWorkers_Final_424@2x

It's the 'face' of the course and goes on cards in the UI and the main course page

image

image

Every course gets an illustration that fits into a square frame. They're not all 'shield' shaped but share a common style:

image

laurosilvacom:
@MaggieAppleton are we updating the original banner with this https://github.com/eggheadio/egghead-next/pull/379#issuecomment-772530185? If so, can you please link to the Figma file?

MaggieAppleton:
Figma: https://www.figma.com/file/MHp5ksBLbyi47JqxkXSSd4/banners-and-share-cards?node-id=1%3A2

laurosilvacom:
@MaggieAppleton I had a hard time centering the course illustration to the background banner since they are two different images. But I removed the bubbles/circles from the original.

dealingwith (approved):

PR #380, opened 2021-02-03
First pass at TypeScript curated page

VontaJamal:
Added Feature section with algorithms, design systems, and state management. I can redo it some other kind of way if this isn't good enough.

image


src/components/search/curated/typescript/index.tsx
```diff
@@ -22,14 +22,19 @@ const SearchTypescript = () => {
       topic={{
         label: 'TypeScript',
         name: 'typescript',
-        description: `TypeScript is a typed superset of JavaScript that compiles to plain JavaScript.`,
+        description: `Over the last several years, TypeScript has been coming on strong with the idea that static types get out of your way and provide deep benefits to code bases of all shapes and sizes, while allowing the flexibility of dynamic types when needed.
```
laurosilvacom, line 25: Added a robust description.

src/components/search/curated/typescript/typescript-page-data.ts
```diff
@@ -32,6 +32,14 @@ const typescriptPageData = [
     title: 'Intermediate',
     name: 'Beyond The Basics',
     resources: [
+      {
+        title: 'Build Custom CLI Tooling with oclif and TypeScript',
```
laurosilvacom, line 36: Shortened titles are totally fine since we are augmenting with static data.

laurosilvacom:
I don't' have a team account on Vercel and can't authorize the deploy. But this is ready to go!

PR #391, opened 2021-02-04
Render disabled button if github already connected

jbranchaud:
If someone has already connected with github, we want this button to be/appear disabled so that they don't think it is an action they can do any longer.


src/pages/user/index.tsx
```diff
@@ -6,6 +6,38 @@ import RequestEmailChangeForm from 'components/users/request-email-change-form'
...
+  }
+
+  return (
+    <a
+      href={`${process.env.NEXT_PUBLIC_AUTH_DOMAIN}/users/github_passthrough?client_id=${process.env.NEXT_PUBLIC_CLIENT_ID}&auth_token=${authToken}`}
+      className={`${buttonStyles} focus:outline-none hover:bg-blue-700`}
+    >
+      {buttonText}
+    </a>
+  )
```
jbranchaud, line 38: This is what I came up with for functionally and visually displaying a "disabled" button (considering the active version is a link). It feels a little verbose, so I'm open to another approach. \cc @vojtaholik

joelhooks: would it be possible to add this url to the `viewer` object (`minimal_user.jbuilder`) so we aren't doing complex url building client side?

What do you think about "already connected to github | remove connection" or similar versus a dead button. THis page desperately needs design support!

jbranchaud: Here is how it looks now when you've already connected:

CleanShot 2021-02-08 at 10 15 26@2x

I agree that we should explain the state of their GitHub connection if they have already connected. I have two variations in mind for that.

1. Leave the inactive button and put some text under it explaining:

CleanShot 2021-02-08 at 10 22 11@2x

2. Include only the text explanation and remove the button:

CleanShot 2021-02-08 at 10 23 01@2x

jbranchaud: > would it be possible to add this url to the `viewer` object (`minimal_user.jbuilder`) so we aren't doing complex url building client side?

I don't think it buys us much to shift the construction of this particular URL over to the server. The client_id and authToken are unique to egghead-next. It feels off to me to construct and serve a client-specific URL from the "general-purpose" API layer.

joelhooks:
image

Made a small tweak

joelhooks (approved):

PR #395, opened 2021-02-05
pull jumbotron from sanity

joelhooks:
https://egghead.io/jumbotron is the same component but with a "transformed resource" from Sanity

### DIFF
--- a/src/components/pages/home/jumbotron/index.tsx
+++ b/src/components/pages/home/jumbotron/index.tsx
@@ -95,12 +95,15 @@ const Jumbotron: FunctionComponent<JumbotronProps> = ({resource}) => {
           </div>
         </div>
       </div>
-      <UniqueBackground className="absolute left-0 top-0 w-full h-full z-0 object-cover" />
+      <UniqueBackground
+        className="absolute left-0 top-0 w-full h-full z-0 object-cover"
+        background={resource.background}
+      />
     </div>
   )
 }
 
-const UniqueBackground = ({className}: any) => {
+const UniqueBackground = ({className, background}: any) => {
   return (
     <Image
       priority={true}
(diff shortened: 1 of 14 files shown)


src/components/search/curated/react/sanity-loading-data.tsx
```diff
@@ -1,76 +1,69 @@
...
-          description,
-          type,
-          path
-        }
-        }
-      }[0]
-  `,
+      const data = await sanityClient.fetch(
+        groq`
+          *[slug.current == $slug]{
```
joelhooks, line 17: this was the first query and it's not great, I think I like the `resources` field more than the specified fields (related and such)

src/pages/jumbotron.tsx
```diff
@@ -0,0 +1,56 @@
...
+      </div>
+    </main>
+  )
+}
+
+export default Jumbotron
+
+export const getServerSideProps: GetServerSideProps = async ({res, params}) => {
+  const load = async () => {
+    const data = await sanityClient.fetch(groq`
```
joelhooks, line 25: this query is a lot more digestible

right now we need to do a lot of transformation from the Sanity content model to our local `resource`. this could be simplified considerably if we changed what local components expect.

studio/schemas/resource.js
```diff
@@ -252,6 +230,13 @@ export default {
       type: 'array',
       of: [{type: 'link'}],
     },
+    {
+      name: 'images',
```
joelhooks, line 234: a stack of labeled image urls so you can collect assets and use the labels to query them

src/components/pages/home/jumbotron/index.tsx
```diff
@@ -95,12 +95,15 @@ const Jumbotron: FunctionComponent<JumbotronProps> = ({resource}) => {
           </div>
         </div>
       </div>
-      <UniqueBackground className="absolute left-0 top-0 w-full h-full z-0 object-cover" />
+      <UniqueBackground
+        className="absolute left-0 top-0 w-full h-full z-0 object-cover"
+        background={resource.background}
```
vojtaholik, line 100: we would like to use SVG instead of raster in some cases, so I'm wondering whether that would be possible with Sanity?

vojtaholik (approved):

PR #399, opened 2021-02-06
use a data loader for loading structured sanity data

joelhooks:
The idea is to make structured data simpler to think about by using a data loading service with a result similar to how we've already been structuring data

### DIFF
--- a/src/lib/pages/home-page-data.ts
+++ b/src/lib/pages/home-page-data.ts
@@ -1,0 +1,41 @@
+import {sanityClient} from 'utils/sanity-client'
+import groq from 'groq'
+
+export async function loadHomePageData() {
+  const slugs = {
+    jumbotron: 'primary-jumbotron-cloudflare-workers-intro',
+  }
+  const data = await sanityClient.fetch(
+    groq`
+      {
+        'jumbotron': ${jumbotronQuery},
+      }
+  `,
+    slugs,
+  )
+
+  return data
+}
+
+const jumbotronQuery = groq`
+*[slug.current == $jumbotron][0]{
+  name,
+  title,
+  description,
+  summary,
+  byline, 
+  meta,
+  path,
+  'slug': resources[][0]->_id,
+  'instructor': collaborators[]->[role == 'instructor'][0]{
+    title,
+    'slug': person->slug.current,
+    'name': person->name,
+    'path': person->website,
+    'twitter': person->twitter,
+    'image': person->image.url
+  },
+  'background': images[label == 'background'][0].url,
+  'image': images[label == 'badge'][0].url,
+}
+`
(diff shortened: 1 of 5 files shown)


src/lib/pages/home-page-data.ts
```diff
@@ -0,0 +1,41 @@
+import {sanityClient} from 'utils/sanity-client'
+import groq from 'groq'
+
+export async function loadHomePageData() {
+  const slugs = {
```
joelhooks, line 5: these are named slugs to use as variables in the composed query below you'll see it referenced as `$jumbotron`

this should mean we can swap it out with a different slug and it'll "just work". You can also use objects and reference them like `$object.slug` if you want to get more tricky.

src/pages/jumbotron.tsx
```diff
@@ -21,39 +22,13 @@ export const Jumbotron = ({resource}: {resource: any}) => {
...
-        'image': images[label == 'badge'][0].url,
-      }
-  `)
-    console.log(data)
-    return data
-  }
+  res.setHeader('Cache-Control', 's-maxage=1, stale-while-revalidate')
 
-  const resource = await load()
+  const resources = await loadHomePageData()
```
joelhooks, line 27: load all the data

src/lib/pages/home-page-data.ts
```diff
@@ -0,0 +1,41 @@
...
+*[slug.current == $jumbotron][0]{
+  name,
+  title,
+  description,
+  summary,
+  byline, 
+  meta,
+  path,
+  'slug': resources[][0]->_id,
+  'instructor': collaborators[]->[role == 'instructor'][0]{
```
zacjones93, line 30: `collaborators[]->[role == 'instructor'][0]`

ah this is cool.  Love being able to filter like this.

It's probably good to filter like this on the front end but this also results in the same data `collaborators[0]->`

src/lib/pages/home-page-data.ts
```diff
@@ -0,0 +1,41 @@
...
+const jumbotronQuery = groq`
+*[slug.current == $jumbotron][0]{
+  name,
+  title,
+  description,
+  summary,
+  byline, 
+  meta,
+  path,
+  'slug': resources[][0]->_id,
```
zacjones93, line 29: This could be written as `'slug': resources[0]->_id,` for more simplicity unless there's a nuance I'm missing.

As far as I can tell, `[]` is a [blank filter that defaults to returning every item in the array](https://www.sanity.io/docs/how-queries-work#some-fine-points-on-arrays-and-projections-649d43d7d179) so in this case we are applying a blank filter and then accessing the first element in the array

src/lib/pages/home-page-data.ts
```diff
@@ -0,0 +1,41 @@
...
+  'instructor': collaborators[]->[role == 'instructor'][0]{
+    title,
+    'slug': person->slug.current,
```
zacjones93, line 32: projection is nice, I wonder if they have something like fragments for this type of thing

src/pages/jumbotron.tsx
zacjones93: So much cleaner 🙌

PR #403, opened 2021-02-09
curated css: build on simplified query

vojtaholik:
image


src/components/search/curated/css/index.tsx
```diff
@@ -1,17 +1,49 @@
...
 import SearchCuratedEssential from '../curated-essential'
+import LoadDataFromSanity from './sanity-loading-data'
+import {slice} from 'lodash'
+import Collection from 'components/pages/home/collection'
 
 const SearchCSS = () => {
+  const data = LoadDataFromSanity()
+  const {resources} = data || {}
+  const levels = slice(resources, 0, 3)
+  const rest = slice(resources, 3, resources?.length)
```
vojtaholik, line 12: I wasn't how to pin point collections I want without hardcoding their `slug` or `id`, so I ended up using their order in Sanity.

joelhooks: beginner, intermediate, advanced are "template collections" and should probably be named/slugged specifically and used if present. Alternatively the resource could be slugged "levels" and contain these three as a group in that resource (nested)

This slice approach is worse than hard coding imo

joelhooks: as a preference I like `take` and `drop` over `slice` for this situation https://lodash.com/docs/4.17.15#take

```js
levels = take(resources, 3)
rest = drop(resources, 3)
```

just easier on the eyes imo

src/components/search/curated/css/sanity-loading-data.tsx
```diff
@@ -1,137 +1,55 @@
...
+export default function LoadDataFromSanity() {
   const [data, setData] = React.useState<any>()
   React.useEffect(() => {
-    const slug = 'rad-css'
+    const slug = 'curated-css-resources'
     async function run() {
       const data = await sanityClient.fetch(
         groq`
-          *[slug.current == 'rad-css']{
+        *[slug.current == $slug][0]{
```
joelhooks, line 12: using a variable here isn't buying us much of anything, but it's a step in the right direction. I think we need to consider how we'd query a generic topic page and structure that vs creating a specific resource that represents the page itself.

[groq "outer projections"](https://www.sanity.io/docs/how-queries-work#queries-that-dont-start-with-an-d8cbfd8dd14a) are interesting for this, but I haven't solved the generic query yet.

joelhooks (approved):
starting to look a lot better

I'm still noodling on what gets hard coded and what should be more generic so changes on the Sanity side are where things are adjusted usually.

"how are pages represented?"

PR #406, opened 2021-02-09
Swap React 360 out of React Page

tayiorbeii:
* Change featured course on React page


.env.development
```diff
@@ -1,4 +1,4 @@
-NEXT_PUBLIC_AUTH_DOMAIN="http://egghead.af:5000"
+NEXT_PUBLIC_AUTH_DOMAIN="https://app.egghead.io"
```
laurosilvacom, line 1: Looks good! But this PR shouldn't change the `.env.development` variables.

tayiorbeii: ah! Good catch!

laurosilvacom (approved):

PR #415, opened 2021-02-10
add basic progress history for user

joelhooks:
This adds **rudimentary** course history in what I'd describe as a brute force sort of way to `/user`

I limited it to 5, but the backing data is paginated. My graphql query is the "brute force" aspect of it and we could really stop and consider what progress looks like in gql. cc @theianjones on that one big time

image


src/components/pages/home/in-progress-collection.tsx
```diff
@@ -15,20 +15,25 @@ const InProgressCollection: FunctionComponent<any> = ({collection}) => {
     series,
     slug,
     collection_progress,
+    progress,
     type,
     path,
     items = [],
   } = collection
 
+  const definedProgress = collection_progress || progress
```
joelhooks, line 24: disparity between REST and gql apis

src/lib/users.ts
```diff
@@ -10,3 +16,107 @@ export async function loadCurrentUser(
...
   return user
 }
+
+export async function loadUserProgress(
+  user_id: number,
+  page = 1,
+  per_page = 5,
+  token?: string,
+): Promise<any> {
+  const query = gql`
```
joelhooks, line 26: this query feels like it's way more beefy than it needs to be but I was just trying to get it done and on the page so we can start to consider it

src/pages/user/index.tsx
```diff
@@ -30,11 +32,23 @@ const User: React.FunctionComponent<
...
   const {stripe_customer_id} = account
   const {viewer, authToken} = useViewer()
+  const [progress, setProgress] = React.useState<any>([])
 
   const {email: currentEmail, accounts, providers} = viewer || {}
   const {slug} = get(accounts, '[0]', {})
   const isConnectedToGithub = providers?.includes('github')
 
+  React.useEffect(() => {
+    const loadProgressForUser = async (user_id: number) => {
```
joelhooks, line 42: I didn't paginate this, but the data is paginated and limited to 5 right now. We could make it "infinite scroll" or something like that if we wanted.

src/pages/user/index.tsx
```diff
@@ -83,6 +98,16 @@ const User: React.FunctionComponent<
             stripeCustomerId={stripe_customer_id}
             slug={slug}
           />
+          <div>
+            {progress.map((item) => {
```
joelhooks, line 102: I don't know if this page is the perfect spot for this, but made sense for now

src/components/pages/home/in-progress-collection.tsx
jbranchaud: Could these be normalized in `src/lib/users.ts` and then this component doesn't need to know about API layer concerns?

E.g. before returning `all_progress` from `loadUserProgress`, you could transform it, something like this:

```javascript
// all the graphql fetching stuff ...

const normalizedData = all_progress.data.map(item => {
  const { progress, ...rest } = item;
  return { ...rest, collection_progress: progress }
})
const normalizedAllProgress = { ...all_progress, data: normalizedData };

return normalizedAllProgress;
```

joelhooks: they could be normalized in several places above this as another project

jbranchaud (approved):
It doesn't seem all that rudimentary to me, looks good 👍

theianjones (approved):
Looks like theres some conflicts with `yarn.lock` file

Issue #409, opened 2021-02-10
Email update can fail silently


dealingwith:
A user who attempts to update their email to an email that already has a user associated with it -- or where there is some other error -- gets a 200 error response "Email address change request cannot be processed". No feedback is given to the user in the UI that the update failed.

The error should cause a message to be displayed to the user. They should probably be directed to support at that point.

see https://github.com/eggheadio/egghead-rails/issues/4343

PR #416, opened 2021-02-10
Show error message when we cannot change the email

jbranchaud:
This PR makes it clear whether the email change request went through or not. If we aren't able to make the change automatically, they user is directed to contact support.

There were going to be 3 or 4 related pieces of React state to manage all of this, so I decided to transition it to an XState machine. That makes the whole thing more manageable.

Related issue: https://github.com/eggheadio/egghead-next/issues/409

https://user-images.githubusercontent.com/694063/107583946-470fa580-6bc1-11eb-8064-686959f5067e.mp4

### DIFF
--- a/src/components/users/request-email-change-form.tsx
+++ b/src/components/users/request-email-change-form.tsx
@@ -3,7 +3,7 @@ import {Formik} from 'formik'
 import * as yup from 'yup'
 import axios from 'axios'
 import {useMachine} from '@xstate/react'
-import {Machine, assign} from 'xstate'
+import {Machine, assign, DoneEventObject} from 'xstate'
 
 const emailChangeSchema = yup.object().shape({
   email: yup.string().email().required('enter your email'),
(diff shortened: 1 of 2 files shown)


src/components/users/request-email-change-form.tsx
```diff
@@ -25,23 +30,20 @@ async function requestEmailChange(newEmail: string) {
...
-
-  const [
-    changeRequestSubmittedFor,
-    setChangeRequestSubmittedFor,
-  ] = React.useState<string | null>(null)
+  const [state, send] = useMachine(requestEmailChangeMachine, {
+    services: {
+      requestChange: (_, event: DoneEventObject) => {
+        return requestEmailChange(event.data.newEmail)
+      },
```
jbranchaud, line 37: I'm not sure if I got the typing right here. The compiler didn't like `Event` type created in the other file, so I re-exported this `DoneEventObject` to satisfy it.

joelhooks (approved):
love to see xstate making things clearer, nice one ⭐️

PR #424, opened 2021-02-11
purchase flow: style account creation step

vojtaholik:
pricing-1@2x


MaggieAppleton:
Nice! Good step in the right direction with progress bar at the top and lovely big green ticks.

What happens if they're already logged in with a free account and click 'join' from the pricing page? Just 'pick a plan' > 'payment'?

vojtaholik:
> What happens if they're already logged in with a free account and click 'join' from the pricing page? Just 'pick a plan' > 'payment'?

They'll skip this step entirely and get the Stripe purchase page, which we can't control much.

joelhooks:
we can’t control it at all, which is part of the charm 😂

joelhooks (approved):

PR #447, opened 2021-02-16
Display current/new email, better error handling

jbranchaud:
Updates to the Email Change Request Confirmation page:

1. Display the current and new email before the user confirms.
2. If the token is already expired when the page loads, tell the user and give them options for how to deal with it.

CleanShot 2021-02-16 at 15 19 54@2x

CleanShot 2021-02-16 at 15 19 46@2x

### DIFF
--- a/src/pages/email-change-request/[token].tsx
+++ b/src/pages/email-change-request/[token].tsx
@@ -1,19 +1,80 @@
 import * as React from 'react'
 import LoginRequired from 'components/login-required'
-import axios from 'axios'
 import {useRouter} from 'next/router'
 import {useViewer} from 'context/viewer-context'
 import toast from 'react-hot-toast'
+import axios from 'axios'
+import isEmpty from 'lodash/isEmpty'
+import {
+  getAuthorizationHeader,
+  getTokenFromCookieHeaders,
+  AUTH_DOMAIN,
+} from 'utils/auth'
 
 async function confirmEmailChangeRequest(token: any) {
-  const {data} = await axios.get(
+  const {data} = await axios.patch(
     `${process.env.NEXT_PUBLIC_AUTH_DOMAIN}/api/v1/email_change_requests/${token}`,
+    {},
+    {
+      headers: {...getAuthorizationHeader()},
+    },
   )
 
   return data
 }
 
-const EmailChangeRequest: React.FunctionComponent = () => {
+type emailChangeRequestData = {
+  new_email: string
+  current_email: string
+}
+
+export async function getServerSideProps(context: any) {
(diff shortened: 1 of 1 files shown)


src/pages/email-change-request/[token].tsx
```diff
@@ -1,19 +1,81 @@
...
   return data
 }
 
-const EmailChangeRequest: React.FunctionComponent = () => {
+type emailChangeRequestData = {
+  new_email: string
+  current_email: string
+}
+
+export async function getServerSideProps(context: any) {
```
Comment (jbranchaud) [inline src/pages/email-change-request/[token].tsx:32]: Request the current/new email addresses from the server in `getServerSideProps`. If the token is expired, used, or unknown, we display an error message with options.

src/pages/email-change-request/[token].tsx
```diff
@@ -27,12 +89,65 @@ const EmailChangeRequest: React.FunctionComponent = () => {
...
+                      support@egghead.io
+                    </a>{' '}
+                    to get further help with this request.
+                  </p>
+                </div>
+              )}
               <div className="flex justify-center items-center w-full">
                 <button
-                  className="text-white bg-red-500 border-0 py-2 px-8 focus:outline-none hover:bg-red-600 rounded"
+                  className={cx(
```
Comment (joelhooks) [inline src/pages/email-change-request/[token].tsx:143]: I'm hesitant to bring in a new library for this single use case. We are generally just using template strings for this circumstance. Can you explain why this is needed?

I think template approach is less likely to break Tailwind.

Regardless, need to import its type definitions or add it as a global module if it's essential.

Comment (jbranchaud) [reply src/pages/email-change-request/[token].tsx:143]: It's not essential, I'll switch this over to template strings. I was kinda surprised that classnames wasn't already being used. I'm not aware of any impact it would have on Tailwind.

jbranchaud:
@joelhooks Updated!

joelhooks:
specifically purging, which is looking in specific places for class names

it might be fine, i haven’t tested it


i’m not against it really just noticed the import

jbranchaud:
> specifically purging, which is looking in specific places for class names

It takes a more naive approach than that and looks through entire files (that are listed in purge paths). The thing to avoid is munging together individual classnames. [source](https://tailwindcss.com/docs/optimizing-for-production#writing-purgeable-html)

CleanShot 2021-02-16 at 16 51 12@2x

> i’m not against it really just noticed the import

It was good pushback, this one instance doesn't really justify a new package. The template string x className ergonomics are better than I was expecting.

joelhooks:
feels like when classnames was invented template strings weren't here yet.

joelhooks (approved):

PR #451, opened 2021-02-17
use graphql to load specific props needed for auth instead of whole playlist

joelhooks:
requires https://github.com/eggheadio/egghead-rails/pull/4375 to merge

### DIFF
--- a/src/lib/playlists.ts
+++ b/src/lib/playlists.ts
@@ -1,5 +1,6 @@
 import {GraphQLClient, request} from 'graphql-request'
 import config from './config'
+import getAccessTokenFromCookie from '../utils/get-access-token-from-cookie'
 
 const graphQLClient = new GraphQLClient(config.graphQLEndpoint)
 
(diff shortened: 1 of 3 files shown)


src/lib/playlists.ts
```diff
@@ -94,6 +95,34 @@ export async function loadAllPlaylists() {
   return all_playlists
 }
 
+export async function loadAuthedPlaylistForUser(slug: string) {
```
joelhooks, line 98: don't like this method name, but the result is an almost instant loading of these props for the user

src/pages/playlists/[slug].tsx
```diff
@@ -14,9 +13,9 @@ type PlaylistProps = {
 const Playlist: FunctionComponent<PlaylistProps> = ({
   playlist: initialPlaylist,
 }) => {
-  const {data} = useSWR(initialPlaylist.url, fetcher)
+  const {data} = useSWR(`${initialPlaylist.slug}`, loadAuthedPlaylistForUser)
```
Comment (joelhooks) [inline src/pages/playlists/[slug].tsx:16]: is this in the code path still?

Comment (theianjones) [reply src/pages/playlists/[slug].tsx:16]: If the playlist isnt indexed then users will be able to access `/playlists/:slug`

theianjones (approved):

PR #455, opened 2021-02-18
Tweaks on the Course and Lesson pages

DrShpongle:
1. Fixed Bookmark, Download and RSS buttons sizes on the Course page
01-fix-buttons

2. Add toast message when user adds/removes the course to/from Bookmarks
02-toast

3. Fix SimpleBar styles for the NextUp list on mobile
4. Multiple clicks on Bookmark button are avoided now

### DIFF
--- a/src/components/layouts/collection-page-layout.tsx
+++ b/src/components/layouts/collection-page-layout.tsx
@@ -400,14 +400,14 @@ const CollectionPageLayout: React.FunctionComponent<CoursePageLayoutProps> = ({
                   >
                     <div className="dark:text-gray-900 flex flex-row items-center border px-2 py-1 rounded hover:bg-gray-200 bg-gray-100 transition-colors text-sm xs:text-base">
                       <BookmarkIcon
-                        className={`w-4 h-4 mr-1`}
+                        className="w-4 h-4 mr-1"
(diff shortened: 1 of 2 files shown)


src/components/layouts/collection-page-layout.tsx
```diff
@@ -110,6 +111,7 @@ const CollectionPageLayout: React.FunctionComponent<CoursePageLayoutProps> = ({
 }) => {
   const courseDependencies: any = getDependencies(course.slug)
   const [isFavorite, setIsFavorite] = React.useState(false)
+  const [clickable, setIsClickable] = React.useState(true)
```
vojtaholik, line 114: something like `isButtonDisabled` might be clearer?

src/components/layouts/collection-page-layout.tsx
```diff
@@ -388,26 +390,39 @@ const CollectionPageLayout: React.FunctionComponent<CoursePageLayoutProps> = ({
                 {toggle_favorite_url ? (
                   <button
                     onClick={() => {
```
vojtaholik, line 392: I think we should DRY this component up, might come handy for saving other resources.

joelhooks (approved):

PR #465, opened 2021-02-21
sanity powered blog with mdx support

joelhooks:
Following Sanity examples (kitchen sink, nextjs blog) this adds Sanity backed blog content at `/blog` to replace a lot of the content at `/learn` that would qualify as a traditional "blog post"


studio/deskStructure.js
```diff
@@ -0,0 +1,25 @@
...
+    'person',
+    'software-library',
+    'post',
+    'page',
+    'siteSettings',
+    'author',
+    'category',
+  ].includes(listItem.getId())
+
+export default () =>
```
joelhooks, line 18: this is where you can start to consider custom publishing workflows

studio/schemas/documents/post.js
```diff
@@ -0,0 +1,116 @@
...
+          type: 'reference',
+          to: {
+            type: 'category',
+          },
+        },
+      ],
+    },
+    {
+      name: 'body',
+      type: 'markdown',
```
joelhooks, line 66: I tried to enjoy their "portable text" but MDX has my heart here for sure.

src/pages/blog/[slug].tsx
```diff
@@ -0,0 +1,117 @@
+import * as React from 'react'
+import groq from 'groq'
+import {sanityClient} from 'utils/sanity-client'
+import imageUrlBuilder from '@sanity/image-url'
+import mdxComponents from 'components/mdx'
+import renderToString from 'next-mdx-remote/render-to-string'
```
Comment (joelhooks) [inline src/pages/blog/[slug].tsx:6]: `next-mdx-remote` has limitations, but works very well for this use case. Primarily I believe you can't import so all components need to be pre-defined.

Comment (kentcdodds) [reply src/pages/blog/[slug].tsx:6]: If that bugs you, checkout mdx-bundler by yours truly 😉

joelhooks:
This looks awesome. Any Next examples yet? Would you be up for showing me how to convert (maybe as a stream?)

kentcdodds:
I've not tried it with next yet, though I'm sure it works. A stream would be fun!

PR #469, opened 2021-02-22
Convert SimpleScript CTA to svg and code

zacjones93:
## Changes made
We have an image currently for the CTA and @joelhooks mentioned coding these up being a better solution vs loading in the image.

I got the design pretty close to the image -- didn't use the custom menlo font for John's name, could be updated there. My concern is how it looks just before the small breakpoint. The card gets squished, not sure how to solve that.

## Updated
Learn web development from John Lindquist on egghead  egghead io
Learn web development from John Lindquist on egghead  egghead io

## Current
Learn web development from John Lindquist on egghead  egghead io


vojtaholik:
> My concern is how it looks just before the small breakpoint. The card gets squished, not sure how to solve that.

Happy to push a fix for that later today.

vojtaholik (approved):

PR #474, opened 2021-02-24
ssr for blog posts

joelhooks:


src/pages/blog/[slug].tsx
```diff
@@ -171,39 +187,16 @@ export async function getStaticProps(context: any) {
         [
           require(`rehype-shiki`),
           {
-            theme: `./src/styles/material-theme-dark.json`,
+            theme: 'nord',
```
Comment (joelhooks) [inline src/pages/blog/[slug].tsx:190]: having trouble getting non-default Shiki theme to load using SSR because in node it's looking for a file path

PR #476, opened 2021-02-24
style "membership confirmed" screens

vojtaholik:
There's likely a lot more more we can do here, so please take this as a first step towards improving this page.

screenshots@2x


vojtaholik:
@joelhooks removed the content CTA and highlighted that the next step is to check their inbox.

joelhooks (approved):

PR #484, opened 2021-03-01
Add Playback speed preference to Home screen video

zacjones93:
This was surprisingly easy!

image


src/components/pages/home/index.tsx
```diff
@@ -17,6 +17,8 @@ import axios from 'utils/configured-axios'
 import InProgressCollection from './in-progress-collection'
 import Jumbotron from './jumbotron'
 import LevelUpCTA from '../../survey/level-up-cta'
+import PlaybackSpeedSelect from 'components/pages/lessons/playback-speed-select'
+import {useEggheadPlayerPrefs} from 'components/EggheadPlayer/use-egghead-player'
```
zacjones93, line 21: We have a hook for PlayerPrefs!

src/components/pages/home/index.tsx
```diff
@@ -346,6 +348,9 @@ const EventSchedule: React.FunctionComponent = () => {
 }
 
 function FeaturedVideoCard(props: {video: any}) {
+  const playerRef = React.useRef(null)
+  const {setPlayerPrefs, playbackRate} = useEggheadPlayerPrefs()
```
zacjones93, line 351: grab relevant items

src/components/pages/home/index.tsx
```diff
@@ -389,17 +394,30 @@ function FeaturedVideoCard(props: {video: any}) {
...
             hls_url={props.video.hls_url}
             dash_url={props.video.dash_url}
             subtitlesUrl={props.video.subtitlesUrl}
             width="100%"
             height="auto"
           />
+          <div className="p-5">
+            {playbackRate && (
+              <PlaybackSpeedSelect
+                playbackRate={playbackRate}
```
zacjones93, line 411: pass playbackRate in

src/components/pages/home/index.tsx
```diff
@@ -389,17 +394,30 @@ function FeaturedVideoCard(props: {video: any}) {
...
+                changePlaybackRate={(rate: number) =>
+                  setPlayerPrefs({playbackRate: rate})
```
zacjones93, line 413: set playbackRate when user would like to

joelhooks (changes requested):
visually this isn't acceptable for presentation on the home page. The additional control bar has a specific intent on video pages that may not be present on the home page. 

imo, we'd be better off making this a clickable image that **navigates the user to the video resource page** and starts playing the video instead of tuning an inline viewing experience for this particular use case

vojtaholik:
another possible option would be to enable bitmovin's default (limited) speed controls for this use case (https://github.com/eggheadio/egghead-next/blob/main/src/components/EggheadPlayer/players/Bitmovin.js#L23)

joelhooks:
@zacjones93 what does this look like now?

joelhooks (approved):

zacjones93:
Reverted the style to what it was previously, now the player itself has an option in settings to update speed

theianjones (changes requested):
> we'd be better off making this a clickable image that navigates the user to the video resource page

I agree with Joel that this should be a clickable image that pushes them to the resource page. Talks like this have slides that you want to read and the experience of watching it in such a small frame is poor

joelhooks:
It'd give us a chance to create a "nice" poster frame. I'm strongly in favor of it over a special variant of the player inline.


The only advantage I can see to inline would be if we allowed it to autoplay (muted) for attention. That's a can of worms though.

DrShpongle:
@zacjones93  Is any help needed here?

zacjones93:
Updating #533 to use an image

PR #495, opened 2021-03-09
put husky@4.3.8 back

DrShpongle:


.husky/_/husky.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/sh
```
jbranchaud, line 1: I think this is from the v5 upgrade. This file should get deleted.

DrShpongle: oh, mixed it up with the one under .git folder!

DrShpongle: fixed

jbranchaud (approved):
Thanks for digging into this 🙌

PR #499, opened 2021-03-11
Add team capacity indications on team page

jbranchaud:
- List # of members / # of seats
- Show warning when seats are full and prompt user to upgrade
subscription

CleanShot 2021-03-15 at 11 34 40@2x

### DIFF
--- a/src/pages/team/index.tsx
+++ b/src/pages/team/index.tsx
@@ -9,29 +9,95 @@ import isEmpty from 'lodash/isEmpty'
 interface TeamData {
   inviteUrl: string
   members: Array<any>
+  numberOfMembers: number
+  capacity: number
   isFull: boolean
 }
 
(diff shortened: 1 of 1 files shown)


joelhooks:
technically speaking they should just be able to add qty to their subscription (portal supports this)

**every** subscription is a "team" and Stripe is able to do the math with no trouble.

jbranchaud:
> technically speaking they should just be able to add qty to their subscription (portal supports this)
> 
> **every** subscription is a "team" and Stripe is able to do the math with no trouble.

Oh, that's better. Do you know off the top of your head if we're already doing that anywhere?

Browsing this at the moment: https://stripe.com/docs/billing/subscriptions/integrating-customer-portal#redirect

jbranchaud:
Marking this as No Merge until I get the Billing Portal linked and working for quantity updates.

joelhooks (approved):

PR #506, opened 2021-03-16
Add Matias case study

zacjones93:
This is the first case study to be added to egghead-next. The rest of our case studies live on [howtoegghead](https://howtoegghead.com/instructor/case-studies/) but it seems about time to migrate those over to egghead proper. 

This adds the case study under `/blog`. It would make sense to migrate the case studies under a `/case-studies/` route with the rest of the data.

**og-image**
Matías Hernández-og-image

**case study** Updated
Matías Hernández leveling up by sharing knowledge  egghead io


laurosilvacom:
We've moved away from using MDX files for the blog. We should consider adding the case studies in Sanity to stay consistent IMO.

laurosilvacom (approved):

PR #512, opened 2021-03-22
Get latest invoice to display current charged amount

theianjones:
The yearly subscription is tiered, so you can't access `price.unit_amount` [stripe doc](https://stripe.com/docs/api/prices/object#price_object-unit_amount).

This grabs the `subscriptions.latest_invoice` so that we can display the `amount_due` to the user instead of 0 dollars.

woo

### DIFF
--- a/src/components/users/subscription-details.tsx
+++ b/src/components/users/subscription-details.tsx
@@ -3,6 +3,7 @@ import axios from 'axios'
 import Link from 'next/link'
 import {track} from '../../utils/analytics'
 import {useViewer} from 'context/viewer-context'
+import get from 'lodash/get'
 
 type SubscriptionDetailsProps = {
   stripeCustomerId: string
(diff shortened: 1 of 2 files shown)


src/components/users/subscription-details.tsx
```diff
@@ -44,13 +45,18 @@ const SubscriptionDetails: React.FunctionComponent<SubscriptionDetailsProps> = (
   }, [stripeCustomerId, slug])
 
   const subscriptionName = subscriptionData && subscriptionData.product?.name
+  const subscriptionUnitAmount = get(
+    subscriptionData,
+    'latestInvoice.amount_due',
+    subscriptionData?.price?.unit_amount,
+  )
```
theianjones, line 52: Getting the price to show from the latest invoice feels like a band-aid solution.

src/pages/api/stripe/billing/session.ts
```diff
@@ -21,21 +21,26 @@ const StripeCheckoutSession = async (
         throw new Error(`no session loaded for ${req.query.session_id}`)
 
       const customer = await stripe.customers.retrieve(customer_id, {
-        expand: ['default_source'],
+        expand: ['default_source', 'subscriptions.data.latest_invoice'],
```
theianjones, line 24: you can expand into stripe list objects with this syntax

joelhooks (approved):

PR #515, opened 2021-03-23
update post-css

DrShpongle:


DrShpongle:
seems impossible until storybook will support webpack 5

vojtaholik:
@DrShpongle looks like Storybook have experimental support for webpack 5 which can be enabled today: https://storybook.js.org/blog/storybook-for-webpack-5/

DrShpongle:
Thanks for the reference, I'll look and try to figure it out.

PR #533, opened 2021-04-03
adding new talk

laurosilvacom:
image


zacjones93:
In #484, there's a discussion of replacing the `EggheadPlayer` with the poster image and linking to the talk page vs having the player inline like it is

laurosilvacom:
That makes sense @zacjones93. Closing this issue for now. This update will be included in the home page refresh.

PR #538, opened 2021-04-05
migrate articles from mdx to sanity

laurosilvacom:
Migrating blog post from MDX to Sanity: 

- [x]  Build a Content Management System for an E-commerce Store with Next.js and Sanity
- [x]  Product Images That Don't Byte with the Next.js Image Component
- [x]  Improve Performance with the Object Pool Design Pattern in JavaScript
- [x]  Utility First Tailwind CSS with Sam Selikoff
- [x]  Performance Task Patterns
- [x] Understanding by Design in a Nutshell

Replaced the hard-coded blog index with a page that loads the articles from Sanity. 

image


next.config.js
```diff
@@ -92,6 +92,36 @@ const learnRoutes = [
...
+  },
+  {
+    source: `/learn/ecommerce/product-images-that-dont-byte-with-the-nextjs-image-component`,
+    destination: `/blog/product-images-that-dont-byte-with-the-nextjs-image-component`,
+    permanent: true,
+  },
+  {
+    source: `/learn/javascript/improve-performance-with-the-object-pool-design-pattern-in-javascript`,
+    destination: `/blog/object-pool-design-pattern`,
+    permanent: true,
```
laurosilvacom, line 108: Since articles have been shared, we redirect the old paths.

src/pages/blog/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+        clipRule="evenodd"
+        d="M4 3a2 2 0 0 0-2 2v10a2 2 0 0 0 2 2h12a2 2 0 0 0 2-2V5a2 2 0 0 0-2-2H4zm12 12H4l4-8 3 6 2-4 3 6z"
+        fill="currentColor"
+      />
+    </g>
+  </svg>
+)
+
+const allArticlesQuery = groq`
+*[_type == "post" && publishedAt < now()]|order(publishedAt desc) {
```
laurosilvacom, line 91: This asterisk represents every document in our dataset.  The following parameter filters by type ‘post'. Every document in Sanity is required to have a type, and the type is always in the _type field. 
 
Additionally, it has a publish date that is lesser (i.e. before) whatever time it is at the moment. We also order the results after the publishing data, so that the newest posts end up first in the list.

zacjones93 (approved):
This looks good, there's room for improvement on the index page to curate the articles.

I noticed Matias's case study page is breaking but that looks to be a separate issue. (TwitterTweetEmbed is not defined). Putting in a fix for that

PR #544, opened 2021-04-07
Show Team Name on Generic Invite page when present

jbranchaud:
image


src/pages/team-invite/[token].tsx
```diff
@@ -32,26 +32,40 @@ const handleJoinTeam = async (token: string, router: any) => {
...
+  } else {
+    return (
+      <>
+        {' '}
+        the <span className="font-bold">{teamName}</span> team
+      </>
+    )
+  }
+}
+
```
Comment (jbranchaud) [inline src/pages/team-invite/[token].tsx:47]: If the team name is present:

> You have been invited by owner@email.com to join *the Team Name team*...

If the team name isn't present:

> You have been invited by owner@email.com to join *their team*...

src/pages/team-invite/[token].tsx
```diff
@@ -101,21 +116,44 @@ const TeamInvite: React.FunctionComponent<{
...
 }
 
-export const getServerSideProps: GetServerSideProps = async ({req, params}) => {
-  const token = params && (params.token as string)
-  const viewTeamInviteUrl = `${AUTH_DOMAIN}/api/v1/accounts/team_invite/${token}`
+interface TeamInviteProps {
+  teamOwnerEmail: string
+  teamName: string | undefined
+  inviteToken: string
+}
```
Comment (jbranchaud) [inline src/pages/team-invite/[token].tsx:123]: Add a TypeScript interface for the props to use across both the `getServerSideProps` and the component definition.

Comment (joelhooks) [reply src/pages/team-invite/[token].tsx:123]: we've been preferring type of interface generally only for the sake of consistency

from my early investigation, the only thing that matters really is picking one or the other

what's your thinking about types vs interfaces?

joelhooks (approved):

src/pages/team-invite/[token].tsx
Comment (jbranchaud) [reply src/pages/team-invite/[token].tsx:123]: > what's your thinking about types vs interfaces?

I've been trying to figure this out. I've heard a lot of people say they prefer interfaces (including the [TypeScript Handbook](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces)). On the other hand, I believe types give some more flexibility in terms of doing [intersections and unions](https://twitter.com/elmpp/status/1378167434755051527?s=20).

Either way, consistency is important. I'd prefer to stick to what is already being done in this codebase.

PR #547, opened 2021-04-08
Add pre-configured graphQLClient 

theianjones:
We dont want to show bundled courses to our learners because they are a part of external sites. They wont have access to them unless they've purchased the course.

bundled

### DIFF
--- a/src/lib/accounts.ts
+++ b/src/lib/accounts.ts
@@ -1,6 +1,6 @@
 import {getGraphQLClient} from '../utils/configured-graphql-client'
 
-export async function loadAccount(slug: string) {
+export async function loadAccount(slug: string, token: string) {
   const query = /* GraphQL */ `
     query getAccount($slug: String!) {
       account(slug: $slug) {
@@ -16,7 +16,7 @@ export async function loadAccount(slug: string) {
     slug,
   }
 
-  const graphQLClient = getGraphQLClient()
+  const graphQLClient = getGraphQLClient(token)
 
   try {
     const {account} = await graphQLClient.request(query, variables)
--- a/src/lib/accounts.ts
+++ b/src/lib/accounts.ts
@@ -1,7 +1,7 @@
-import {GraphQLClient} from 'graphql-request'
+import graphQLClient, {
+  setAuthorizationHeaderForClient,
+} from '../utils/configured-graphql-client'
 import config from './config'
-
-const graphQLClient = new GraphQLClient(config.graphQLEndpoint)
 
 export async function loadAccount(slug: string, token: string) {
   const query = /* GraphQL */ `
@@ -19,13 +19,7 @@ export async function loadAccount(slug: string, token: string) {
     slug,
   }
 
-  const authorizationHeader = token && {
-    authorization: `Bearer ${token}`,
-  }
-
-  graphQLClient.setHeaders({
-    ...authorizationHeader,
-  })
+  setAuthorizationHeaderForClient(graphQLClient, token)
 
   try {
     const {account} = await graphQLClient.request(query, variables)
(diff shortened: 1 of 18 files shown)


src/lib/accounts.ts
```diff
@@ -19,13 +19,7 @@ export async function loadAccount(slug: string, token: string) {
...
   }
 
-  const authorizationHeader = token && {
-    authorization: `Bearer ${token}`,
-  }
-
-  graphQLClient.setHeaders({
-    ...authorizationHeader,
-  })
+  setAuthorizationHeaderForClient(graphQLClient, token)
```
theianjones, line 22: `setHeaders` will wipe all the existing headers so I added a helper function that will add the single header to the existing ones.

src/lib/config.ts
```diff
@@ -1,7 +1,16 @@
+const SITE_CLIENT_ID = process.env.NEXT_PUBLIC_CLIENT_ID
+
+const headers = SITE_CLIENT_ID
+  ? {
+      'X-SITE-CLIENT': SITE_CLIENT_ID,
```
theianjones, line 5: this is how we scope our requests to egghead content

src/lib/accounts.ts
joelhooks: another approach would be to `const graphQLClient = getGraphClient(token)` vs importing the instance and then updating it later.

src/utils/configured-graphql-client.ts
```diff
@@ -0,0 +1,21 @@
...
+  client: GraphQLClient,
+  authToken: string | undefined,
+) => {
+  if (authToken) {
+    const authorizationHeader = `Bearer ${authToken}`
+    client.setHeader('authorization', authorizationHeader)
+  }
+}
+
+export const getGraphQLClient = (token?: string) => {
```
joelhooks, line 14: this should additionally check to see if a token is available and use it if one isn't passed in.

there are no circumstances where we want to avoid the auth if it is available.

jbranchaud, line 14: I'm glad you tackled this refactor 🙌 I almost scratched this itch the other day when setting up the boilerplate for a graphql request.

src/utils/configured-graphql-client.ts
```diff
@@ -0,0 +1,17 @@
...
+    headers: config.headers,
+  })
+
+  const authToken = getAccessTokenFromCookie()
+
+  if (authToken) {
+    graphQLClient.setHeader('Authorization', `Bearer ${authToken}`)
+  }
+
+  return graphQLClient
```
theianjones, line 16: consolidated the graphql client to this function. It will automatically add the `x-site-client` header and add the auth token as well if its available.

src/pages/team/index.tsx
```diff
@@ -241,11 +240,7 @@ const IconLink: FunctionComponent<{className?: string}> = ({
 export const getServerSideProps: GetServerSideProps<TeamPageProps> = async function (
   context: any,
 ) {
-  const {eggheadToken} = getTokenFromCookieHeaders(
-    context.req.headers.cookie as string,
-  )
```
jbranchaud, line 246: The `getAccessTokenFromCookie` call works for client-side requests. In these server-side requests where we call `getTokenFromCookieHeaders`, you need to make the call where you have access to the `context.req`.

Perhaps the thing to do is for `getGraphQLClient` to accept an optional `token` argument and fallback to `getAccessTokenFromCookie` if that isn't passed in.

joelhooks: yes, that is likely the case

if a token is passed in, use it. If it isn't, check the client cookies and use that if available

src/utils/configured-graphql-client.ts
theianjones: changed it to accept a token like you suggested!

joelhooks (approved):
⭐️

PR #548, opened 2021-04-09
Load Digital Gardening Section From Sanity

zacjones93:
@laurosilvacom, @theianjones, @MaggieAppleton, @dealingwith got on a call to structure the `Digital Gardening for Developers` section data on the home page in Sanity and subsequently load that data.

It should look exactly like it does in production.

### DIFF
--- a/src/components/pages/home/homepage-data.ts
+++ b/src/components/pages/home/homepage-data.ts
@@ -1,36 +1,5 @@
 const homepageData = [
   {
-    id: 'digital-gardening-featured',
-    title: 'Featured',
-    resources: [
-      {
-        name: 'Content Authoring',
-        title: 'Creating a Digital Garden CLI with Rust',
-        byline: 'Chris Biscardi・1h 10m',
-        image:
-          'https://d2eip9sf3oo6c2.cloudfront.net/playlists/square_covers/000/429/801/full/rust-garden-cli_424_2x.png',
-        path: '/courses/creating-a-digital-garden-cli-with-rust-34b8',
-      },
-      {
-        name: 'Start a Blog',
-        title: 'Build a Developer Blog with Gatsby',
-        byline: 'Laurie Barth・35m',
-        image:
-          'https://d2eip9sf3oo6c2.cloudfront.net/playlists/square_covers/000/425/621/full/developer_blog_1000_2x.png',
-        path: '/courses/build-a-developer-blog-with-gatsby-bd96',
-      },
-      {
-        name: 'Create a Portfolio',
-        title: 'Build a site from scratch with Next.js',
-        byline: 'Tomasz Łakomy・37m',
-        image:
-          'https://d2eip9sf3oo6c2.cloudfront.net/playlists/square_covers/000/387/163/square_480/netlify-ts.png',
-        path:
-          '/playlists/build-a-blog-with-next-js-typescript-emotion-and-netlify-adcc',
-      },
-    ],
-  },
-  {
     id: 'jumbotron',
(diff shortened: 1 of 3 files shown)


src/components/pages/home/homepage-data.ts
```diff
@@ -1,35 +1,4 @@
...
-        name: 'Create a Portfolio',
-        title: 'Build a site from scratch with Next.js',
-        byline: 'Tomasz Łakomy・37m',
-        image:
-          'https://d2eip9sf3oo6c2.cloudfront.net/playlists/square_covers/000/387/163/square_480/netlify-ts.png',
-        path:
-          '/playlists/build-a-blog-with-next-js-typescript-emotion-and-netlify-adcc',
-      },
-    ],
-  },
```
zacjones93, line 32: dont need this hard coded no longer!

src/pages/index.tsx
```diff
@@ -18,11 +21,44 @@ const IndexPage: FunctionComponent = () => {
...
+    description
+  },
+  resources[]{
+    title,
+    byline,
+    'name': content[title == 'name'][0].description,
+    'path': resources[]->[0].path,
+    'image': resources[]->[0].image
+  }
+}
```
zacjones93, line 51: thanks to @theianjones we don't need to transform the data on the front end.

This queries the feature data for this section. When we have more section (paths) we want to build out, we can parameterize this query.

src/pages/index.tsx
```diff
@@ -18,11 +21,44 @@ const IndexPage: FunctionComponent = () => {
...
+`
+
+export async function getStaticProps() {
+  const digitalGardeningFeature = await sanityClient.fetch(
+    digitalGardeningQuery,
+  )
+
+  return {
+    props: {
+      sections: [digitalGardeningFeature],
```
zacjones93, line 61: making `sections` an array because we don't want to make the data coming in tied to the view

src/components/pages/home/index.tsx
```diff
@@ -18,7 +18,9 @@ import InProgressCollection from './in-progress-collection'
 import Jumbotron from './jumbotron'
 import LevelUpCTA from '../../survey/level-up-cta'
 
-const Home: FunctionComponent = () => {
+const Home: FunctionComponent<any> = ({sections}) => {
+  const [featureCallOut] = sections
```
zacjones93, line 22: We can call this whatever we want

src/pages/index.tsx
```diff
@@ -18,11 +21,44 @@ const IndexPage: FunctionComponent = () => {
...
+          <Home sections={sections} />
         </div>
       </main>
     </>
   )
 }
 
 export default IndexPage
+
+const digitalGardeningQuery = groq`
```
joelhooks, line 33: nest step: this query could be organized outside of this file and imported so you'd end up with a library of queries for pulling in specifically shaped data for presentation. It might make sense to pair shaped data like this to a named component (which might be a page?) as well

src/components/pages/home/index.tsx
joelhooks, line 22: suggestion:  think this array aliasing approach is probably not great. I'd consider using named object, similar to `homePageData`

`sections` would be an object and not an array so the names would stay consistent and not "well, the first one in the array is always X and the 7th is Y" situation.

src/pages/index.tsx
joelhooks: suggestion: in my heart of hearts I think we need to consider features as pages

joelhooks (approved):
I left a couple of comments below but they aren't deal breakers. Remove RFC when you want it to deploy.

PR #551, opened 2021-04-12
Add In progress component to homepage

MaggieAppleton:
## In-progress Homepage Components

This adds an "in progress" component to the homepage.
If a user is logged in, the three most recent courses they've started will appear at the top of the page:

Screenshot 2021-04-12 at 12 19 41

Mobile view:

Screenshot 2021-04-12 at 12 19 41

<br/>
<br/>

If the user is not logged in, they'll just see the jumbotron:

Screenshot 2021-04-12 at 12 21 17

<br/>
<br/>

## New In-progress components

To give us a bit more flexibility, I created a second version of the `in-progress-collection` component, called `mini-progress-collection`
This makes a smaller version - tighter spacing, a smaller image, and doesn't list the lesson name of the course in progress.
I increased the image size and spacing on the larger versions to create contrast between the two.

Larger `in-progress-collection` component:
image

Smaller `mini-progress-collection` component:
image

This could probably also be done by passing a prop, but figured this was a good enough solution for now.

<br/>
<br/>

## Time left display error

For some reason, the time remaining in the course isn't showing up on the in-progress cards.

Line 102 in `src/components/pages/home/in-progress-collection.tsx`
Line 95 in `src/components/pages/home/mini-progress-collection.tsx`

Both use this snippet:

```
          {isInProgress && (
            <h2 className=" uppercase font-semibold text-xs text-gray-600 dark:text-gray-300 mb-1">
              Lesson {lessons_left} of {lesson_count}{' '}
              {time_left
                ? `| ${convertTimeWithTitles(time_left)} minutes left`
                : ''}
            </h2>
          )}
```

But when the cards render, `time_left` doesn't exist, so it isn't showing up

<br/>
<br/>

H/t @dealingwith for helping me through the tougher bits of this.

giphy


vojtaholik:
LGTM

I made a couple minor adjustments such as adding a `small` prop to `in-progress-collection` component and handling case when `viewer.name` is empty.  looks like `time_left` isn't currently present in our graphql endpoint.

one concern I have is that it takes quite a while for the user progress to load, resulting in a significant layout shift. looking into adding a skeleton loader right now

MaggieAppleton:
Cool, I just wrote up a storybook component for the collections.

I also isolated the whole in-progress section into a `<InProgressSection />` component.
Was trying to get it to render as a single storybook component, but couldn't load in the dummy data without storybook throwing a fit.
So the component works when imported into the home index page, but doesn't have a storybook story

vojtaholik:
I think it would be better to use story that already exists for this component: https://design.egghead.io/?path=/story/components-users-dashboard-activity--continue-watching

MaggieAppleton:
> I think it would be better to use already existing story for this component: https://design.egghead.io/?path=/story/components-users-dashboard-activity--continue-watching

🤦🏻‍♀️ Sorry, I completely missed that this already exists. Will go remove the duplicate, lol.

MaggieAppleton:
How and where do you grab the data to auto-fill in a storybook component like this?

Eg. the dummy data for `completedLessons` and `inProgressSeries` that you pass tp the components in pages/users/dashboard/activity/activity.stores.mdx

MaggieAppleton:
I tried to combine the in-progress components in /home and /user/dashboard since they were duplicates. I screwed up the way individual lessons display, so the image isn't rendering. Tried to fix it but I'm very lost at this point.

vojtaholik:
> How and where do you grab the data to auto-fill in a storybook component like this?

I grabbed them [here](https://github.com/eggheadio/egghead-next/blob/main/src/components/pages/users/dashboard/activity/activity.stories.mdx#continue-watching). But since `loadUserProgress` is using graphql endpoint, you can [take the query](https://github.com/eggheadio/egghead-next/blob/main/src/lib/users.ts#L24-L97) and run it on `app.egghead.af:5000/graphiql`

vojtaholik:
this is kind of a weird state which happens when watching lessons in random order rather than linearly:


so I was thinking we could symbolize individual lessons with links and highlight completed ones like so:


what do you think @MaggieAppleton?

MaggieAppleton:
Yeah switching the blue progress bar to individual highlighted sections for each lesson would make much more sense. I had a hard time figuring out whether it was displaying overall progress, or individual lesson progress.

The new UI mockup is a big improvement over the existing version 👍

vojtaholik (approved):
go go go

dealingwith:
Has everyone done the CLA or does it just need to be kicked? 😄

PR #552, opened 2021-04-12
Fix Team page dark-mode and clean up some types

jbranchaud:
The dark mode styles weren't apply the Team page because it was wrapped in `<div className="prose">`. Once I added in the `dark:prose-dark` class name, most of it clicked into place. There were individual items that needed to be adjusted after that.

Here is the Before:

CleanShot 2021-04-08 at 09 34 51@2x

And the After:

CleanShot 2021-04-12 at 13 03 55@2x


src/components/team/copy-to-clipboard.tsx
```diff
@@ -0,0 +1,58 @@
+import * as React from 'react'
```
jbranchaud, line 1: The Team page file was getting large, so I extracted this to a standalone component to reduce the noise and remove clear up some imports.

src/lib/teams.ts
```diff
@@ -1,6 +1,8 @@
 import {getGraphQLClient} from '../utils/configured-graphql-client'
 
-export async function loadTeams(token: string) {
+type TeamResponse = {data: Array<any> | undefined}
+
+export async function loadTeams(token: string): Promise<TeamResponse> {
```
jbranchaud, line 5: Makes the typing/type errors clearer.

src/pages/team/index.tsx
```diff
@@ -104,7 +103,7 @@ const AtCapacityNotice = ({teamData}: {teamData: TeamData}) => {
               onClick={() => {
                 track(`clicked manage membership`)
               }}
-              className="transition-all duration-150 ease-in-out text-white font-semibold rounded-md"
+              className="transition-all duration-150 ease-in-out font-semibold rounded-md dark:text-yellow-400 dark:hover:text-yellow-300"
```
jbranchaud, line 106: I'm struggling to get this bit of dark-mode styling to apply. I don't know how tailwindcss precedence is determined, but the prose dark-mode light blue link style is taking precedence over these element-specific styles.

The yellow looks a lot better against the burnt orange, than the light blue does.

@DrShpongle @vojtaholik any ideas on how to make the yellow apply?

src/pages/team/index.tsx
```diff
@@ -120,21 +119,20 @@ type TeamPageProps = {
...
     }
-  }, [loading, teamDataAvailable])
+  }, [teamDataNotAvailable])
 
   return (
     <LoginRequired>
-      {!loading && !!teamData && (
-        <div className="lg:prose-lg prose xl:prose-xl max-w-screen-xl mx-auto mb-24">
+      {!!teamData && (
+        <div className="lg:prose-lg prose dark:prose-dark xl:prose-xl max-w-screen-xl mx-auto mb-24">
```
jbranchaud, line 135: This was the primary issue: use `dark:prose-dark`

DrShpongle:
@jbranchaud I think it's better to get rid of `prose` class at all here instead of override it in some way. Let me tackle it. I'll push in your branch, does it work?

jbranchaud:
> @jbranchaud I think it's better to get rid of `prose` class at all here instead of override it in some way. Let me tackle it. I'll push in your branch, does it work?

@DrShpongle yes, feel free to push to this branch. Thanks for your help!

DrShpongle (approved):

PR #553, opened 2021-04-13
Extract currency and fix presence check

jbranchaud:
The presence check for the `subscriptionPrice` should involve the unit
amount and currency derived in the statements above.


theianjones (approved):
👍

PR #554, opened 2021-04-13
rework how data is imported from sanity for home page

joelhooks:
recorded a video: https://www.loom.com/share/a774bd04f14342f2a9c62d17a1f216f7

### DIFF
--- a/src/components/pages/home/homepage-data.ts
+++ b/src/components/pages/home/homepage-data.ts
@@ -1,5 +1,5 @@
-const homepageData = [
-  {
+const homepageData = {
+  jumbotron: {
     id: 'jumbotron',
     title: 'Create a Digital Garden CLI with Rust',
     byline: 'new course',
@@ -19,7 +19,7 @@ const homepageData = [
         'https://d2eip9sf3oo6c2.cloudfront.net/instructors/avatars/000/000/275/square_128/headshot_post_gym_dec_2018_500.png',
     },
   },
-  {
+  'modern-layouts-with-css-grid': {
     id: 'modern-layouts-with-css-grid',
     title: 'Build Modern Layouts with CSS Grid',
     name: 'Write modern CSS',
@@ -40,7 +40,7 @@ const homepageData = [
         'https://d2eip9sf3oo6c2.cloudfront.net/instructors/avatars/000/000/458/medium/IMG_20190627_100655_466.jpg',
     },
   },
-  {
+  video: {
     id: 'video',
     name: 'Optimize your Learning',
     title: 'Learning Tips Every Developer Should Know',
@@ -63,7 +63,7 @@ There are a number of tried and true methods that will allow you to learn more e
     subtitlesUrl:
       'https://app.egghead.io/api/v1/lessons/egghead-egghead-talks-learning-tips-every-developer-should-know/subtitles',
   },
-  {
+  nextjsVideoApp: {
     id: 'nextjsVideoApp',
     name: 'Create AI-powered Transcripts',
     title: 'Video Analysis App with NextJS and Symbl.ai',
@@ -76,7 +76,7 @@ There are a number of tried and true methods that will allow you to learn more e
     description: `Add Conversation Intelligence to your video or voice application without training Machine Learning models.
     `,
   },
-  {
+  featured: {
     id: 'featured',
     title: 'Featured',
     resources: [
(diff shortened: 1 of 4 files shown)


src/components/pages/home/homepage-data.ts
```diff
@@ -1,5 +1,5 @@
-const homepageData = [
-  {
+const homepageData = {
```
joelhooks, line 1: as an object we don't need an `id` prop since we can just access it as a prop, which is in fact an id!

src/components/pages/home/index.tsx
```diff
@@ -7,61 +7,44 @@ import {map, get, find, isEmpty} from 'lodash'
...
-  const [featureCallOut] = sections
 
+const Home: FunctionComponent<any> = ({homePageData}) => {
   const location = 'home landing'
   const {viewer, loading} = useViewer()
   const [currentCourse, setCurrentCourse] = React.useState<CardResource>()
   const currentCourseUrl = viewer?.current_course?.url
 
-  const video: any = find(homepageData, {id: 'video'})
+  const video: any = get(homePageData, 'video')
```
joelhooks, line 24: we use `get` instead of `find` because they are just props

src/components/pages/home/index.tsx
```diff
@@ -7,61 +7,44 @@ import {map, get, find, isEmpty} from 'lodash'
...
 import useEggheadSchedule, {ScheduleEvent} from 'hooks/use-egghead-schedule'
 import {track} from 'utils/analytics'
 import Collection from './collection'
 import axios from 'utils/configured-axios'
 import InProgressCollection from './in-progress-collection'
 import Jumbotron from './jumbotron'
-import LevelUpCTA from '../../survey/level-up-cta'
-
-const Home: FunctionComponent<any> = ({sections}) => {
-  const [featureCallOut] = sections
```
joelhooks, line 22: removed this concept entirely

src/components/pages/home/index.tsx
```diff
@@ -7,61 +7,44 @@ import {map, get, find, isEmpty} from 'lodash'
 import Textfit from 'react-textfit'
 import Markdown from 'react-markdown'
 import {useViewer} from 'context/viewer-context'
-import homepageData from './homepage-data'
```
joelhooks, line 10: homepageData is now imported at the index and passed in as a prop

src/pages/index.tsx
```diff
@@ -30,35 +31,21 @@ const IndexPage: FunctionComponent = ({sections}: any) => {
...
-  resources[]{
-    title,
-    byline,
-    'name': content[title == 'name'][0].description,
-    'path': resources[]->[0].path,
-    'image': resources[]->[0].image
-  }
+const featureQuery = groq`
+{
+  'featureCallOut': ${digitalGardeningQuery}
```
joelhooks, line 36: we can construct stacks of queries that resolve into named properties and are also used elsewhere like the features page!

src/pages/learn/digital-gardening/index.tsx
```diff
@@ -0,0 +1,38 @@
+import * as React from 'react'
+import groq from 'groq'
+import {sanityClient} from 'utils/sanity-client'
+
+const DigitalGardening: React.FC<any> = ({data}) => {
```
joelhooks, line 5: instead of transient "features" what if these were "page first"?

laurosilvacom (approved):

src/pages/index.tsx
```diff
@@ -30,35 +31,21 @@ const IndexPage: FunctionComponent = ({sections}: any) => {
...
-    digitalGardeningQuery,
-  )
+  const sanityHomePageData = await sanityClient.fetch(featureQuery)
 
   return {
     props: {
-      sections: [digitalGardeningFeature],
+      homePageData: {
+        ...staticHomePageData,
+        ...sanityHomePageData,
```
zacjones93, line 47: ahh this is nice

src/pages/learn/digital-gardening/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+  return <h1>{data.title}</h1>
+}
+
+export default DigitalGardening
+
+export const digitalGardeningQuery = groq`*[_type == 'resource' && slug.current == "digital-gardening-for-developers"][0]{
```
zacjones93, line 11: I also love the idea that the query lives with the page and you can import that in other places that you'd like to use it.

PR #559, opened 2021-04-15
Add billing summary to team page

jbranchaud:
Add Billing Summary details to the team page so that a team owner can see what their plan looks like, what they are paying, and click through to the Stripe Billing Portal to manage it all.

Here is what it looks like:

CleanShot 2021-04-15 at 10 50 24@2x

Here is the original mock up. Note that I didn't match the details from this mock up exactly. I started with a simpler approach that puts the core information on the screen.

Depending on what details are important, we can take a second pass on this to include any or all of:
- Discount information
- Computed monthly breakdown for each learner

(Invoices, also pictured in this mock up, are coming up next.)

image

### DIFF
--- a/src/components/team/billing-section.tsx
+++ b/src/components/team/billing-section.tsx
@@ -5,6 +5,19 @@ import {format} from 'date-fns'
 import Link from 'next/link'
 import {track} from '../../utils/analytics'
 
+const formatAmountWithCurrency = (
+  amountInCents: number,
+  currency: string,
+): string => {
+  if (!amountInCents || !currency) return ''
+
+  return new Intl.NumberFormat('en-US', {
+    style: 'currency',
+    currency: currency,
+    minimumFractionDigits: 0,
+  }).format(amountInCents / 100)
+}
+
 const BillingSection = ({
   stripeCustomerId,
   slug,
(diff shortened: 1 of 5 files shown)


src/components/users/subscription-details.tsx
```diff
@@ -15,34 +15,10 @@ const SubscriptionDetails: React.FunctionComponent<SubscriptionDetailsProps> = (
...
-            account_slug: slug,
-          },
-        })
-        .then(({data}) => {
-          if (data) {
-            setSubscriptionData(data)
-          }
-        })
-    }
-  }, [stripeCustomerId, slug])
```
jbranchaud, line 45: Moving these details into a shared hook

src/components/team/billing-section.tsx
```diff
@@ -0,0 +1,100 @@
+import * as React from 'react'
+import useSubscriptionDetails from 'hooks/use-subscription-data'
+import get from 'lodash/get'
+import {format} from 'date-fns'
+import Link from 'next/link'
+import {track} from '../../utils/analytics'
+
+const BillingSection = ({
```
jbranchaud, line 8: This billing section displays toward the bottom of the Team Page. It lets the team owner know:
- the amount on their latest invoice
- the next billing data
- how many learners are available through their subscription

src/components/team/billing-section.tsx
```diff
@@ -0,0 +1,100 @@
...
+            href="mailto:support@egghead.io"
+          >
+            support@egghead.io
+          </a>
+        </span>
+      </div>
+      {!loading && (
+        <div className="flex flex-col space-y-2 border border-gray-300 mt-4 p-2">
+          <div className="flex flex-row justify-between">
+            <span>egghead team annual</span>
```
jbranchaud, line 65: I hardcoded this here. For team accounts is there a non-annual option?

joelhooks:
I think they can choose any of the durations. they are all tiered.

jbranchaud:
I've updated the styles a little and factored in both tiered pricing and the upcoming invoice to get an accurate display of billing details.

For instance, here is the dark-mode display for a monthly subscription with 3 members and then after it gets upgraded to 5 members.

CleanShot 2021-04-19 at 13 57 52@2x

CleanShot 2021-04-19 at 13 59 36@2x

Here is a person with a yearly team subscription with 3 seats:

CleanShot 2021-04-19 at 14 07 14@2x

And a quarterly sub with 5 seats which qualifies for a lower tiered price:

CleanShot 2021-04-19 at 14 08 47@2x

And here is what the Billing Details look like after a subscription has been cancelled:

CleanShot 2021-04-19 at 14 32 16@2x

joelhooks (approved):
this looks fantastic

PR #565, opened 2021-04-19
Add right padding to searchbar input

zacjones93:
The X button felt a little cramped being so close to the search button

## Before:
image

## After:
image


vojtaholik (approved):
good catch!

PR #575, opened 2021-04-21
update digital garden feature to use named sanity resources

zacjones93:
This is using a restructured Sanity object where we are querying for the resource title so that we can add resources without breaking the front end.

One thing that I don't understand... `Create a digital garden CLI with Rust` course image and path aren't being returned from the API even though they are present in the query in sanity studio

image

### DIFF
--- a/src/pages/learn/digital-gardening/index.tsx
+++ b/src/pages/learn/digital-gardening/index.tsx
@@ -142,18 +142,6 @@ export const digitalGardeningQuery = groq`*[_type == 'resource' && slug.current 
       image,
     },
   },
-  related[0]{
-    title,
-    description,
-    'cta': content[title == 'cta'][0]{
-      description
-    },
-    resources[]{
-      title,
-      'path': slug.current,
-      byline,
-      image,
-    },
   },
 }`
 
--- a/src/pages/learn/digital-gardening/index.tsx
+++ b/src/pages/learn/digital-gardening/index.tsx
@@ -8,6 +8,7 @@ import Card from 'components/pages/home/card'
 import {track} from 'utils/analytics'
 
 const DigitalGardening: React.FC<any> = ({data}) => {
+  console.log({courses: data.featured.courses})
   return (
     <div className="sm:-my-5 -my-3 -mx-5 p-5 dark:bg-gray-900 bg-gray-50">
       <div className="mx-auto max-w-screen-xl">
@@ -63,7 +64,7 @@ const DigitalGardening: React.FC<any> = ({data}) => {
 
         <div>
           <div className="grid lg:grid-cols-12 grid-cols-1 gap-5 mt-5">
-            {data.resources.map((resource: any) => {
+            {data.featured.courses.map((resource: any) => {
               return (
                 <Card
                   className="col-span-4 text-center"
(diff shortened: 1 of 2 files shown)


src/pages/learn/digital-gardening/index.tsx
```diff
@@ -108,7 +108,7 @@ const DigitalGardening: React.FC<any> = ({data}) => {
 
 export default DigitalGardening
 
-export const digitalGardeningQuery = groq`*[_type == 'resource' && slug.current == "digital-gardening-for-developers"][0]{
+export const digitalGardeningQuery = groq`*[_type == 'resource' && slug.current == "digital-gardening-for-developers-v2"][0]{
```
zacjones93, line 111: querying v2 for now, slugs have to be unique in sanity

src/pages/learn/digital-gardening/index.tsx
```diff
@@ -122,12 +122,25 @@ export const digitalGardeningQuery = groq`*[_type == 'resource' && slug.current
...
-  resources[]{
+  'featured': resources[title == 'Featured digital gardening courses'][0]{
+ 		'courses': resources[]{
+    	title,
+    	byline,
+      'name': content[title == 'name'][0].description,
+    	'path': resources[]->[0].path,
+    	'image': resources[]->[0].image
+  	}
+  },
```
zacjones93, line 133: This displays weird on github but is formatted as you would expect.

resources is now a query for the resource titled 'Featured digital gardening courses'. Now we can add other sections to the page without the front end automatically rendering what it finds.

src/pages/learn/digital-gardening/index.tsx
```diff
@@ -63,7 +63,7 @@ const DigitalGardening: React.FC<any> = ({data}) => {
 
         <div>
           <div className="grid lg:grid-cols-12 grid-cols-1 gap-5 mt-5">
-            {data.resources.map((resource: any) => {
+            {data.featured.courses.map((resource: any) => {
```
zacjones93, line 66: rendering featured courses on the page

src/pages/learn/digital-gardening/index.tsx
```diff
@@ -122,12 +122,25 @@ export const digitalGardeningQuery = groq`*[_type == 'resource' && slug.current
...
+  'featured': resources[title == 'Featured digital gardening courses'][0]{
+ 		'courses': resources[]{
+    	title,
+    	byline,
+      'name': content[title == 'name'][0].description,
+    	'path': resources[]->[0].path,
+    	'image': resources[]->[0].image
+  	}
+  },
+	'talks': resources[title == 'Infrastructure for Digital Gardens'][0]{
```
zacjones93, line 134: moved the talks up to the resources section (previously in related resources) and now filtering by resource name

src/pages/learn/digital-gardening/index.tsx
```diff
@@ -80,17 +80,17 @@ const DigitalGardening: React.FC<any> = ({data}) => {
             <div className="flex lg:flex-row flex-col items-center justify-center sm:space-x-10 sm:space-y-0 space-y-5 0 w-full xl:pr-16">
               <div className="mx-auto">
                 <h2 className="text-xs text-yellow-600 dark:yellow-green-300 uppercase font-semibold mb-2 text-center">
-                  {data.related.cta.description}
+                  {data.talks.cta}
```
zacjones93, line 83: similarly iterating over talks

src/pages/learn/digital-gardening/index.tsx
joelhooks: ya this is better imo. i think query by title is brittle and we should use _IDs or slugs

src/pages/learn/digital-gardening/index.tsx
laurosilvacom: This approach makes more sense!

zacjones93:
I'm not sure what issue I was running into but I solved it by deleting the resource reference to the Rust CLI course and adding it back in. It's like the reference was corrupted somehow

src/pages/learn/digital-gardening/index.tsx
```diff
@@ -122,12 +122,25 @@ export const digitalGardeningQuery = groq`*[_type == 'resource' && slug.current
...
+    description,
+    'cta': content[title == 'cta'][0].description,
+		resources[]{
+      title,
+      'path': slug.current,
+      byline,
+      image,
+    },
   },
   related[0]{
```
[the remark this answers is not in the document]
zacjones93: I deleted this! good catch, that part of the query was returning no data

laurosilvacom (approved):

PR #570, opened 2021-04-20
AWS curated landing page

wjohnson85:
Curated AWS Landing page 

image

image

PR #578, opened 2021-04-22
search: correctly count refinements + scroll to hits

vojtaholik:
fixes https://github.com/eggheadio/egghead-next/pull/570#issuecomment-824432785

### DIFF
--- a/src/components/search/index.tsx
+++ b/src/components/search/index.tsx
@@ -72,8 +72,7 @@ const Search: FunctionComponent<SearchProps> = ({
   const numberOfRefinements =
     get(searchState, 'refinementList.instructor_name', []).length +
     get(searchState, 'refinementList._tags', []).length +
-    get(searchState, 'refinementList.type', []).length +
-    get(searchState, 'page', 0)
+    get(searchState, 'refinementList.type', []).length
 
   const refinementRef = React.useRef(null)
   useClickAway(refinementRef, () => setShowFilter(false))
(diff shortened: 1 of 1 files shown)


src/components/search/index.tsx
```diff
@@ -72,8 +72,7 @@ const Search: FunctionComponent<SearchProps> = ({
   const numberOfRefinements =
     get(searchState, 'refinementList.instructor_name', []).length +
     get(searchState, 'refinementList._tags', []).length +
-    get(searchState, 'refinementList.type', []).length +
-    get(searchState, 'page', 0)
```
vojtaholik, line 76: this was causing incorrect refinements count, think it was added by mistake

src/components/search/index.tsx
```diff
@@ -209,6 +207,7 @@ const Search: FunctionComponent<SearchProps> = ({
 
           <div className="dark:bg-gray-900 bg-gray-50  md:-mt-5">
             <div className="mb-10 pb-10 xl:px-0 px-5 max-w-screen-xl mx-auto dark:bg-gray-900">
+              <ScrollTo scrollOn="page" />
               <Hits />
```
vojtaholik, line 211: moved this here so that paginating on curated page won't force scrolls below curated content

src/components/search/index.tsx
```diff
@@ -72,8 +72,7 @@ const Search: FunctionComponent<SearchProps> = ({
   const numberOfRefinements =
     get(searchState, 'refinementList.instructor_name', []).length +
     get(searchState, 'refinementList._tags', []).length +
-    get(searchState, 'refinementList.type', []).length +
-    get(searchState, 'page', 0)
+    get(searchState, 'refinementList.type', []).length
```
vojtaholik, line 75: this was causing incorrect refinements count, think it was added by mistake

vojtaholik:
> Not sure if there are other cases we should be checking

Should be fine, it's not being used anywhere else.

zacjones93 (approved):

PR #580, opened 2021-04-23
New Storybook card components and sections

MaggieAppleton:
This adds a number of new components to our storybook documentation.

## New Cards

1. `<CardHorizontal />` – originally defined in the home index page, now defined in a separate component in `components/pages/home/cards`
image

2. `<CardVerticalLarge />` – ditto above
image

3. `<CardVerticalWithStack />` - ditto above
image

4. `<PodcastCard />` - designed to match our podcast data objects
image

5. `<ArticleCard / >` - designed to match our article data objects. Takes a `horizontal` optional prop to return a small, horizontal version
image

The podcast and article cards are very rough design-wise. @laurosilvacom @zacjones93 and I threw them together on a gardening call. Can be used but need refinement.

## New Homepage Section Layouts

The second round of new storybook components are a set of layouts intended to go on the homepage. These are defined under `components/pages/home/features`

1. `<WhatsNew />`is designed to be the lead feature on the homepage. It includes the `<Jumbotron />` component inside it, and is clearly labelled with a "What's New" title.
image

At the moment the placeholder cards beneath the jumbotron haven't been designed. Chucked standard horizontal cards in there. Need to mockup better solutions in Figma then implement.

2. `<FeaturedSectionDetail />` is for sections such as out Digital Gardening for Devs feature. It's designed to live high up on the homepage. It accepts a title, feature image, description, and any number of resource cards below that point.
image

3. `<FeaturedSectionSimple />` is a simplified version that only has a title and any number of featured resource cards below it.
image

This PR shouldn't change anything that's user-facing. It only cleans up the structure of our card components and adds them to storybook.


vojtaholik:
It's probably too late now for this PR, but ideally we should be thinking about generic card component structure first and layer on top of that. This is starting to get chaotic, we aren't reusing atom components properly, and thus the more card components we add, the harder it will be to refactor later. Joel started [this](https://design.egghead.io/?path=/story/components-card--default-story) and I should have followed up, but didn't have time to do so yet. We were looking at examples from [semantic-ui](https://react.semantic-ui.com/views/card/) and [spectrum](https://spectrum.adobe.com/page/cards/) as an inspiration in terms of structure. there's also some really good stuff in [this twitter thread](https://twitter.com/jhooks/status/1352656727213273094?s=20).

joelhooks:
ya, this is a challenge because the foundational work that needs to be done is “very advanced” component development and system design and the more we toss variants onto the pile the more difficult it will be to bring it back in line

I appreciate you bringing this up Vojta.

MaggieAppleton:
Very true it would be much better to structure to cards for scalability. I don't personally have enough knowledge or understanding of component composition and architecture to help much. I'm just slightly stuck on being able to work on homepage improvements without a card architecture system in place, so started winging it.
Eg. we currently don't have any way to feature articles, podcasts or course clubs within cards.

Can we break down the challenge into some tangible steps? A Figma library of current card structures and requirements?

MaggieAppleton:
Started screenshotting all the current cards into a Figma file to give us space to figure it out. Link in Roam.

MaggieAppleton:
Also note @zacjones93 has an open PR building off this that's related #583

src/components/pages/home/card/article-card.tsx
```diff
@@ -0,0 +1,107 @@
...
+      className={
+        horizontal
+          ? 'flex sm:flex-row flex-col max-w-2xl rounded-lg shadow-lg overflow-hidden mb-4'
+          : 'max-w-md rounded-lg shadow-lg overflow-hidden mb-4'
+      }
+      key={path}
+    >
+      {image && path && (
+        <Link
+          className={
```
vojtaholik, line 41: className on Link was causing build to fail, should be on anchor below instead.

PR #582, opened 2021-04-23
Add Team Billing Summary back in

jbranchaud:
- Add Team Billing Summary back in. It must have been cut during a bad merge.
- This also adds an undefined check on the subscription data, so that
instead of erroring we just don't render the component.
- Handle Next Billing Date w/o upcoming invoice.
If it is too early in the billing period or if the subscription was just
created, there may not be an upcoming invoice generated yet. In that
case, we see the subscription is active and display 'Pending' for now.

### DIFF
--- a/src/components/team/billing-section.tsx
+++ b/src/components/team/billing-section.tsx
@@ -30,6 +30,8 @@ const BillingSection = ({
     slug,
   })
 
+  if (subscriptionData === undefined) return null
+
   const currency = get(
     subscriptionData,
     'latestInvoice.currency',
--- a/src/components/team/billing-section.tsx
+++ b/src/components/team/billing-section.tsx
@@ -4,6 +4,7 @@ import get from 'lodash/get'
 import {format} from 'date-fns'
 import Link from 'next/link'
 import {track} from '../../utils/analytics'
+import isEmpty from 'lodash/isEmpty'
 
 const formatAmountWithCurrency = (
   amountInCents: number,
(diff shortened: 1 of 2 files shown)


src/pages/team/index.tsx
```diff
@@ -163,6 +163,10 @@ const Team = ({team: teamData}: TeamPageProps) => {
             Current Team Members <TeamComposition teamData={teamData} />
           </h2>
           <MemberTable members={teamData.members} />
+          <BillingSection
+            stripeCustomerId={teamData.stripeCustomerId}
+            slug={teamData.accountSlug}
+          />
```
jbranchaud, line 169: These are the lines that got lost in a bad merge somewhere.

src/components/team/billing-section.tsx
```diff
@@ -72,14 +75,31 @@ const BillingSection = ({
...
+  } else if (activeSubscription && nextPaymentAttempt) {
+    nextBillDateDisplay = format(
+      new Date(
+        get(subscriptionData, 'upcomingInvoice.next_payment_attempt') * 1000,
+      ),
+      'yyyy/MM/dd',
+    )
+  } else {
+    nextBillDateDisplay = '-'
+  }
```
jbranchaud, line 102: I'm not super excited about adding this big if/else statement in, but 1) we don't always have the upcoming invoice and 2) we want to always display _something_ instead of erroring out on invalid dates or a missed edge case.

src/components/team/billing-section.tsx
```diff
@@ -72,14 +75,31 @@ const BillingSection = ({
...
-  const nextBillDate: string | undefined = activeSubscription
-    ? format(
-        new Date(
-          get(subscriptionData, 'upcomingInvoice.next_payment_attempt') * 1000,
-        ),
-        'yyyy/MM/dd',
-      )
-    : undefined
+
+  // Figure out what to display for the Next Billing Date.
```
joelhooks, line 79: The `current_period_*` props on the `Subscription` object are likely to give a good indication https://stripe.com/docs/api/subscriptions/object

joelhooks (approved):

PR #584, opened 2021-04-23
Pull data from sanity for What's New section

zacjones93:
This is an addition to @MaggieAppleton's PR #580 *PRing into Maggies branch freaked vercel previews out so I just made another PR into main

I had some trouble with styling and the new card components so I ended up just rolling with `CardHorizontal`

This is proof of concept, would love a discussion about how everyone expects data to be coming in for a feature like this. To me, it makes sense to have primary, side, and bottom, collections which we should be able to swap out with any resources moving forward. 

The way the data comes in has implications for how the `WhatsNew` component is structured. I'm not sure I like the array destructuring that I do in the component.

Having the `What's New` component as a page would make sense as we could make larger queries for courses, articles, podcasts, etc.

I'm currently rendering the same component on the home screen as well as the what's new route.

## What's New Page

## Home Page


src/components/pages/home/jumbotron/index.tsx
```diff
@@ -61,7 +67,9 @@ const Jumbotron: FunctionComponent<JumbotronProps> = ({resource}) => {
               </h2>
               <Link href={path}>
                 <a
-                  className="sm:text-2xl md:text-4xl text-xl max-w-screen-lg font-extrabold leading-tighter hover:text-yellow-500"
+                  className={`sm:text-2xl md:text-4xl text-xl max-w-screen-lg font-extrabold leading-tighter ${
+                    titleHover ? titleHover : 'hover:text-blue-600'
+                  }`}
```
zacjones93, line 72: Added a `titleHover` Prop because we change the hardcoded with value with every update to the Jumbotron. Since we are starting to use the Jumbotron elsewhere, it make sense to make this dynamic

src/components/pages/home/jumbotron/index.tsx
```diff
@@ -88,7 +96,13 @@ const Jumbotron: FunctionComponent<JumbotronProps> = ({resource}) => {
...
                     alt={instructor.name}
                   />
-                  <span className="group-hover:text-orange-200">
+                  <span
+                    className={`group-hover:text-orange-200 ${
+                      instructorHover
+                        ? instructorHover
+                        : 'group-hover:text-blue-300'
+                    }`}
+                  >
```
zacjones93, line 105: Same with the `instructorHover` text

src/pages/index.tsx
```diff
@@ -34,6 +35,7 @@ export default IndexPage
 const featureQuery = groq`
 {
   'featureDigitalGardening': ${digitalGardeningQuery},
+  'whatsNew': ${whatsNewQuery},
```
zacjones93, line 38: adds the query to the featureQuery that we send on the home page

src/pages/learn/whats-new/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+import Link from 'next/link'
+import React, {FunctionComponent} from 'react'
+import Jumbotron from 'components/pages/home/jumbotron'
+import CardHorizontal from 'components/pages/home/card/card-horizontal'
+
+const WhatsNewPage: FunctionComponent<any> = ({data}) => {
+  const {primary, side, bottom} = data
+  const [jumbotron, secondPrimary] = primary.resources
+  const [firstSide, secondSide] = side.resources
+  const [firstBottom, secondBottom] = bottom.resources
```
zacjones93, line 13: This feels weird but might be ok since these items should change frequently

src/pages/learn/whats-new/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+          <img
+            className="absolute top-0 left-0 z-0 w-full"
+            src={fancyCardBackground}
+            alt=""
+          />
+        </div>
+      </a>
+    </Link>
+  )
+}
```
zacjones93, line 87: This could be a candidate for a new card in /components

src/pages/learn/whats-new/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+	'bottom': resources[slug.current == 'bottom-resources'][0]{
+    resources[]{
+     'name': type,
+      title,
+      path,
+      byline,
+      image,
+    }
+  },
+}`
```
zacjones93, line 128: Query for resource

src/pages/learn/whats-new/index.tsx
```diff
@@ -0,0 +1,138 @@
+import groq from 'groq'
```
joelhooks, line 1: this particular page can be a root `/new` page

no need to put it in `/learn` since it's a "fixture"

I like "learn" when it reads like a sentence.

thing of `new` and a root itself that could have sub-pages

joelhooks (commented):
This looks great! I'll let somebody else approve, but great to see this direction.

MaggieAppleton:
I really like the direction this is headed. See new discusses re: card architecture on #580 though. Might need to wait until we've sorted out that issue until we can move forward with this

joelhooks:
there’s no sense waiting for anything, but we need to be aware of the spiraling complexity we are likely creating


there’s nobody to sort out the cards

PR #585, opened 2021-04-23
Replacing the Bitmovin player

joelhooks:
This replaces the Bitmovin player on video resource (lesson) pages.

I tested it fairly well, but could use some additional help!

image


theianjones:
I notices that the playback speed selection gets cleared if you:
- select 1.5x (or anything other than 1.0)

PR #590, opened 2021-04-26
Wire up 'Remove' team member button (with XState)

jbranchaud:
This PR adds the ability of an Account Owner to remove members from their team. This is great for a situation where one of the users leaves the company and the owner wants to free up that seat for a new hire.

- Add a 'Remove' button next to each member.
- Removing a member is destructive, so clicking it will open an XState-powered Confirm dialog.
- Confirming the dialog will fire off a request to the server.
- If the request succeeds, the list of users will be updated and a success toast appears.
- If the request fails, then the list is untouched and a failure toast appears.

https://user-images.githubusercontent.com/694063/116145430-f3323a00-a6a2-11eb-8400-c8f0a6afd96d.mp4


src/components/dialog.tsx
```diff
@@ -1,5 +1,6 @@
 import React, {FunctionComponent} from 'react'
 import {DialogOverlay, DialogContent} from '@reach/dialog'
+import IconX from 'components/icons/icon-x'
```
jbranchaud, line 3: Move this icon to a shared location so that multiple components can use it.

src/components/team/member-table.tsx
```diff
@@ -1,59 +1,98 @@
...
+  setMembers,
+}: {
+  accountId: number
+  members: any[]
+  setMembers: any
+}) => {
+  const initialContext = {accountId, setMembers}
+  const [current, send] = useMachine(confirmationDialogMachine, {
+    context: initialContext,
+  })
```
jbranchaud, line 19: Initialize the XState machine that powers the Confirm dialog. The initial context includes the `accountId` (needed for constructing the DELETE request URL) and `setMembers` (the setter for updating the list of members displayed on the page).

src/components/team/member-table.tsx
```diff
@@ -1,59 +1,98 @@
...
+        state={current.value}
+        isOpen={current.value !== 'closed'}
+        onClose={() => {
+          send('CANCEL')
+        }}
+        onConfirm={() => {
+          send('CONFIRM')
+        }}
+        member={current.context.memberToRemove}
+      />
```
jbranchaud, line 33: The confirmation dialog renders here adjacent to the list of members.

src/components/team/member-table.tsx
```diff
@@ -1,59 +1,98 @@
...
+                        <p
+                          onClick={() => {
+                            send({type: 'OPEN_DIALOG', payload: {member}})
+                            // removeTeamMember(id)
+                          }}
+                          className="cursor-pointer px-2 inline-flex text-xs leading-5 font-semibold rounded-full bg-red-100 dark:bg-red-700 text-red-800 dark:text-red-100 hover:shadow"
+                        >
+                          Remove
+                        </p>
+                      )}
```
jbranchaud, line 71: Remove button, triggers the Confirm Dialog to open. This 'remove' button does not show up for owners, only members can be removed.

src/components/team/remove-member-confirm-dialog.tsx
```diff
@@ -0,0 +1,88 @@
+import React, {FunctionComponent} from 'react'
```
jbranchaud, line 1: This is the 'Remove Team Member' confirm dialog. It shows the name/email of the member you are about to remove from your team and asks you to confirm you want to do it.

You can 'Cancel', 'x', or hit 'Escape'. Or you can hit 'Confirm' which will trigger a side-effect action in the machine.

src/machines/confirmation-dialog-machine.ts
```diff
@@ -0,0 +1,169 @@
+// Source: https://xstate-catalogue.com/machines/confirmation-dialog
```
jbranchaud, line 1: The machine.



This state machine is an expanded/tweaked version of the [Confirmation Dialog machine from XState Catalogue](https://xstate-catalogue.com/machines/confirmation-dialog).

The primary paths are:

1. `Closed -(Open Dialog)-> Open (Idle) -(Cancel)-> Closed`
2. `Closed -(Open Dialog)-> Open (Idle) -(Confirm)-> Open (Execute Action) -(onDone)-> Closed`

See [the graph at the top of this page](https://xstate-catalogue.com/machines/confirmation-dialog) for exhaustive details.

If the `Confirm` transition is triggered. The machine will make an API request to attempt to remove the member from the team.

If that request succeeds, it will go down the `onDone` path which closes the dialog, updates the list of members to be displayed, and triggers a success toast.

If that request fails, it will go down the `onError` path which closes the dialog, leaves the list of members as is, and triggers a failure toast.

src/pages/team/index.tsx
```diff
@@ -25,9 +25,13 @@ export type TeamData = {
...
-const TeamComposition = ({teamData}: {teamData: TeamData}) => {
-  const {numberOfMembers, capacity} = teamData
-
+const TeamComposition = ({
+  capacity,
+  numberOfMembers,
+}: {
+  capacity: number
+  numberOfMembers: number
+}) => {
```
jbranchaud, line 34: Passing in the two needed values instead of the whole `teamData` object.

src/pages/team/index.tsx
```diff
@@ -117,11 +121,12 @@ const AtCapacityNotice = ({teamData}: {teamData: TeamData}) => {
 }
 
 type TeamPageProps = {
-  team: TeamData | undefined
+  team: TeamData | null
 }
 
 const Team = ({team: teamData}: TeamPageProps) => {
   const router = useRouter()
+  const [members, setMembers] = React.useState<any[]>(teamData?.members || [])
```
jbranchaud, line 129: Using component state to manage the list of team members so that it can be updated with a `setMembers` call if someone is removed.

src/pages/team/index.tsx
```diff
@@ -131,44 +136,52 @@ const Team = ({team: teamData}: TeamPageProps) => {
     }
   }, [teamDataNotAvailable])
 
+  if (teamData === null) return null
+
   return (
     <LoginRequired>
-      {!!teamData && (
```
jbranchaud, line 136: Moving the presence check up here to reduce nesting and help with some TS type narrowing.

jbranchaud:
I'm getting a few type errors that don't seem right to me. Based on the docs and the runtime behavior of this machine, the first argument to an `action` is the `context` which is typed with `ConfirmationDialogMachineContext`, but ts-server is flagging this as incompatible, saying it should match type `ConfirmationDialogMachineEvent`. It's probably a misunderstanding I have about the XState's types. I'm not seeing how these type errors make sense.

CleanShot 2021-04-26 at 15 56 11@2x

Based on these type definitions, I wouldn't expect the type error that is appearing above.

CleanShot 2021-04-26 at 16 00 35@2x

jbranchaud:
Tagging @davidkpiano, he offered to take a look at the type errors that are cropping up.

mattpocock:
@jbranchaud There is some known weirdness with the `assign` action in v4. The cleanest way to get good typings with it is to always use this syntax:

`assign((context) => { return { something: true } })`

Adding the `context` piece, even if unused, will fix these errors.

On the error in the `executeAction` service - this looks funky. Are you sure the context/event generics passed to the machine are declared properly? It looks like it's moaning about an event type, but you're only referencing context there.

jbranchaud:
> Are you sure the context/event generics passed to the machine are declared properly? It looks like it's moaning about an event type, but you're only referencing context there.

@mattpocock I'm fairly confident they are. The generics/type setup for the machine is almost entirely untouched from what I pulled in from XState Catalogue. The digging I did into [XState's types](https://github.com/davidkpiano/xstate/blob/master/packages/core/src/Machine.ts)—which I'm definitely not an expert at reading through—suggests to me that this is declared correctly. First is the Context generic and second is the Event generic.

And great call on the type error on the `assign`s -- using the function argument instead of just passing an object cleared that part up.

Thanks for looking at it. Your work on XState/XState Catalogue is much appreciated.

jbranchaud:
> > Are you sure the context/event generics passed to the machine are declared properly? It looks like it's moaning about an event type, but you're only referencing context there.
> 
> @mattpocock I'm fairly confident they are. The generics/type setup for the machine is almost entirely untouched from what I pulled in from XState Catalogue. The digging I did into [XState's types](https://github.com/davidkpiano/xstate/blob/master/packages/core/src/Machine.ts)—which I'm definitely not an expert at reading through—suggests to me that this is declared correctly. First is the Context generic and second is the Event generic.
> 
> And great call on the type error on the `assign`s -- using the function argument instead of just passing an object cleared that part up.
> 
> Thanks for looking at it. Your work on XState/XState Catalogue is much appreciated.

😯 Whoa, so fixing the `assign` calls untangled something for typescript and it is now appropriately typing the stuff in `executeAction`. The compiler is happy now and things are typed in the way I was originally expecting.

mattpocock:
@jbranchaud Fabulous. You're welcome!

src/machines/confirmation-dialog-machine.ts
```diff
@@ -0,0 +1,182 @@
...
+    id: 'confirmationDialog',
+    initial: 'closed',
+    context: {
+      accountId: undefined,
+      memberToRemove: undefined,
+      setMembers: () => {},
+    },
+    states: {
+      closed: {
+        id: 'closed',
```
theianjones, line 62: is there a reason you us the state id here? since the name of the state is `closed` couldnt we omit the id and just use `closed` instead of `#closed` in the actions?

src/components/team/remove-member-confirm-dialog.tsx
```diff
@@ -0,0 +1,91 @@
...
+import {DialogOverlay, DialogContent} from '@reach/dialog'
+import IconX from 'components/icons/icon-x'
+import isObject from 'lodash/isObject'
+import {StateValue} from 'xstate'
+
+type DialogProps = {
+  state: StateValue
```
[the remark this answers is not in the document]
jbranchaud: I did try typing this as `state: string | { idle: string }`. It is not simply a string because of the nested states. However even this union type didn't work. XState types this as `StateValue` which I didn't see a way of narrowing to that union type. Declaring it as `StateValue` cleared up the type inference and then I do some explicit runtime checking in `remove-member-confirm-dialog.tsx`.

src/components/team/remove-member-confirm-dialog.tsx
```diff
@@ -0,0 +1,91 @@
...
+}
+
+const RemoveMemberConfirmDialog: FunctionComponent<DialogProps> = ({
+  state,
+  onClose,
+  member,
+  onConfirm,
+}) => {
+  const isOpen = state !== 'closed'
+  const loading = isObject(state) && state?.open === 'executingAction'
```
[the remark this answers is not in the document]
jbranchaud: https://github.com/eggheadio/egghead-next/pull/590#discussion_r621578974

src/machines/confirmation-dialog-machine.ts
```diff
@@ -0,0 +1,182 @@
...
+  errorMessage?: string
+  accountId: number | undefined
+  memberToRemove: Member | undefined
+  setMembers: Dispatch<SetStateAction<any[]>>
+}
+
+export type ConfirmationDialogMachineEvent =
+  | {
+      type: 'OPEN_DIALOG'
+      action: () => Promise<void>
```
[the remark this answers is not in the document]
jbranchaud: The `action` is something that came directly from the machine starter I used ([here](https://xstate-catalogue.com/machines/confirmation-dialog)). It is not being used, so I should remove it.

I would like to take a separate pass over this machine and see if I can move the `removeTeamMember` function out of the machine and instead pass it in as the `action`... which is probably the more idiomatic way of doing things. I could use this same technique to move the Success and Error Toast actions outside of the machine. This would make the machine less complex and keep it a little more generic.

CleanShot 2021-04-27 at 15 34 41@2x

With regard to your question of whether the generic should be `void`: I'm not sure what would make sense here. It might just be that they gave it void as boilerplate and it is up to the specific implementation to give it a type that makes sense.

src/machines/confirmation-dialog-machine.ts
jbranchaud: This is another case of, "this is what the template gave me." It seems likely that I could make the adjustment you are suggesting.

CleanShot 2021-04-27 at 15 39 28@2x

jbranchaud: Interesting... so when I remove the `id: 'closed'` and reference that state directly, I get the following error when trying to transition out of the `open.idle` state with `CANCEL`. So perhaps when transitioning between levels of nesting the `id` is needed? 🤔 

CleanShot 2021-04-27 at 17 36 55@2x

mattpocock: That's correct - the id is necessary to route you from the nested state to a sibling of its parent

joelhooks (approved):

PR #591, opened 2021-04-27
users can have multiple accounts breaking team page

joelhooks:
multiple accounts is likely a long-term problem and we need to resolve it.

IMO it’s a mistake to have “accounts” and “team accounts” as separate data. 

Accounts are accounts are accounts.

Because of the multiple “teams” (all accounts are teams) using the zero-index to get the “first” is resulting in an empty non-team account

### DIFF
--- a/src/pages/team/index.tsx
+++ b/src/pages/team/index.tsx
@@ -9,7 +9,7 @@ import {track} from 'utils/analytics'
 import {loadTeams} from 'lib/teams'
 import TeamName from '../../components/team/team-name'
 import {getTokenFromCookieHeaders} from 'utils/auth'
-import isEmpty from 'lodash/isEmpty'
+import {isEmpty, find} from 'lodash'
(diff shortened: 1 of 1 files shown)


src/pages/team/index.tsx
```diff
@@ -180,13 +181,12 @@ export const getServerSideProps: GetServerSideProps<TeamPageProps> = async funct
     context.req.headers.cookie as string,
   )
 
-  const {data: teams} = await loadTeams(eggheadToken)
+  const {data: teams = []} = await loadTeams(eggheadToken)
 
-  let team: TeamData | undefined = undefined
+  const fetchedTeam = find(teams, (team) => team.capacity > 0)
```
joelhooks, line 186: instead of assuming `0` (first) item is team account we will find an account with a capacity over `0` if it exists.

Also removed the `undefined` typing by adding an array as default.

jbranchaud (approved):

PR #595, opened 2021-04-27
consolidates portfolio clubs into single cta

joelhooks:
image

### DIFF
--- a/src/components/app/header.tsx
+++ b/src/components/app/header.tsx
@@ -11,7 +11,7 @@ import {useRouter} from 'next/router'
 import {useTheme} from 'next-themes'
 import useCio from '../../hooks/use-cio'
 import {Form, Formik} from 'formik'
-import OnlinePresenceCTA from '../survey/online-presence-cta'
+import ProjectClubCTA from '../survey/project-club'
 
 const Header: FunctionComponent = () => {
   const router = useRouter()
@@ -36,7 +36,7 @@ const Header: FunctionComponent = () => {
         {viewer ? (
           <div className={className}>
             {children}
-            <OnlinePresenceCTA variant="header" />
+            {viewer.is_pro && <ProjectClubCTA variant="header" />}
             <Feedback
               user={viewer}
               className="px-3 py-2 hover:bg-gray-100 dark:hover:bg-gray-800 dark:hover:text-white active:bg-gray-200 rounded-md inline-flex transition-all ease-in-out duration-300 leading-tight"
@@ -96,13 +96,6 @@ const Header: FunctionComponent = () => {
         ) : (
           <div className={className}>
             {children}
-            <div className="hidden lg:block">
-              <Link href="/own-your-online-presence">
-                <a className="inline-flex justify-center items-center px-4 py-2 rounded-md bg-blue-600 text-white transition-all hover:bg-blue-700 ease-in-out duration-200">
-                  Own Your Online Presence
-                </a>
-              </Link>
-            </div>
             <Link href="/login" activeClassName="bg-gray-100 dark:bg-gray-400">
               <a
                 onClick={() =>
--- a/src/components/customer-io/email-entry-form.tsx
+++ b/src/components/customer-io/email-entry-form.tsx
@@ -1,0 +1,60 @@
+import * as React from 'react'
+import {track} from 'utils/analytics'
+import EmailForm from '../cta/email/email-form'
+import useCio from 'hooks/use-cio'
+import {useViewer} from 'context/viewer-context'
+import {requestSignInEmail} from 'utils/request-signin-email'
+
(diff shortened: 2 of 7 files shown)


src/components/app/header.tsx
```diff
@@ -36,7 +36,7 @@ const Header: FunctionComponent = () => {
         {viewer ? (
           <div className={className}>
             {children}
-            <OnlinePresenceCTA variant="header" />
+            {viewer.is_pro && <ProjectClubCTA variant="header" />}
```
joelhooks, line 39: I want to use the header for this CTA, but want to make sure it's only members.

@MaggieAppleton I think this CTA is an interesting spot for some light personalization. We could have logic here that decided what we wanted to ask the user to do right now.

Lots of potential.

src/components/app/header.tsx
```diff
@@ -96,13 +96,6 @@ const Header: FunctionComponent = () => {
         ) : (
           <div className={className}>
             {children}
-            <div className="hidden lg:block">
-              <Link href="/own-your-online-presence">
-                <a className="inline-flex justify-center items-center px-4 py-2 rounded-md bg-blue-600 text-white transition-all hover:bg-blue-700 ease-in-out duration-200">
-                  Own Your Online Presence
```
joelhooks, line 102: not showing this to logged-out users intentionally.

src/components/customer-io/email-entry-form.tsx
```diff
@@ -0,0 +1,60 @@
+import * as React from 'react'
+import {track} from 'utils/analytics'
+import EmailForm from '../cta/email/email-form'
+import useCio from 'hooks/use-cio'
+import {useViewer} from 'context/viewer-context'
+import {requestSignInEmail} from 'utils/request-signin-email'
+
+const EmailEntryForm: React.FC<any> = ({
```
joelhooks, line 8: made this generic. Still could use some work to make it nice.

src/components/customer-io/email-entry-form.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const {subscriber, cioIdentify} = useCio()
+  const {viewer} = useViewer()
+
+  const onSubmit = async (values: any) => {
+    const {email} = values
+
+    let id = subscriber?.id || viewer?.contact_id
+
+    if (!id) {
+      const {contact_id} = await requestSignInEmail(email)
```
joelhooks, line 24: Customer.io isn't email-based, we use a "contact id" so if we don't know the user we need to make one based on the email.

For this specific PR it doesn't matter since we know the customer if they got here through our CTA, but if they stumble on it it will make them an "account"

src/components/survey/project-club.tsx
```diff
@@ -0,0 +1,110 @@
...
+    heading: `Join a React Project Club`,
+    subheading: `Let's build a business oriented protfolio project together`,
+    type: `cta-link`,
+    url: `/clubs/portfolio-project`,
+    button_label: `Join a Project Club!`,
+    final: true,
+  },
+}
+
+const ProjectClubCTA: React.FunctionComponent<{
```
joelhooks, line 37: I used the survey mechanism for this, which is silly, but functional

src/pages/clubs/redux/index.tsx
```diff
@@ -59,12 +59,23 @@ const reduxClub: React.FC<any> = ({data}) => {
 
 export default reduxClub
 
+export const reduxClubQuery = groq`*[_type == 'resource' && slug.current == "build-business-oriented-portfolio"][0]{
```
joelhooks, line 62: this was being imported but I didn't understand why since it was only used in one spot that I could tell.

src/pages/clubs/redux/index.tsx
```diff
@@ -59,12 +59,23 @@ const reduxClub: React.FC<any> = ({data}) => {
...
+      description,
+      "slug": slug.current,
+      image,
+      summary,
+    }
+}`
+
 export async function getStaticProps() {
-  const data = await sanityClient.fetch(reduxClubQuery)
+  const {redux} = await sanityClient.fetch(reduxClubQuery)
```
joelhooks, line 74: just give the component what it needs instead of generic `data`

src/components/app/header.tsx
laurosilvacom: Nice. Making it only members makes sense.

src/pages/clubs/redux/index.tsx
zacjones93: A good point hadn't thought about this. Or just assumed that we would be querying only the data that was needed

src/components/app/header.tsx
zacjones93: love how easy it is to check for viewer status

PR #596, opened 2021-04-28
Add spacing and titles to homepage

MaggieAppleton:
This divides the cards on the homepage into sections and adds titles to each group of resources.

---

Some of the section layouts are still a bit awkward (portfolio clubs and gardening), and we still need to do some gardening work making the section titles match the content, but it feels presentable for now.


vojtaholik:
nit: from the screenshot it feels like headings are quite inconsistent and have different letter spacing, sizes, and alignment. I would also say that space between sections feel bit too large, especially on smaller screens (13" and less) — mostly apparent when there is no heading present to visually divide given sections.

MaggieAppleton:
The headings are indeed different sizings and alignments on purpose 😉
Generally when we have standard sections like "What's New," "Staff picks," and "Most popular" we should keep sizing consistent.
For the two special features (currently build a dev portfolio and digital gardening) I think we should have room to play with different layouts and typography. These kind of feature sections should stand out from the other categories.
Perhaps we can do that while keeping typography consistent if we find another way to differentiate them?

I used mt-32 for all the section breaks, but I can't figure out a way for tailwinds to let to me vh for margin values.
I swapped it out for `mt-12 border-t pt-12 sm:mt-16 sm:pt-16` which gives us a nice subtle border dividing sections up instead of just open space, and makes the gaps smaller on mobile.

image

MaggieAppleton:
Just updated the two feature sections to be less awkward and stand out more

image

image

laurosilvacom (approved):

PR #597, opened 2021-04-28
Removed level up CTA from lesson page

MaggieAppleton:
Our lesson pages had a leftover CTA box to schedule 1:1 career planning chats. Since we're not running that variety of value path interview anymore, we should remove the box.

image


zacjones93 (approved):
This makes sense

PR #598, opened 2021-04-28
Add light mode to homepage feature cards

zacjones93:
We support light and dark modes for the site so our custom CTA's should support that as well

This attempts to solve that issue by loading two different themed backgrounds based on the current theme set in the app.

One thing I'm not handling here is the case where an asset is missing. So if we didn't have a light background set, none would show for that.

I am thinking about turning this functionality into a hook for reuse elsewhere. I can see this being useful for our curated and instructor page CTA cards.

Light:
light mode

Dark:
dark mode

### DIFF
--- a/src/components/pages/home/card/index.tsx
+++ b/src/components/pages/home/card/index.tsx
@@ -17,6 +17,8 @@ export type CardResource = {
   instructor?: any
   background?: string
   lightBackground?: string
+  featureCardBackground?: string
+  lightFeatureCardBackground?: string
 }
 
 type CardProps = {
--- a/src/components/pages/home/card/index.tsx
+++ b/src/components/pages/home/card/index.tsx
@@ -16,6 +16,7 @@ export type CardResource = {
   resources?: CardResource[]
   instructor?: any
   background?: string
+  lightBackground?: string
 }
 
 type CardProps = {
--- a/src/components/pages/home/jumbotron/index.tsx
+++ b/src/components/pages/home/jumbotron/index.tsx
@@ -3,7 +3,6 @@ import {CardResource} from 'components/pages/home/card'
 import Markdown from 'react-markdown'
 import Link from 'next/link'
 import Image from 'next/image'
-import {useTheme} from 'next-themes'
 import {get} from 'lodash'
 import {bpMinMD} from 'utils/breakpoints'
 import {track} from 'utils/analytics'
@@ -11,26 +10,15 @@ import {track} from 'utils/analytics'
 type JumbotronProps = {
   resource: CardResource
   textColor?: String
+  background?: String
 }
 
 const Jumbotron: FunctionComponent<JumbotronProps> = ({
   resource,
   textColor,
+  background,
 }) => {
-  const {theme} = useTheme()
-  const {
-    path,
-    image,
-    title,
-    byline,
-    instructor,
-    background,
-    lightBackground,
-    description,
-  } = resource
-
-  console.log(theme)
-  const themedBackground = theme === 'dark' ? background : lightBackground
+  const {path, image, title, byline, instructor, description} = resource
 
   return (
     <div
@@ -74,7 +62,7 @@ const Jumbotron: FunctionComponent<JumbotronProps> = ({
               </h2>
               <Link href={path}>
                 <a
-                  className={`sm:text-2xl md:text-4xl text-xl max-w-screen-lg font-extrabold leading-tighter text-gray-900 dark:text-white hover:text-yellow-500`}
+                  className={`sm:text-2xl md:text-4xl text-xl max-w-screen-lg font-extrabold leading-tighter text-gray-900 dark:text-white `}
                   onClick={() =>
                     track('clicked jumbotron resource', {
                       resource: path,
@@ -119,7 +107,7 @@ const Jumbotron: FunctionComponent<JumbotronProps> = ({
       </div>
       <UniqueBackground
         className="absolute left-0 top-0 w-full h-full z-0 object-cover"
-        background={themedBackground}
+        background={background}
       />
     </div>
   )
--- a/src/components/pages/home/jumbotron/index.tsx
+++ b/src/components/pages/home/jumbotron/index.tsx
@@ -9,13 +9,11 @@ import {track} from 'utils/analytics'
 
 type JumbotronProps = {
   resource: CardResource
-  textColor?: String
   background?: String
 }
 
 const Jumbotron: FunctionComponent<JumbotronProps> = ({
   resource,
-  textColor,
   background,
 }) => {
   const {path, image, title, byline, instructor, description} = resource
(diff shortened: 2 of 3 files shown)


src/components/pages/home/card/index.tsx
```diff
@@ -16,6 +16,9 @@ export type CardResource = {
   resources?: CardResource[]
   instructor?: any
   background?: string
+  lightBackground?: string
+  featureCardBackground?: string
+  lightFeatureCardBackground?: string
```
zacjones93, line 21: I'm not sure how to extend a card type but in this situation ([Looked at this](https://stackoverflow.com/questions/41385059/possible-to-extend-types-in-typescript) but didn't get far

zacjones93: So this adds the possible backgrounds to the exported card type that we use

src/components/pages/home/jumbotron/index.tsx
```diff
@@ -67,7 +60,7 @@ const Jumbotron: FunctionComponent<JumbotronProps> = ({
               </h2>
               <Link href={path}>
                 <a
-                  className={`sm:text-2xl md:text-4xl text-xl max-w-screen-lg font-extrabold leading-tighter hover:text-yellow-500`}
+                  className={`sm:text-2xl md:text-4xl text-xl max-w-screen-lg font-extrabold leading-tighter text-gray-900 dark:text-white hover:text-blue-600 dark:hover:text-blue-300`}
```
zacjones93, line 63: set the hover of the jumbotron to our default action color

src/components/pages/home/jumbotron/index.tsx
```diff
@@ -78,32 +71,23 @@ const Jumbotron: FunctionComponent<JumbotronProps> = ({
...
               </Link>
-              <Link href={instructor.path}>
-                <a
-                  className="mt-4 flex items-center space-x-2 text-base group"
-                  onClick={() =>
-                    track('clicked instructor in jumbotron', {
-                      instructor: instructor.slug,
-                    })
-                  }
-                >
```
zacjones93, line 89: we don't link to instructors anywhere else on the page so removed that from the jumbotron

src/pages/new/index.tsx
```diff
@@ -8,8 +8,15 @@ import Markdown from 'react-markdown'
...
 import Card, {CardResource} from 'components/pages/home/card'
 import Jumbotron from 'components/pages/home/jumbotron'
+import {useTheme} from 'next-themes'
+
+const WhatsNewPage: FunctionComponent<WhatsNewResource> = ({resource}) => {
+  const [mounted, setMounted] = React.useState(false)
+  const {theme} = useTheme()
+
+  React.useEffect(() => setMounted(true), [])
+  if (!mounted) return null
```
zacjones93, line 18: Was getting a hydration issue where the opposite background was being set on the first load. This ensures that the theme is properly set before rendering card backgrounds

src/pages/new/index.tsx
```diff
@@ -18,17 +25,25 @@ const WhatsNewPage: FunctionComponent<any> = ({resource}) => {
     thirdSecondaryResource,
   ] = secondary.resources
 
+  const {background, lightBackground} = jumbotron
+  const {lightFeatureCardBackground, featureCardBackground} = secondPrimary
+
+  const themedBackground = theme === 'light' ? lightBackground : background
+  const themedFeatureCardBackground =
+    theme === 'light' ? lightFeatureCardBackground : featureCardBackground
```
zacjones93, line 33: check to see what theme is being used and render the proper background.

I haven't handled the case where data is not present. This was starting to get verbose so I thought I'd PR this and get the teams thoughts. This would be why I'd want to extract into a hook

src/pages/new/index.tsx
```diff
@@ -18,17 +25,25 @@ const WhatsNewPage: FunctionComponent<any> = ({resource}) => {
...
+
   return (
     <section className="sm:-my-5 -my-3 mx-auto max-w-screen-xl">
       <h2 className="md:text-xl text-lg sm:font-semibold font-bold mb-3 dark:text-white">
         What's New
       </h2>
-      <Jumbotron resource={jumbotron} textColor="text-green-400" />
+      <Jumbotron resource={jumbotron} background={themedBackground} />
```
zacjones93, line 40: jumbotron doesn't care what background it gets, it just renders!

src/pages/new/index.tsx
```diff
@@ -18,17 +25,25 @@ const WhatsNewPage: FunctionComponent<any> = ({resource}) => {
...
       <div className="grid grid-cols-1 lg:grid-cols-2 gap-4 mt-4">
         <div className="h-full grid gap-4">
           <CourseFeatureCard
             className="h-auto row-span-2 w-full"
             resource={secondPrimary}
+            featureCardBackground={themedFeatureCardBackground!}
```
zacjones93, line 46: same goes for the CourseFeatureCard

src/pages/new/index.tsx
```diff
@@ -120,15 +135,25 @@ export const CardHorizontal: FunctionComponent<{
   )
 }
 
-const CourseFeatureCard = ({resource, className}: any) => {
+type CourseFeatureCardProps = {
+  resource: CardResource
+  featureCardBackground: string
+  className?: string
+}
```
zacjones93, line 142: CeatureCard type!

src/pages/new/index.tsx
```diff
@@ -139,8 +164,15 @@ const CourseFeatureCard = ({resource, className}: any) => {
         <div className="flex flex-col items-center h-full">
           <div className="relative z-10 flex flex-col h-full justify-between  items-center sm:p-8 p-5">
             <div className="flex flex-col items-center">
-              <Image src={image} width={200} height={200} alt={title} />
-              <h2 className="text-xl font-bold min-w-full mt-4 sm:mt-14 mb-2 leading-tighter group-hover:underline">
+              {image && (
+                <Image
+                  src={get(image, 'src', image)}
```
zacjones93, line 169: I'm not quite sure why I have to use `get` this way but it's what typescript wanted.. This is how we call get the image URL in the base card resource

src/pages/new/index.tsx
```diff
@@ -158,6 +190,14 @@ const CourseFeatureCard = ({resource, className}: any) => {
   )
 }
 
+type WhatsNewResource = {
+  resource: {
+    title: string
+    primary: {resources: CardResource[]}
+    secondary: {resources: CardResource[]}
+  }
+}
```
zacjones93, line 199: WhatsNewResource type!

src/pages/new/index.tsx
```diff
@@ -168,7 +208,9 @@ export const whatsNewQuery = groq`*[_type == 'resource' && slug.current == "what
     	path,
     	image,
       'background': images[label == 'banner-image-blank'][0].url,
+      'lightBackground': images[label == 'light-banner-image-blank'][0].url,
       'featureCardBackground': images[label == 'feature-card-background'][0].url,
+      'lightFeatureCardBackground': images[label == 'light-feature-card-background'][0].url,
```
zacjones93, line 213: More data

MaggieAppleton:
I don't think these are working at the moment. Making everything light in light mode, and everything dark in dark mode means our banners aren't standing out. Seems to overcomplicate the problem?

The banners we've had in the past worked great on both light mode and dark mode. It's definitely possible to keep designing banners that work in both cases, and we shouldn't need to design separate backgrounds for each.

zacjones93:
that's a good point. designing cards that 'pop' and look good on both dark and light mode is a much simpler approach than overengineering the display of CTAs

PR #602, opened 2021-04-28
Pull sanity data for curated instructor pages

zacjones93:
All of our curated instructor data is currently hard coded on the front end.

This loads instructor data from sanity and passes it into curated instructor pages as `sanityInstructor`

One note, the linter was going ham on all the files so I committed with `no-verify`. Not sure what would have changed there but would like someone to run this on their machine.

Learn web development from Stephanie Eckles on egghead  egghead io
Learn web development from Stephanie Eckles on egghead  egghead io


src/components/search/index.tsx
```diff
@@ -191,7 +193,10 @@ const Search: FunctionComponent<SearchProps> = ({
           {!isEmpty(instructor) && (
             <div className="mb-10 pb-8 xl:px-0 px-5 mx-auto dark:bg-gray-900">
               {shouldDisplayLandingPageForInstructor(instructor.slug) && (
-                <InstructorCuratedPage instructor={instructor} />
+                <InstructorCuratedPage
+                  instructor={instructor}
+                  sanityInstructor={sanityInstructor}
+                />
```
zacjones93, line 199: pass sanity data down

src/components/search/instructors/index.tsx
```diff
@@ -14,6 +15,7 @@ const InstructorsIndex: any = {
   'flavio-corpa': SearchFlavioCorpa,
   'hiro-nishimura': SearchHirokoNishimura,
   'chris-biscardi': SearchChrisBiscardi,
+  'stephanie-eckles': SearchStephanieEckles,
```
zacjones93, line 18: add stephanie to the instructors index for curated pages

src/components/search/instructors/stephanie-eckles/index.tsx
```diff
@@ -0,0 +1,294 @@
...
+      />
+      <section className="grid grid-cols-4 gap-3 -mt-10 mb-10 pb-10 xl:px-0 px-5 max-w-screen-xl mx-auto dark:bg-gray-900 w-full">
+        <ProjectStack className="col-span-1" data={projects.resources} />
+        <div className="col-span-3 grid grid-cols-2 gap-3">
+          <CardHorizontal className="col-span-2" resource={secondCourse} />
+          <CardHorizontal className="col-span-1" resource={thirdCourse} />
+          <CardHorizontal className="col-span-1" resource={fourthCourse} />
+        </div>
+      </section>
+    </div>
```
zacjones93, line 51: display courses and projects on page

src/components/search/instructors/stephanie-eckles/index.tsx
```diff
@@ -0,0 +1,294 @@
...
+    	path,
+      byline,
+    	image,
+      'background': images[label == 'feature-card-background'][0].url,
+      'instructor': collaborators[]->[role == 'instructor'][0]{
+      	'name': person->.name
+    	},
+    }
+  },
+}`
```
zacjones93, line 77: sanity query

src/lib/instructors.ts
```diff
@@ -36,3 +39,26 @@ export async function loadInstructor(slug: string) {
...
+}
+
+export const loadSanityInstructor = async (selectedInstructor: string) => {
+  if (!canLoadSanityInstructor(selectedInstructor)) return
+
+  const query = sanityInstructorHash[selectedInstructor]
+  if (!query) return
+
+  return await sanityClient.fetch(query)
+}
```
zacjones93, line 64: @theianjones helped me implement this. 

`loadSanityInstructor` uses a type guard (`canLoadSanityInstructor`) to check that the instructor slug (the one that will be passed in via `getServerSideProps`) is present in the defined `sanityInstructorHash` as a key. If it is, that key is associated with the query that it will call. 

This makes sure that we are only querying for instructor data on the pages that we want to load sanity data onto.

src/pages/q/[[...all]].tsx
```diff
@@ -211,6 +215,8 @@ export const getServerSideProps: GetServerSideProps = async function ({
     )
     try {
       initialInstructor = await loadInstructor(instructorSlug)
+
+      sanityInstructor = await loadSanityInstructor(instructorSlug)
     } catch (error) {
```
Comment (zacjones93) [inline src/pages/q/[[...all]].tsx:220]: When an instructor filter is selected we will load sanity data (if it passes the type guard)

src/components/search/index.tsx
```diff
@@ -34,6 +34,7 @@ type SearchProps = {
   searchClient?: any
   searchState?: any
   instructor?: any
```
joelhooks, line 36: this stands out to me. I do t like “sanity” as special data. could we merge this into a si file “instructor” object so it’s “just an instructor” and then display data based on the shape of instructor and the available properties

src/components/search/instructors/stephanie-eckles/index.tsx
```diff
@@ -0,0 +1,290 @@
...
+import ExternalTrackedLink from 'components/external-tracked-link'
+
+export default function SearchStephanieEckles({
+  instructor,
+  sanityInstructor,
+}: {
+  instructor: any
+  sanityInstructor: any
+}) {
+  const combinedInstructor = {...instructor}
```
joelhooks, line 22: this could actually be combined at a higher level

src/pages/q/[[...all]].tsx
```diff
@@ -211,6 +215,8 @@ export const getServerSideProps: GetServerSideProps = async function ({
     )
     try {
       initialInstructor = await loadInstructor(instructorSlug)
+
```
Comment (joelhooks) [inline src/pages/q/[[...all]].tsx:218]: here’s where I’d consider a single “load instructor” that loads from all sources

src/pages/q/[[...all]].tsx
```diff
@@ -211,6 +212,13 @@ export const getServerSideProps: GetServerSideProps = async function ({
...
+      sanityInstructor = await loadSanityInstructor(instructorSlug)
+
+      initialInstructor = {
+        ...initialInstructor,
+        ...sanityInstructor,
+      }
```
Comment (zacjones93) [inline src/pages/q/[[...all]].tsx:221]: None of the data is being overridden but it makes sense to me that sanity data would take precedence over the `initialInstructor` data

Comment (joelhooks) [reply src/pages/q/[[...all]].tsx:221]: I think this pattern makes a lot of sense so we don't have two+ data sources at the component, we just bring it together at the server level.

Comment (joelhooks) [reply src/pages/q/[[...all]].tsx:221]: since we are using a `stale-while-revalidate` cache header, it should be performant!

Comment (zacjones93) [reply src/pages/q/[[...all]].tsx:221]: I agree, it was also super easy to refactor

src/lib/instructors.ts
```diff
@@ -36,3 +39,26 @@ export async function loadInstructor(slug: string) {
 
   return instructor
 }
+
+const sanityInstructorHash = {
+  'stephanie-eckles': stephanieEcklesQuery,
```
laurosilvacom, line 44: I like this pattern!

laurosilvacom (approved):

PR #603, opened 2021-04-29
Fix path and slug on homepage for concurrent react course.

### LINKED ISSUE #599: 'Concurrent React from Scratch' talk card on homepage linking to wrong page
https://github.com/eggheadio/egghead-next/blob/f691468877e4af81fec54d090288c835a4a461ee/src/components/pages/home/homepage-data.ts#L43-L60

51 is the offending line - it appears to be pointing to the CSS grid course.

Free annual subscription pls x

(thanks for the ace content!)

jimpala:
Fixes #599.

kramer nodding

xxx


laurosilvacom:
Fixed in https://github.com/eggheadio/egghead-next/pull/596

PR #605, opened 2021-04-30
add typescript course CTA on curated typescript page

zacjones93:
Adds latest typescript course to TypeScript curated page.

Tweaked some margin/spacing on cards for consistency

In-Depth TypeScript Tutorials for 2021  egghead io
In-Depth TypeScript Tutorials for 2021  egghead io
In-Depth TypeScript Tutorials for 2021  egghead io
In-Depth TypeScript Tutorials for 2021  egghead io

### DIFF
--- a/src/components/search/curated/typescript/index.tsx
+++ b/src/components/search/curated/typescript/index.tsx
@@ -35,7 +35,7 @@ Love them or hate them, static types are here to stay, and at the very least an 
       pageData={typescriptPageData}
       CTAComponent={CourseFeatureCard}
     >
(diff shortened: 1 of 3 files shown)


src/components/search/curated/typescript/index.tsx
```diff
@@ -69,4 +73,50 @@ Love them or hate them, static types are here to stay, and at the very least an
   )
 }
 
+const CourseFeatureCard = ({resource, className}: any) => {
```
joelhooks, line 76: suggestion: I think there's a strong case for just adding these components to a folder for the courses so we have a library of "feature" cards for these resources that we can use across the site. 🤔

joelhooks (approved):

PR #607, opened 2021-05-03
this removes some duplication in the "header button as survey" scheme

joelhooks:
I had cut and pasted the "header button that's actually a survey" so this removes that


src/components/app/header.tsx
```diff
@@ -131,14 +131,16 @@ const Header: FunctionComponent = () => {
...
           {!sm && !isTopics && (
-            <Link href="/topics">
-              <a
-                onClick={() => track(`clicked browse`, {location: 'header'})}
-                className="inline-flex justify-center items-center px-4 py-2 rounded-md bg-blue-600 text-white transition-all hover:bg-blue-700 ease-in-out duration-200"
-              >
-                Browse
-              </a>
-            </Link>
+            <div className={`${isSearch && 'w-full1'}`}>
```
joelhooks, line 134: this button was also out of alignment on the search page since the search bar isn't there.

src/pages/topics.tsx
```diff
@@ -18,7 +19,15 @@ const Tags: FunctionComponent<TagsProps> = ({tags}) => {
...
-              <a className="space-x-2 flex flex-row items-center justify-start rounded-lg hover:shadow-sm border border-transparent hover:border-gray-200 sm:p-5 p-4 w-full transition-all ease-in-out duration-150">
+              <a
+                onClick={() =>
+                  track(`clicked topic`, {
+                    location: 'topic page',
+                    topic: tag.slug,
+                  })
+                }
+                className="space-x-2 flex flex-row items-center justify-start rounded-lg hover:shadow-sm border border-transparent hover:border-gray-200 sm:p-5 p-4 w-full transition-all ease-in-out duration-150"
+              >
```
laurosilvacom, line 30: Makes sense to add the tracking.

laurosilvacom (approved):

PR #609, opened 2021-05-04
adds free and pro filter to search

joelhooks:
This required some work on the search index to expose this property, but now it can be used like any other search facet to filter search. Also persists in the browser query.

Didn't spend a lot of time with layout (obviously lol) 

image


zacjones93:
This looks good to me, @laurosilvacom and I noticed a behavior where podcasts and talks (along with their content types) get filtered out of search results when pro/free is selected.

This makes sense since they don't have an `access_state` associated with them (as far as I know). Might make sense to default them to free?

joelhooks:
@zacjones93 if you'd like to enhance in a future PR that's great, I'm personally just working on the pro content and not as concerned about the rest right now.

Generally, I think "yes and..." ideas should be a comment in the "approve this pr" comment box ;)

zacjones93 (approved):
Right, that makes sense. Will keep that as a note

PR #611, opened 2021-05-04
article course card with prose override

joelhooks:
This is using a `remark` plugin that opens and closes a `prose` enabled div around inline JSX elements.

As written we'd need to agree on a global MDX style, or just use this for `/blog` which is fine too. The overall prose div is defined in the plugin.

Just doing `/blog` is probably sensible, for my example, I'm using a template and `withMdx` core page because it was easier to demonstrate.

Works though!

image


joelhooks:
tagged `RFC` because there's still quite a bit of work to do.

next.config.js
```diff
@@ -4,9 +4,11 @@ const withPlugins = require(`next-compose-plugins`)
 const rehypeShiki = require(`rehype-shiki`)
 const checkEnv = require(`@47ng/check-env`).default
 const withImages = require(`next-images`)
+const {withProse} = require('./src/utils/remark/with-prose')
 const withMDX = require(`@next/mdx`)({
   extension: /\.mdx?$/,
   options: {
+    remarkPlugins: [withProse],
```
joelhooks, line 11: if we skipped it here we could use this for just the blog articles which makes sense to me.

src/utils/remark/with-prose.js
```diff
@@ -0,0 +1,53 @@
...
+        insideProse = false
+        return [{type: 'jsx', value: '</div>'}, node]
+      }
+      if (!insideProse && !isJsNode(node)) {
+        insideProse = true
+        return [
+          {
+            type: 'jsx',
+            value:
+              '<div className="prose dark:prose-dark sm:prose-lg lg:prose-xl mt-5 max-w-none">',
```
joelhooks, line 39: this is for the whole page and if you define `prose` on an element above the MDX it won't work.

src/utils/remark/with-prose.js
```diff
@@ -0,0 +1,53 @@
...
+
+module.exports.withProse = () => {
+  return (tree) => {
+    console.log('WITH PROSE')
+    let insideProse = false
+    tree.children = tree.children.flatMap((node, i) => {
+      console.log('PROSE', insideProse, isJsNode(node), node)
+      if (insideProse && isJsNode(node)) {
+        insideProse = false
+        return [{type: 'jsx', value: '</div>'}, node]
```
vojtaholik, line 31: I had to replace some `section` elements we were using in articles with `div` for the build to pass since it's only expecting divs. That seems like something that could be improved, otherwise we'll just have to remember it and use divs.

vojtaholik:
definitely the best solution I've seen so far! ⭐

joelhooks:
I isolated this to JUST /blog articles and not all MDX - we can expand later but this should be "safe" for now

PR #612, opened 2021-05-07
tries to find an account with a subscription

joelhooks:
we were using `[0]` and grabbing the first account which resulted in errors for users with multiple accounts (for... reasons). Now we grab the account with a subscription first, and then default to the `first` if one isn't found, and finally a "mock" account if nothing is found.

If a user had multiple accounts with multiple subscriptions, this won't work.

Something to consider.

### DIFF
--- a/src/pages/user/index.tsx
+++ b/src/pages/user/index.tsx
@@ -3,8 +3,7 @@ import {loadAccount} from 'lib/accounts'
 import LoginRequired, {LoginRequiredParams} from 'components/login-required'
 import {useViewer} from 'context/viewer-context'
 import RequestEmailChangeForm from 'components/users/request-email-change-form'
-import get from 'lodash/get'
-import isEmpty from 'lodash/isEmpty'
+import {get, isEmpty, find, first} from 'lodash'
 import SubscriptionDetails from 'components/users/subscription-details'
 import {loadUserProgress} from 'lib/users'
 import InProgressResource from 'components/pages/users/dashboard/activity/in-progress-resource'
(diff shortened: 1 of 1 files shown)


src/pages/user/index.tsx
```diff
@@ -25,33 +24,45 @@ const GithubConnectButton: React.FunctionComponent<{
...
+      if (viewerId) {
+        const {data} = await loadUserProgress(viewerId)
         setProgress(data)
       }
     }
 
-    loadProgressForUser(viewer.id)
-  }, [viewer?.id])
+    loadProgressForUser(viewerId)
+  }, [viewerId])
```
joelhooks, line 62: I find effect dependencies confusing so I like to do this to make them marginally less confusing. YMMV 😬

src/pages/user/index.tsx
```diff
@@ -25,33 +24,45 @@ const GithubConnectButton: React.FunctionComponent<{
...
+  // prefer the account with a subscription, otherwise grab the first, or just
+  // an empty object (which would be an error state, but possible)
+  return (
+    find<ViewerAccount>(
+      accounts,
+      (account: ViewerAccount) => account.subscriptions?.length > 0,
+    ) ||
+    first<ViewerAccount>(accounts) || {slug: ''}
+  )
 }
```
jbranchaud, line 40: Should we also filter/find based on the `subscription.quantity`? My assumption is that the _Account_ page is for viewing individual subscriptions, in which case we'd only want to grab an account that has a subscription with a quantity (seat) of `1`.

I'm also curious to know how we are ending up with users that have multiple accounts. Is it Bad Data™ or is it a valid scenario that we need to better account for? Maybe some data migration scenario led to a user having multiple accounts 🤔

jbranchaud (approved):

PR #617, opened 2021-05-10
abstraction for a card component

joelhooks:
We had three separate exported implementations of a "horizontal card" across the site. Looked like cut-n-paste to me.

In this PR I introduce a new generic `Card` base. It's doing considerably less than the previous `Card`.

I'll add some more specific commentary inline.


src/components/card/index.tsx
```diff
@@ -1,13 +1,170 @@
...
+}
+
+type CardProps = {
+  quiet?: boolean
+  horizontal?: boolean
+}
+
+const cardDefaultClasses = `bg-white dark:bg-gray-800 dark:text-gray-200 shadow-sm rounded-lg overflow-hidden sm:p-8 p-5`
+
+const Card = React.forwardRef(function Card(
```
joelhooks, line 24: I used some patterns in Reach for this component, and it's "overkill" but it also creates a very nice markup structure when it compiles and gives us a lot of forwards options with the card.

src/components/card/index.tsx
```diff
@@ -1,13 +1,170 @@
...
+  {
+    children,
+    as: Comp = 'div',
+    quiet = false,
+    horizontal = false,
```
joelhooks, line 29: I put `horizontal` in here thinking that it might be where we want to control the general layout, but it would require quite a bit of internal complexity and defining rules/styles to manage it so in the end, I decided that we could use these for semantic structure primarily and let the internal layout be mostly managed in composed structures like `HorizontalResourceCard`

left it in for discussion

src/components/card/index.tsx
```diff
@@ -1,13 +1,170 @@
...
   return (
-    <div
-      className={`bg-white dark:bg-gray-800 dark:text-gray-200 shadow-sm rounded-lg overflow-hidden sm:p-5 p-4`}
-    >
+    <CardContext.Provider value={context}>
+      <Comp
+        className={className}
+        {...props}
+        ref={forwardRef}
+        data-egghead-card=""
```
joelhooks, line 54: Reach does this and it allows you to target elements with global styles. I think this is an interesting approach!

src/components/layouts/collection-page-layout.tsx
```diff
@@ -638,7 +638,7 @@ const CollectionPageLayout: React.FunctionComponent<CoursePageLayoutProps> = ({
                   {relatedResources.map((resource: any) => {
                     return (
                       <div key={resource.slug}>
-                        <CardHorizontal
+                        <HorizontalResourceCard
```
joelhooks, line 641: `HorizontalResourceCard` is a drop in replacement for `CardHorizontal`

src/components/pages/home/index.tsx
```diff
@@ -514,73 +506,6 @@ type CardProps = {
   memberTitle?: string
 }
 
-export const CardHorizontal: FunctionComponent<{
```
joelhooks, line 517: this is gone now replaced with `HorizontalResourceCard`

src/components/card/horizontal-resource-card.tsx
```diff
@@ -0,0 +1,85 @@
...
+const HorizontalResourceCard: React.FC<any> = ({
+  resource,
+  location,
+  className = 'border-none my-4',
+  ...props
+}) => {
+  className = `${className} flex sm:flex-row flex-col sm:space-x-5 space-x-0 sm:space-y-0 space-y-5 items-center sm:text-left text-center`
+
+  return (
+    <Card {...props} className={className}>
```
joelhooks, line 25: fundamentally all of this translates to a stack of divs, so the abstraction feels heavy right now, but I think it sets the stage for a robust and flexible card to iterate on.

Still lots of decisions to make. I like the semantics that this structured approach makes in the component hierachy.

DrShpongle:
That's looking cool and seems like solid basement for the farther improvement/detailing 👍

src/components/card/index.tsx
DrShpongle: Totally agree with such approach when we have separate wrapper for each type of card with ability to manage card's subcomponents instead of complicating Card component itself.

src/components/card/index.tsx
```diff
@@ -1,13 +1,170 @@
...
+    children,
+    as: Comp = 'div',
+    quiet = false,
+    horizontal = false,
+    className,
+    ...props
+  },
+  forwardRef,
+) {
+  if (!quiet) {
```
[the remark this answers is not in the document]
zacjones93: ah I like this

zacjones93 (approved):
Love the flexibility this first iteration allows, this looks great!

PR #621, opened 2021-05-11
replace legacy card throughout application

joelhooks:
this removes all references to the previous `Card` component

there are some alignment and structure tweaks that need to happen still


theianjones (approved):
Nice. Love that theres way more lines removed than added 👏

PR #623, opened 2021-05-12
fix css page

vojtaholik:
this fixes [/q/css](https://egghead.io/q/css)


src/components/search/curated/css/css-page-data.ts
```diff
@@ -11,8 +11,6 @@ const CSSPageData: any = {
       resources: [
         {
           _id: '099b01d5-f852-4d74-81a6-7e23e84527ab',
-          image:
-            'https://d2eip9sf3oo6c2.cloudfront.net/tags/images/000/000/175/full/csslang.png',
```
vojtaholik, line 15: I had removed images from level cards (beginner, intermediate, advanced) because `VerticalResourceCollectionCard` is assuming that if there's an image, there's also a link, which isn't the case here. And we also weren't rendering the image anyway.

joelhooks:
feels like the vertical resource card should be smarter imo (as well as this)

joelhooks (approved):

PR #626, opened 2021-05-13
online presence articles (via sanity)

joelhooks:
This migrates some email quizzing components over from testing accessibility and also introduces article loading from Sanity using mdx-bundler.

This required an upgrade to node 14.


src/pages/own-your-online-presence/[slug].tsx
```diff
@@ -0,0 +1,68 @@
...
+      <main className="prose dark:prose-dark sm:prose-lg lg:prose-xl mt-5 max-w-none">
+        <Component />
+      </main>
+    </article>
+  )
+}
+
+export default OnlinePresenceArticle
+
+export const getServerSideProps: GetServerSideProps = async function ({
```
Comment (joelhooks) [inline src/pages/own-your-online-presence/[slug].tsx:28]: this is the functional work in this PR. The actual quiz stuff is a `noop`  and will be connected in a future PR.

package.json
```diff
@@ -122,6 +123,7 @@
     "tincanjs": "^0.50.0",
     "ts-toolbelt": "^9.6.0",
     "video.js": "^7.11.8",
+    "xdm": "^1.11.0",
```
kentcdodds, line 126: Shouldn't need to depend on this directly. It's brought in automatically by mdx-bundler.

src/pages/own-your-online-presence/[slug].tsx
```diff
@@ -0,0 +1,68 @@
...
+  `
+
+      const {resource} = await sanityClient.fetch(articleQuery)
+
+      const result = await bundleMDX(resource.article, {
+        xdmOptions(options) {
+          options.remarkPlugins = [...(options.remarkPlugins ?? [])]
+          return options
+        },
+      })
```
Comment (kentcdodds) [inline src/pages/own-your-online-presence/[slug].tsx:50]: Since you're not adding more `remarkPlugins`, you can remove this altogether.

```suggestion
      const result = await bundleMDX(resource.article)
```

kentcdodds (commented):
Cool! Just two thoughts.

PR #636, opened 2021-05-18
implements dynamic header CTA button

joelhooks:
We have the ability to examine a learners known attributes and dynamically diplay a CTA in the header based on what we know. This is a simple example with a switch statement that could be expanded to include more variety of CTAs.

### DIFF
--- a/src/components/app/header.tsx
+++ b/src/components/app/header.tsx
@@ -12,10 +12,12 @@ import {useTheme} from 'next-themes'
 import useCio from '../../hooks/use-cio'
 import {Form, Formik} from 'formik'
 import ProjectClubCTA from '../survey/project-club'
+import OnlinePresenceCTA from '../survey/online-presence-cta'
 
 const Header: FunctionComponent = () => {
   const router = useRouter()
   const {viewer, loading} = useViewer()
+  const {subscriber, loadingSubscriber} = useCio()
   const {sm} = useBreakpoint()
   const [isOpen, setOpen] = React.useState<boolean>(false)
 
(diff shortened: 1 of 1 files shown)


src/components/app/header.tsx
```diff
@@ -35,6 +37,20 @@ const Header: FunctionComponent = () => {
       ),
     )
 
+  let ActiveCTA: React.FC = () => null
+
+  switch (true) {
```
joelhooks, line 42: the first case below that is `true` will win.

we might consider changing it to "become a member" as a default for folks that aren't? lots of ways to adjust this.

tayiorbeii: +1 to "Become a Member" as default CTA for now

tayiorbeii (approved):

MaggieAppleton:
I'd really like us to save the CTA in the navbar only for core business actions – sign up new users with a 'Get Started' CTA, convert free users to members with a 'become a member' CTA, and then move to a CTA-free navbar for full members.

Things like 'own your online presence' and various clubs would work great as announcement banners right underneath the navbar that users can close/dismiss if they're not interested (or have already signed up)

This banner design is very basic/placeholder and the CTA button on it can say something more compelling than 'Learn More,' but just mocked it up to explain the banner concept.

image

This approach saves the big blue permanent navbar buttons only for core membership-related actions, but allows us the flexibility to let users know about fleeting campaigns like clubs and self-paced value paths in a very noticeable way.
Making it a dismissable banner also means it'll be less invasive or frustrating for members who don't have any interest in the offering.

MaggieAppleton:
I realised I just dumped a bunch of suggestions on here that go beyond the scope of customising the CTA button based on context ;)

I'll move these to a more complete pitch in the discussions. Stoked to see us add user context functionality so we can move towards the navbar dreams described above

MaggieAppleton:
Completed the writeup: https://github.com/eggheadio/egghead-next/discussions/637

PR #647, opened 2021-05-20
Fix turning off subtitles for new player

theianjones:
turn off

### DIFF
--- a/src/pages/lessons/[slug].tsx
+++ b/src/pages/lessons/[slug].tsx
@@ -5,11 +5,12 @@ import {isEmpty, get, first, isFunction, filter} from 'lodash'
 import {useMachine} from '@xstate/react'
 import {Tabs, TabList, Tab, TabPanels, TabPanel} from '@reach/tabs'
 import {playerMachine} from 'machines/lesson-player-machine'
-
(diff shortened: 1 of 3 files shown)


src/pages/lessons/[slug].tsx
```diff
@@ -512,14 +512,18 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
                             key={lesson?.subtitles_url}
                             order={11}
                             selected={subtitle}
-                            onChange={(track: TextTrack) => {
+                            onChange={(track?: TextTrack) => {
```
Comment (theianjones) [inline src/pages/lessons/[slug].tsx:515]: track is undefined when you turn subtitles off

src/pages/lessons/[slug].tsx
```diff
@@ -512,14 +512,18 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
...
+                              const updatedSubtitlePref = track
+                                ? {
+                                    id: track.id,
+                                    kind: track.kind,
+                                    label: track.label,
+                                    language: track.language,
+                                  }
+                                : defaultSubtitlePreference
```
Comment (theianjones) [inline src/pages/lessons/[slug].tsx:523]: if track is undefined, use the player default preference (off)

joelhooks (approved):

PR #649, opened 2021-05-21
Curated Topic Page - Node

laurosilvacom:
This curates the Node.js topic page. Part of the gardening efforts outlined in https://github.com/eggheadio/egghead-next/discussions/646.

[Live preview](https://egghead-io-nextjs-git-ls-topic-page-eggheadio1.vercel.app/q/node)

image


src/components/search/curated/node/index.tsx
```diff
@@ -1,17 +1,101 @@
 import React from 'react'
+import {NextSeo} from 'next-seo'
+import Topic from '../../components/topic'
 import nodePageData from './node-page-data'
```
laurosilvacom, line 4: We're still loading the data statically. Sanity data can't be sourced within a component, only inside `/pages`. 

Should we be sourcing data from Sanity for the topic pages? 

What's the balance between making sure that the work gets done and creating a new pattern?

lsminter (commented):
This looks beautiful! No issues from my point of view.

lsminter (approved):

PR #653, opened 2021-05-24
Update button styles in navbar

MaggieAppleton:
This changes the styles on two buttons in the navbar. It's a baby step towards some of the changes proposed in this pitch – https://github.com/eggheadio/egghead-next/discussions/637

The eventual goal here is to reduce the number of boldly styled CTAs in the header, and direct users attention to the main ones: "Get Started" (onboarding flow for new users) and "Become a Member"

1. Changes the 'Browse' button in the navbar to be a quiet secondary button (no blue background, gray-100 background on hover)
2. Removes the blue 'search' button from the search bar in favour of users pressing enter to complete their search query.

---

Preview of the updated navbar:

image

---

Preview of hover/focus styles on each element.

image

This doesn't change the functionality of any of the buttons. Checked it on mobile and with keyboard navigation and everything still works as expected.


DrShpongle:
Looks good to me 👍

joelhooks (approved):
this is great Maggie, thanks! As this overall plan unfolds I'd love if we could start measuring the effects of the changes on click rates and that sort of thing.

the strategy can be pretty simple by identifying actions and giving them the appropriate metadata (in this case I think `location` is plenty) and then creating the reporting views

if there are any bits you'd like to measure that way, let me know and I can help configure it

PR #655, opened 2021-05-24
update sanity query to be consistent with rails

zacjones93:
The sanity data coming in was different than the rails attribute so when the course data wasn't set in sanity, no instructor name would be shown. I converted the sanity data to be consistent with the rails data so now the objects merge instead of having separate `slug` and `url` attributes. We were only checking for `url` on the front end.

While making a sanity object for a course is a lot easier now, this shouldn't ever be an issue or something we should have to check

tweak

### DIFF
--- a/src/components/layouts/collection-page-layout.tsx
+++ b/src/components/layouts/collection-page-layout.tsx
@@ -132,8 +132,7 @@ const CollectionPageLayout: React.FunctionComponent<CoursePageLayoutProps> = ({
         byline: 'Colby Fayock・1h 4m・Course',
         image:
           'https://d2eip9sf3oo6c2.cloudfront.net/playlists/square_covers/000/412/781/thumb/ecommerce-stripe-next.png',
-        path:
-          '/playlists/create-an-ecommerce-store-with-next-js-and-stripe-checkout-562c',
+        path: '/playlists/create-an-ecommerce-store-with-next-js-and-stripe-checkout-562c',
         slug: 'create-an-ecommerce-store-with-next-js-and-stripe-checkout-562c',
         description: `This is a practical project based look at building a working e-commerce store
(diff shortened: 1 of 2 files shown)


src/components/layouts/collection-page-layout.tsx
```diff
@@ -257,8 +256,13 @@ const CollectionPageLayout: React.FunctionComponent<CoursePageLayoutProps> = ({
     [],
   ).map((lesson: LessonResource) => lesson.slug)
 
-  const {name, avatar_url, url: instructor_slug, bio_short, twitter} =
-    instructor || {}
+  const {
+    full_name: name,
+    avatar_url,
+    slug,
```
zacjones93, line 262: converted `name` to `full_name` in the sanity query to match the rails `full_name` attribute that comes in.

zacjones93: similarly, using `slug` to be consistent with rails

src/components/layouts/collection-page-layout.tsx
```diff
@@ -420,7 +424,7 @@ const CollectionPageLayout: React.FunctionComponent<CoursePageLayoutProps> = ({
                   <InstructorProfile
                     name={name}
                     avatar_url={avatar_url}
-                    url={instructor_slug}
+                    url={slug}
```
zacjones93, line 427: use slug

src/lib/courses.ts
```diff
@@ -34,9 +35,9 @@ const courseQuery = groq`
     'image': person->image.url
   },
 	'instructor': collaborators[]->[role == 'instructor'][0]{
-    'name': person->name,
+    'full_name': person->name,
     'avatar_url': person->image.url,
-    'url': person->slug.current
+    'slug': person->slug.current
```
zacjones93, line 40: update sanity query for consistency with rails

joelhooks (approved):

PR #659, opened 2021-05-27
En/initial cue popover

DrShpongle:
Logic of show/hide still needs polish!

### DIFF
--- a/src/components/player/cue-bar.tsx
+++ b/src/components/player/cue-bar.tsx
@@ -67,6 +67,7 @@ export default class CueBar extends React.Component<any> {
 const NoteCue: React.FC<any> = ({cue, duration, className}) => {
   const note = JSON.parse(cue.text)
   const [active, setActive] = React.useState(false)
+  const [popupIsShown, setPopupIsShowing] = React.useState(false)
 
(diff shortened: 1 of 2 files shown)


src/components/player/cue-bar.tsx
```diff
@@ -49,24 +53,65 @@ export default class CueBar extends React.Component<any> {
...
       </div>
     )
   }
 }
 
-const NoteCue: React.FC<any> = ({cue, duration, className}) => {
+const CuePopup: React.FC<any> = ({note, popupPersists, currentCuePopup}) => {
+  return note?.title === currentCuePopup || popupPersists ? (
+    <div
+      className="absolute w-40 min-h-[4rem] rounded-md bg-white p-3 text-xs top-0 left-0 z-10 text-black border border-gray-400"
```
joelhooks, line 81: when in the context of the player I think being consistent with the player CSS styles would be beneficial

same structure, same naming conventions, use of global css, etc

this means no Tailwind or emotion

this isn't critical, but I think it will also make the player work more portable for us so easier to just do it now instead of later

joelhooks (approved):

PR #679, opened 2021-06-07
fix state at end of completed video

joelhooks:
image

### DIFF
--- a/src/components/pages/lessons/overlay/course-pitch-overlay.tsx
+++ b/src/components/pages/lessons/overlay/course-pitch-overlay.tsx
@@ -1,0 +1,92 @@
+import Link from 'next/link'
+import Image from 'next/image'
+import * as React from 'react'
+import {track} from 'utils/analytics'
+import noop from 'utils/noop'
+
+const CoursePitchOverlay: React.FunctionComponent<{
+  lesson: any
+  onClickRewatch?: () => void
+}> = ({lesson, onClickRewatch = noop}) => {
+  const courseImage = lesson?.collection?.square_cover_480_url
+  return (
+    <>
+      {courseImage && (
+        <div className="w-16 h-16 md:w-32 md:h-32 lg:w-40 lg:h-40 xl:w-48 xl:h-48 relative flex-shrink-0">
+          <Image
+            src={courseImage}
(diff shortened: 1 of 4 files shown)


src/components/pages/lessons/overlay/course-pitch-overlay.tsx
```diff
@@ -0,0 +1,92 @@
+import Link from 'next/link'
+import Image from 'next/image'
+import * as React from 'react'
+import {track} from 'utils/analytics'
+import noop from 'utils/noop'
+
+const CoursePitchOverlay: React.FunctionComponent<{
```
joelhooks, line 7: direct cut and paste from the next lesson overlay

src/pages/lessons/[slug].tsx
```diff
@@ -136,8 +137,25 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
...
+  const [isIncomingAnonViewer, setIsIncomingAnonViewer] =
+    React.useState<boolean>(false)
+
+  React.useEffect(() => {
+    const storage = window?.sessionStorage
+    if (!storage) return
+
+    const prevPath = storage.getItem('prevPath')
+
+    if (!prevPath && !subscriber && !viewer) {
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:152]: if there is no previous path logged then we know they just dropped into this lesson and they aren't anon if we have a subscriber or viewer

src/machines/lesson-player-machine.ts
```diff
@@ -27,12 +28,13 @@ export type PlayerStateEvent =
   | {type: 'PLAY'}
   | {type: 'PAUSE'}
   | {type: 'LOADED'}
-  | {type: 'LOADED'; lesson: any; viewer: any}
+  | {type: 'LOADED'; lesson: any; viewer: any; isIncomingAnonViewer?: boolean}
```
jbranchaud, line 31: You added `isIncomingAnonViewer` to the type, but it isn't set or used within the machine. It looks like the lesson page gets by fine with it as component local state, so I think you can ditch the type change here.

jbranchaud, line 31:
```suggestion
  | {type: 'LOADED'; lesson: any; viewer: any}
```

PR #683, opened 2021-06-09
Updated the course title on a course card

lsminter:
The title for `Getting Started with Redux` changed to `Fundamentals of Redux Course from Dan Abramov`. 

OLD
Screen Shot 2021-06-09 at 11 53 51 AM

NEW
Screen Shot 2021-06-09 at 12 13 13 PM

The course cards are hardcoded in `src/components/search/instructors/` so I updated that in this PR.

### DIFF
--- a/src/components/search/instructors/dan-abramov/dan-abramov-page-data.ts
+++ b/src/components/search/instructors/dan-abramov/dan-abramov-page-data.ts
@@ -23,8 +23,8 @@ He started using React for front-end development when he was working with a U.S-
     name: 'Just starting out with JavaScript',
     resources: [
       {
-        title: 'Getting Started with Redux',
-        path: '/courses/getting-started-with-redux',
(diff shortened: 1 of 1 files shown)


joelhooks:
as a rule matching the link text/title to the specific page/content title isn't critical as long as the link text gives the learner enough context to make a decision


it's also fine if it does match, but we don't need to specifically be too concerned

lsminter:
Ok. Lauro and I both thought that the new title sounded a little weird. `Fundamentals of Redux Course from Dan Abramov` doesn't flow very well. I was thinking about changing it to `Fundamentals of Redux`, excluding `Course` plus the instructor's name since it's always displayed wherever we display the course name.

joelhooks:
this particular course can probably get left alone unless you are running experiments and tracking analytics (versus going by gut feel/assumption)

PR #689, opened 2021-06-11
CuePlayer: Update control bar layout and tooltips

DrShpongle:
image


DrShpongle:
full screen view:

Image 2021-06-11 at 9 17 25 PM


and desktop:

screencapture-egghead-af-3000-video-test-2021-06-11-21_18_00

src/pages/video-test.tsx
```diff
@@ -48,19 +48,20 @@ const EggheadPlayer: React.FC<{videoResource: VideoResource}> = ({
...
         >
-          <div className="relative z-10 lg:col-span-9">
+          <div className="relative z-10 lg:col-span-9 pb-[4.5rem]">
             <Player
               muted
               autoplay
               crossOrigin="anonymous"
               className="font-sans"
               poster={videoResource.poster}
+              // className="cueplayer-react-fullscreen"
```
DrShpongle, line 64: uncomment it and switch to fullscreen to get styles applied

joelhooks: this can be removed
```suggestion
```

src/pages/video-test.tsx
```diff
@@ -174,6 +177,38 @@ const NotesTabContent: React.FC<{cues: VTTCue[]}> = ({cues}) => {
   )
 }
 
+const PlayerContainer: React.ForwardRefExoticComponent<any> = React.forwardRef<
```
joelhooks, line 180: I made the container a component so we could dynamically add classes, similar to `Player` - actually just cut and pasted most of those classes, we might make these different? You get the idea.

DrShpongle:
- Removed side bar in full screen mode
- added "divider" to control bar so that fullscreen toggle, options bar and speed rate are shifted to the right side

DrShpongle:
Here is sketch of how it might look made by @vojtaholik:

joelhooks (approved):

PR #690, opened 2021-06-11
Hiroko's name was being exported as Hiro

lsminter:
There was a bug where Hiro had two names, Hiro and Hiroko. If you search for Hiro, three lessons would should up, if you searched for Hiroko, all of her lessons would show up. 

Screen Shot 2021-06-11 at 11 43 02 AM

Screen Shot 2021-06-11 at 12 00 13 PM

In `egghead-next/src/components/search/instructors/index.tsx`, Hiroko's name was being exported as `hiro` instead of `hiroko` and I believe that's why these were showing up. 

Screen Shot 2021-06-11 at 1 08 47 PM

After changing the name, nothing shows up after searching for `hiro`.


src/components/search/instructors/index.tsx
```diff
@@ -16,7 +16,7 @@ const InstructorsIndex: any = {
   'john-lindquist': SearchJohnLindquist,
   'laurie-barth': SearchLaurieBarth,
   'flavio-corpa': SearchFlavioCorpa,
-  'hiro-nishimura': SearchHirokoNishimura,
+  'hiroko-nishimura': SearchHirokoNishimura,
```
lsminter, line 19: Just this small change.

zacjones93:
This looks great, we need to be consistent. 

The slug in the instructor page data needs to be updated as well

https://github.com/eggheadio/egghead-next/blob/main/src/components/search/instructors/hiroko-nishimura/hiroko-nishimura-page-data.ts#L8

zacjones93 (changes requested):
This looks great, we need to be consistent.

The slug in the instructor page data needs to be updated as well

https://github.com/eggheadio/egghead-next/blob/main/src/components/search/instructors/hiroko-nishimura/hiroko-nishimura-page-data.ts#L8

Also, it's company policy for all PR's to include a GIF.

src/components/search/instructors/hiroko-nishimura/hiroko-nishimura-page-data.ts
```diff
@@ -5,7 +5,7 @@ const HirokoNishimuraPageData = [
     bio: `AWS Community Hero. Founder of AWS Newbies and LinkedIn Learning instructor.`,
     twitterHandle: 'hirokonishimura',
     websiteURL: 'https://hirokonishimura.com/',
-    slug: 'resources-by-hiro-nishimura',
+    slug: 'resources-by-hiroko-nishimura',
```
lsminter, line 8: Completely missed this one. Thanks Zac.

zacjones93 (approved):

PR #697, opened 2021-06-15
adds cues as array and uses actions

joelhooks:
Screen Recording 2021-06-15 at 02 17 48 PM


src/components/player/cue-bar.tsx
```diff
@@ -46,10 +46,10 @@ const useCue = (cue: VTTCue, actions: any) => {
       if (active) {
         actions.activateMetadataTrackCue(cue)
       } else {
-        actions.activateMetadataTrackCue(null)
+        actions.deactivateMetadataTrackCue(cue)
```
joelhooks, line 49: added a new deactivate action

src/components/player/cue-bar.tsx
```diff
@@ -79,22 +79,23 @@ const NoteCue: React.FC<any> = ({
...
-  const show = () => setVisible(true)
-  const hide = () => setVisible(false)
+
   const open = () => {
-    show()
-    setPersist(true)
-    onClickCue()
+    setVisible(true)
+    // if we seek to the correct time, the note is displayed
+    actions.seek(cue.startTime)
```
joelhooks, line 89: there was a difference here with show/open, but I removed that for now. IMO a cue is active or not so I'm relying on that mostly

also, instead of tracking clicks or other approach I "seek" the player to the appropriate time (cue start) and it "just works"

src/components/player/cue-bar.tsx
```diff
@@ -79,22 +79,23 @@ const NoteCue: React.FC<any> = ({
...
   const close = () => {
-    hide()
-    setPersist(false)
+    setVisible(false)
   }
-  const visible = cue === player.activeMetadataTrackCue
+
+  // added seeking to the list here but getting some janky perf issues
+  const visible =
+    player.activeMetadataTrackCues.includes(cue) && !player.seeking
```
joelhooks, line 98: this prevents cue popups when actively seeking. I'm sure this is why eduweb uses the full height seek bar

src/components/player/cue-bar.tsx
```diff
@@ -124,9 +125,7 @@ const NoteCue: React.FC<any> = ({
       onClickOutside={close}
     >
       <div
-        onMouseOver={show}
         onClick={open}
-        onMouseOut={() => !persist && hide()}
```
joelhooks, line 129: just removing this for now, they gotta click

src/components/player/cue-bar.tsx
vojtaholik: this sounds good in theory, however I'm getting weird results: 

plus sidebar stopped scrolling to active note?

src/components/player/cue-bar.tsx
vojtaholik: nice!

src/components/player/cue-bar.tsx
```diff
@@ -111,9 +111,7 @@ const NoteCue: React.FC<any> = ({
               {note.title}
             </span>
           )}
-          <TruncateMarkup lines={2}>
-            <div>{note.description}</div>
-          </TruncateMarkup>
+          <div className="line-clamp-2">{note.description}</div>
```
DrShpongle, line 114: we have really good support of [webkit-line-clamp](https://caniuse.com/?search=webkit-line-clamp) CSS feature

PR #698, opened 2021-06-16
adds dynamic loading to video-test

joelhooks:


package.json
```diff
@@ -126,8 +132,11 @@
     "tailwindcss": "^2.1.4",
     "tincanjs": "^0.50.0",
     "ts-toolbelt": "^9.6.0",
+    "unist-util-visit": "^2.0.3",
```
joelhooks, line 135: this is an older version of this library

https://github.com/syntax-tree/unist-util-visit/issues/23

apparently the entire unified ecosystem is making a principled stance on using ESM modules which is incompatible with Next.js (and most of the rest of the world) at this time. This causes a massive headache and debugging nightmare to implement, but it's a principled stance and open source so you can only complain about it so much I guess.

Kind of sucks when you just want to make stuff, particularly since it'd be relatively simple to support commonjs syntax for now.

This issue, in fact, has caused mdx to be forked :trollface: 

I don't like it.

src/components/player/cue-bar.tsx
```diff
@@ -98,7 +98,7 @@ const NoteCue: React.FC<any> = ({
   const visible =
     player.activeMetadataTrackCues.includes(cue) && !player.seeking
   const startPosition = `${(cue.startTime / duration) * 100}%`
-  const note = JSON.parse(cue.text)
+  const note = cue.text
```
joelhooks, line 101: not using JSON, which might be something we test for in the future. We **can** just not needed for the current approach.

src/pages/api/github-load-notes.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    res.status(200).json(test)
+  } else {
+    res.status(200).end()
+  }
+}
+
+export default loadGithubNotes
+
+export async function loadNotesFromUrl(url: string) {
```
joelhooks, line 20: this loads mdx from a url and parses it looking for timestamp components which creates JS objects that are used to make a webvtt.

it's currently very fragile! works though

DrShpongle (approved):
great update 👍

package.json
ChristianMurphy: Maintainer of unified and mdx here, sorry you ran into a spot of trouble.
This characterization is unfair on a few fronts:
1. It is not a libraries fault if a build tool does not support standard JavaScript syntax.
2. Next has been working to add ESM support, and merged https://github.com/vercel/next.js/pull/27069 to address this specifically
3. MDX was forked over a disagreement on how to handle presentation context and runtime https://github.com/mdx-js/mdx/pull/1425 and https://github.com/mdx-js/mdx/issues/1385, it has nothing to do with ESM
4. Unified had experimented with adding a CJS fallback to support older versions of CRA and Next, but due to some quirks of webpack 4, it actually made things harder for adopters than pure ESM :man_shrugging: 

If you have questions on ESM or MDX in general, feel free to reach out to the MDX community https://github.com/mdx-js/mdx/discussions

joelhooks: Thanks for this detailed list of corrections.

PR #699, opened 2021-06-16
Add Ownership Transfer for Team Accounts

jbranchaud:
- Add form to transfer ownership from Team Mgmt page
- Add page to accept an Ownership Transfer invite

https://user-images.githubusercontent.com/694063/122292628-cc7ecb80-cebb-11eb-9e0a-452d6f8b3aeb.mp4


src/pages/account-ownership-transfer/[guid].tsx
```diff
@@ -0,0 +1,172 @@
...
+  }
+}
+
+const AccountOwnershipTransfer: React.FunctionComponent<accountOwnershipTransferData> =
+  ({validInvite, ownerEmail}) => {
+    const router = useRouter()
+    const {guid} = router.query
+
+    console.log({validInvite, ownerEmail})
+
```
Comment (jbranchaud) [inline src/pages/account-ownership-transfer/[guid].tsx:78]:
```suggestion
```

src/components/team/account-ownership-transfer.tsx
```diff
@@ -0,0 +1,103 @@
...
+          <button
+            className={`text-white bg-green-600 border-0 py-2 px-4 focus:outline-none rounded-md
+                    ${
+                      loading || inviteeEmail === ''
+                        ? 'cursor-not-allowed opacity-50'
+                        : 'hover:bg-green-700'
+                    }`}
+            type="button"
+            disabled={loading || inviteeEmail === ''}
+            onClick={async () => {
```
joelhooks, line 55: can we get some `track` calls in here? seems like good analytics

src/components/team/account-ownership-transfer.tsx
```diff
@@ -0,0 +1,103 @@
...
+                setInviteeEmail('')
+
+                toast.success(
+                  'Your account ownership transfer invitation has been sent.',
+                  {
+                    icon: '✅',
+                  },
+                )
+              } catch (e) {
+                toast.error(
```
joelhooks, line 70: when we start digging into these kinds of errors, how should we capture them on our side?

A honeybadger project might help. We had bad luck with their autocapture, but manually it could provide insight into business processes gone wrong.

src/pages/account-ownership-transfer/[guid].tsx
```diff
@@ -0,0 +1,170 @@
...
+  } catch (e) {
+    if (e.response.status === 404) {
+      props.validInvite = false
+    }
+  }
+
+  return {
+    props,
+  }
+}
```
Comment (joelhooks) [inline src/pages/account-ownership-transfer/[guid].tsx:70]: style: this was confusing for me because there is a mix and match style in this function. `let` is used so we can assign new references to `props` but then we also use property accessors to mutate the initial reference.

I think that **any** error would indicate an invalid invite?

This is how I'd think about it removing the variable entirely.

```suggestion
export async function getServerSideProps(context: any) {
  const {guid} = context.params
  const {eggheadToken} = getTokenFromCookieHeaders(
    context.req.headers.cookie as string,
  )

  try {
    const {data}: {data: {owner_email: string}} = await axios.get(
      `${AUTH_DOMAIN}/api/v1/account_ownership_transfer_invitations/${guid}`,
      {
        headers: {
          Authorization: `Bearer ${eggheadToken}`,
        },
      },
    )
    
    return {props: {
      validInvite: true,
      ownerEmail: data.owner_email,
    }}
  } catch (e) {
    res.setHeader('Location', initialLesson.path)
    res.statusCode = e.response.status
    console.error(e.response.data.msg, guid)
    res.end()
    return {props: {validInvite = false}}
  }
}
```

src/pages/account-ownership-transfer/[guid].tsx
```diff
@@ -0,0 +1,170 @@
...
+
+    transferSucceeded = true
+  } catch (e) {
+    transferSucceeded = false
+  }
+
+  return transferSucceeded
+}
+
+type accountOwnershipTransferData = {
```
Comment (joelhooks) [inline src/pages/account-ownership-transfer/[guid].tsx:32]: style: types, by convention in this codebase anyway, are always CamelCase

src/pages/account-ownership-transfer/[guid].tsx
```diff
@@ -0,0 +1,170 @@
...
+import toast from 'react-hot-toast'
+import axios from 'axios'
+import {
+  getAuthorizationHeader,
+  getTokenFromCookieHeaders,
+  AUTH_DOMAIN,
+} from 'utils/auth'
+
+async function confirmAccountOwnershipTransfer(guid: string) {
+  let transferSucceeded
```
Comment (joelhooks) [inline src/pages/account-ownership-transfer/[guid].tsx:13]: style[nbd]: for most of this codebase, we avoid `let` entirely and leverage `const` with multiple returns. This isn't critical, but it's a style drift so I'm bringing it up.

is using `let` and sticking to a single return value a general approach you prefer?

src/pages/account-ownership-transfer/[guid].tsx
```diff
@@ -0,0 +1,170 @@
...
+                <div className="flex justify-center items-center w-full">
+                  <button
+                    className={`text-white bg-red-500 border-0 py-2 px-8 focus:outline-none rounded
+                    ${
+                      validInvite
+                        ? 'hover:bg-red-600'
+                        : 'cursor-not-allowed opacity-50'
+                    }`}
+                    disabled={!validInvite}
+                    onClick={async () => {
```
Comment (joelhooks) [inline src/pages/account-ownership-transfer/[guid].tsx:136]: this could use analytics as well!

joelhooks (approved):
I left mostly style comments so approving this and adding an RFC tag so you can remove it to merge when you'd like.

Nothing critical.

src/components/team/account-ownership-transfer.tsx
jbranchaud: I'm adding Ahoy tracking calls for both the successful and failed invite requests.

It would be nice to have a more dedicated error capture approach setup. I had assumed that Honeybadger was probably already being used throughout this codebase, but I did a search and see that it isn't.

We can start capturing specific things like this using the Honeybadger JS lib and see how that works for errors in these kinds of business processes.

I agree that we should use a separate Honeybadger project for this.

src/pages/account-ownership-transfer/[guid].tsx
Comment (jbranchaud) [reply src/pages/account-ownership-transfer/[guid].tsx:13]: I default to using `const` 99% of the time. If I end up using `let`, it is generally because I'm finding myself needing a value at a higher level of scope than where it will be assigned.

Looking back at this function, the `let` isn't doing much for me here. I would prefer multiple returns in this case. I seem to remember that there was a lot more going on in this method and that the `let` is the artifact of a simplifying refactoring.

In Ruby code I prefer a single return over multiple returns if it doesn't make the code awkward. I find that makes the flow and logic easier to reason about. Especially in a language with implicit returns. I bet that preference bleeds into the JS that I write without even thinking about it.

PR #700, opened 2021-06-17
Cueplayer tweaks

DrShpongle:
What was done:
- fixed styles to prevent collision in build
- add SimpleBar (for nice scroll bars) to the Notes list

What I failed with:
- auto scroll to active note while video plays. It scrolls on click on Cue though, but it works weird (I suspect it is because we have multiple active cues, can't figure out exactly where the state is leaking).

https://share.getcloudapp.com/12u4Ax1p

### DIFF
--- a/src/pages/video-test.tsx
+++ b/src/pages/video-test.tsx
@@ -28,7 +28,7 @@ import ReactMarkdown from 'react-markdown'
 import CueBar from 'components/player/cue-bar'
 import ControlBarDivider from 'components/player/control-bar-divider'
 import {useEggheadPlayerPrefs} from 'components/EggheadPlayer/use-egghead-player'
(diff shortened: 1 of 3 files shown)


src/pages/video-test.tsx
```diff
@@ -127,17 +128,14 @@ const EggheadPlayer: React.FC<{
                   {!isEmpty(cues) && <Tab>Notes</Tab>}
                   <Tab>Lessons</Tab>
                 </TabList>
-                <TabPanels
-                  id="notes-tab-scroll-container"
```
DrShpongle, line 131: I think that we want to have two separate scrollable containers - one for notes, second for lessons. So I removed scrolling behavior form the `TabPanels`

joelhooks (approved):

PR #702, opened 2021-06-17
add basic styles and solve warnings for cueplayer

joelhooks:

### DIFF
--- a/src/components/player/control-bar-divider.tsx
+++ b/src/components/player/control-bar-divider.tsx
@@ -4,6 +4,6 @@ const ControlBarDivider: React.FC<{
   key: string
   order: number
   className: string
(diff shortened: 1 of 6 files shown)


src/components/player/control-bar-divider.tsx
```diff
@@ -4,6 +4,6 @@ const ControlBarDivider: React.FC<{
   key: string
   order: number
   className: string
-}> = (props) => <div {...props} />
+}> = ({className}) => <div className={className} />
```
joelhooks, line 7: for divider we don't need to push all props to div and seemed like only `className` was needed. This made a lot of warning disappear in console.

DrShpongle (approved):

PR #712, opened 2021-06-22
Lessons list in sidebar on video-test page

DrShpongle:
- got lesson's list in Lessons tab
- Lessons tab is "tab 0" now

qqq


src/pages/video-test.tsx
```diff
@@ -267,6 +282,7 @@ export const getServerSideProps: GetServerSideProps = async function ({
   return {
     props: {
       videoResource,
+      currentLessonSlug: lesson,
```
joelhooks, line 285: i think this is always the slug of the videoResource?

src/types.ts
```diff
@@ -35,6 +35,7 @@ export type VideoResource = LessonResource & {
   hls_url: string
   thumb_url: string
   subtitles_url: string
+  collection: any
```
joelhooks, line 38: type here is []Resource perhaps

joelhooks (approved):

PR #716, opened 2021-06-22
Update course page to pull sanity dependency versions

zacjones93:
@lsminter noticed versions were not being pulled from sanity, this fixes that

hype


src/components/layouts/collection-page-layout.tsx
```diff
@@ -232,14 +233,22 @@ const CollectionPageLayout: React.FunctionComponent<CoursePageLayoutProps> = ({
...
   )
 
   logCollectionResource(course)
 
   const courseTags = tags.map((tag: any) => {
-    const version = get(dependencies, tag.name)
+    const ogVersion = get(dependencies, tag.name)
+    const sanityTag = find(sanityDependencies, {name: tag.name}).version
+
+    const version = !isEmpty(sanityTag) ? sanityTag : ogVersion
```
zacjones93, line 246: The sanity data is shaped a little different than how we were hard coding it in `courseDependencies.json` so I'm using find to grab the right dependency that matches the tag name then pulling the version off of that dependency.

`version` now switches on if a `sanityTag` is present or not

src/lib/courses.ts
```diff
@@ -46,7 +46,8 @@ const courseQuery = groq`
       "slug": slug.current,
       path,
       name,
-      'image': image.url
+      'label': name,
+      'image_url': image.url
```
zacjones93, line 50: update property names coming in from sanity to match current use

laurosilvacom (approved):

PR #721, opened 2021-06-23
Updates portfolio club copy

tayiorbeii:
* Updates landing page copy for the portfolio club
copy preview

handwriting gif


joelhooks:
i fundamentally think the club pitch is stronger and that this sequence can run even if the pitch is about a synchronous club experience

tayiorbeii:
> i fundamentally think the club pitch is stronger and that this sequence can run even if the pitch is about a synchronous club experience

Something I had thought about was combining the two a bit more, and removing the reference to an upcoming club.

"Knowing what goes into a badass portfolio is one thing, but joining a club is how you really get it done."

joelhooks:
removing the specific dates and such is absolutely needed but the idea i’m trying to express is that we are gathering interest specifically for live developer portfolio project clubs

the email series is intended to provide some i goal direction and not just gather email addresses and do nothing with them and leave them in silence

if we want to really do the work then we’d monitor, segment, and schedule chats with these folks to refine the portfolio project club concept and work towards a self-paced variant that might start from the **same initial CTA**

imo, this work should support the next clubs and gather candidates for participation

tayiorbeii:
@joelhooks @MaggieAppleton this copy is updated and ready for review

tayiorbeii:
Closing this PR for now

PR #726, opened 2021-06-23
switches to note tab automatically

joelhooks:

### DIFF
--- a/src/components/EggheadPlayer/use-egghead-player.ts
+++ b/src/components/EggheadPlayer/use-egghead-player.ts
@@ -214,7 +214,31 @@ export const defaultSubtitlePreference = {
   language: null,
 }
 
-const defaultPlayerPreferences = {
+export type PlayerPrefs = {
+  volumeRate: number
+  playbackRate: number
+
+  autoplay: boolean
+  videoQuality: {
+    bitrate: any
+    height: any
+    id: string
+    width: any
+  }
+  subtitle: {
+    id: any
+    kind: any
+    label: any
+    language: any
+  }
+  muted: boolean
+  theater: boolean
+  defaultView: string
+  activeSidebarTab: number
+  muteNotes: boolean
+}
+
+const defaultPlayerPreferences: PlayerPrefs = {
   volumeRate: 80,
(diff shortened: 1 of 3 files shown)


src/components/EggheadPlayer/use-egghead-player.ts
```diff
@@ -228,9 +252,8 @@ const defaultPlayerPreferences = {
   muted: false,
   theater: false,
   defaultView: 'transcript',
-  sideBar: {
-    activeTab: 0,
-  },
+  activeSidebarTab: 0,
```
joelhooks, line 255: flattened this for ease of access

src/pages/video-test.tsx
```diff
@@ -40,9 +40,10 @@ const EggheadPlayer: React.FC<{
   notesUrl: string
 }> = ({videoResource, notesUrl}) => {
   const playerContainer = React.useRef<any>()
-  const playerPrefs = useEggheadPlayerPrefs()
+  const {setPlayerPrefs, getPlayerPrefs} = useEggheadPlayerPrefs()
   const {player} = usePlayer()
   const {activeMetadataTracks = []} = player
+  const {activeSidebarTab} = getPlayerPrefs()
```
joelhooks, line 46: was having a difficult time getting this to update from the prefs hook until I made a getter which now reads at ender and updates

I thought calling `set` on a `useState` would trigger the update but for some reason we aren't getting a refreshed value even though the underlying state is update.

src/components/player/cue-bar.tsx
```diff
@@ -97,8 +98,14 @@ const NoteCue: React.FC<any> = ({
   }
 
   React.useEffect(() => {
-    setVisible(player.activeMetadataTrackCues.includes(cue) && !player.seeking)
-  }, [player.activeMetadataTrackCues, player.seeking, cue])
+    if (!muteNotes) {
```
joelhooks, line 101: this adds a preference but no UI to control it, set the preference to mute!

DrShpongle (approved):
Cool 👍

PR #727, opened 2021-06-24
updated at on courses and sanity preview url

joelhooks:
added a link to sanity preview and updated at back in

https://www.loom.com/share/fa80c136b4f643299c9694e44019fac4


zacjones93 (approved):
Love how simple adding that plugin is

PR #730, opened 2021-06-24
Use specific error messages from server when Ownership Transfer Invite cannot be sent

jbranchaud:
There are different ways that sending an Account Ownership Transfer
Invitation can fail. There are user-friendly messages coming from the
server explaining what went wrong. This commit wires those up to the
toast that displays on error.

https://user-images.githubusercontent.com/694063/123320811-dfb31c00-d4f7-11eb-9054-5099d30ce46f.mp4


joelhooks:
this is a comment about the specific message here, but I think we could still create the user and send a magic link?

joelhooks (approved):

PR #733, opened 2021-06-25
CuePlayer tweaks

DrShpongle:
- fixed autoscroll in SideBar (it didn't work before when Note Popups are muted)


joelhooks:
if something is “muted” i’d assume it would stop moving the screen around

DrShpongle:
> if something is “muted” i’d assume it would stop moving the screen around

I was sure that "mute" belongs to popups only. Going to revert.

joelhooks:
ya that’s why i did it. i don’t feel strongly if we think it should still scroll, but i was leaning towards respecting the learners request to make it “quite”

would still highlight, just not activate tab or autoscroll

PR #745, opened 2021-07-02
Pass ref and href to anchor from next/link

theianjones:
Since we have an onClick in our inner `<a/>` we need to pass the
ref, Link.onClick, and href to the `<a/>`.

[next link docs](https://nextjs.org/docs/api-reference/next/link)

link


joelhooks:
thanks for this work, but I don't think it was the right direction

PR #749, opened 2021-07-06
state management for notes

joelhooks:

### DIFF
--- a/src/components/pages/lessons/overlay/wrapper.tsx
+++ b/src/components/pages/lessons/overlay/wrapper.tsx
@@ -1,0 +1,15 @@
+import React, {FunctionComponent} from 'react'
+
+const OverlayWrapper: FunctionComponent<{
+  children: React.ReactNode
+}> = ({children}) => {
+  return (
+    <div className="lg:aspect-w-16 lg:aspect-h-9">
+      <div className="bg-gray-800 text-white bg-opacity-90 flex flex-col items-center justify-center px-4 py-6 min-h-[16rem] md:min-h-[24rem] lg:min-h-0">
+        {children}
+      </div>
+    </div>
+  )
+}
+
+export default OverlayWrapper
--- a/src/components/player/cue-bar.tsx
+++ b/src/components/player/cue-bar.tsx
@@ -6,6 +6,7 @@ import {scroller} from 'react-scroll'
 import {useEggheadPlayerPrefs} from 'components/EggheadPlayer/use-egghead-player'
 import ReactMarkdown from 'react-markdown'
 import {track} from 'utils/analytics'
+import {useNotesCues} from './index'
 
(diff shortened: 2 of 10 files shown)


src/components/player/index.tsx
```diff
@@ -57,9 +55,8 @@ const VideoResourcePlayer: React.FC<VideoResourcePlayerProps> = ({
   ...props
 }) => {
   const {setPlayerPrefs, getPlayerPrefs} = useEggheadPlayerPrefs()
-  const {hasNotes} = useNotesCues(videoResource)
+  const hasNotes = !isEmpty(videoResource?.staff_notes_url)
```
joelhooks, line 58: this was causing intense redraws

src/pages/lessons/[slug].tsx
```diff
@@ -420,157 +407,166 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
...
-                <CoursePitchOverlay
-                  lesson={lesson}
-                  onClickRewatch={() => {
-                    send('VIEW')
-                    if (actualPlayerRef.current) {
-                      actualPlayerRef.current.play()
-                    }
+            <div
+              className={`relative ${
+                isFullscreen ? 'lg:col-span-12' : 'lg:col-span-9'
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:419]: we had local state for fullscreen so I adapted this to use that instead of needing player

having `usePlayer` above caused a lot of redrawing!

src/components/pages/lessons/overlay/wrapper.tsx
```diff
@@ -0,0 +1,15 @@
+import React, {FunctionComponent} from 'react'
+
+const OverlayWrapper: FunctionComponent<{
```
joelhooks, line 3: just extracted this into it's own component

src/components/player/cue-bar.tsx
```diff
@@ -119,21 +116,21 @@ const NoteCue: React.FC<any> = ({
...
-      // don't automatically pop if muted
-      setVisible(
-        player.activeMetadataTrackCues.includes(cue) && !player.seeking,
-      )
-      setPlayerPrefs({activeSidebarTab: 1})
-    }
-  }, [player.activeMetadataTrackCues, player.seeking, cue, muteNotes])
+    const isVisible = !muteNotes && cueActive && !seeking && playerReadyEnough
+    setVisible(isVisible)
+  }, [setPlayerPrefs, cueActive, seeking, muteNotes, playerReadyEnough])
```
joelhooks, line 133: since these are all computer, it makes it more accurate to tun the effect, but I am still getting a brief flash of the previously active note when you switch lessons while it's playing

DrShpongle (approved):
That's great, thanks for comments 👍

Issue #722, opened 2021-06-23
Add Browse Flyout Menu to Navbar


MaggieAppleton:
Our learners routinely report that browsing the site is difficult and they have trouble finding content. This is a larger information architecture problem we need to holistically address in the future. I propose we add a browse flyout menu as one small, tangible step towards improving the browsing experience in the short term.

Our curated homepage approach helps people see what's new and featured, but doesn't help people get a big picture view of all the available content and topics.

We improved this slightly when we added a "browse" button to the navigation bar which links to `/topics`. However, our topics page is a long list of all available topics in alphabetical order. This doesn't direct people to pages where we have the most content (JavaScript, React, etc.) and doesn't make clear all the types of content (articles, podcasts, etc.)

I propose we add a flyout menu to our existing browse button. This will give users quicker, easier access to any topics page no matter where they are on the site.
This feature should:
- Show the flyout menu when learners click on the browse button
- Hide the flyout menu when learners click the browse button when the menu is open. If they click outside the menu anywhere else on the page it should also hide the flyout menu
- Offer a list of 12-15 topic links in order of popularity and content volume. Clicking one of these takes you to the relevant curated topic page. We can base these off stats from Mixpanel on our most clicked topics:
![image](https://user-images.githubusercontent.com/5599295/123085105-cc318380-d419-11eb-9af3-2adcb98fe8ab.png)

- Offer links to our `/podcasts`,  `/talks`, and `/articles` pages at the bottom to help learners discover and quickly access these
- Allow learners to still navigate to the full `/topics` page via a "browse all topics" button.
- Use the Tailwinds CSS flyout menu components as a base: https://tailwindui.com/components/marketing/elements/flyout-menus. We shouldn't need to deviate from these designs much.
- Follow the established behaviour many other educational websites use for this design pattern. Take a look at the browse flyout behaviour on [skillshare](https://www.skillshare.com/home), [masterclass](https://www.masterclass.com/), and [datacamp](https://www.datacamp.com/)
- All links in the menu should be wrapped with a tracking function that logs the location (browse menu) and the topic name (javascript, articles, etc.)
- Menu needs to meet a11y accessibility standards: [How to Create Drop-Down and Fly-Out Menus That Are Web-Accessible](https://blog.hubspot.com/website/accessible-drop-down-menus)

Mockups using the Tailwinds CSS base are available in Figma in the `egghead Design System` file. Look for the `Navigation` page in the left-hand sidebar to find them:

[...]

PR #753, opened 2021-07-07
Flyout Menu - Ready 

NicollGuarnizo:
Demo

Flyout Menu referred in this [issue](https://github.com/eggheadio/egghead-next/issues/722) : 
   1. Meet all the functionality asked in the issue
   2. Button is hidden in mobile version (as the current Browse button)
   3. Design for light&dark mode


MaggieAppleton:
This looks great!

The only thing we might want to consider adding here is tracking so we get to get data about how much people are using it.

We have a `track` component in `utils/analytics` that we can pass to each link. Here's an example of it being used on the main jumbotron: https://github.com/eggheadio/egghead-next/blob/fbe23ecce50a84ca89bb671a6fa8130fb41e8265/src/components/pages/home/jumbotron/index.tsx#L66

NicollGuarnizo:
Track components added

src/components/app/header/index.tsx
```diff
@@ -345,3 +346,155 @@ const IconX = () => (
...
+                        onClick={() => track(`clicked topic`, {resource: item.href})}
+                        className="flex items-start rounded-lg transition ease-in-out duration-150 hover:bg-gray-100 dark:hover:bg-gray-900 py-2 px-3"
+                      >
+                        <p className="text-base font-medium text-gray-700 transition ease-in-out duration-150 dark:text-white hover:text-black">
+                          {item.name}
+                        </p>
+                      </a>
+                    ))}
+                    <div className="mr-6 text-base font-medium  transition ease-in-out duration-150 rounded-lg hover:bg-gray-100 dark:hover:bg-gray-900 py-2 px-3">
+                      <a href="/topics" className="text-blue-500 onClick={() => track(`clicked all topics`, {location})}">
```
MaggieAppleton, line 470: Looks like the `onClick={() => track(clicked all topics, {location})}` function is declared inside `className` here

src/components/app/header/index.tsx
```diff
@@ -345,3 +346,155 @@ const IconX = () => (
...
+    {
+      name: 'Python',
+      href: '/q/python',
+    },
+    {
+      name: 'Go',
+      href: '/q/go',
+    },
+  ]
+  const callsToAction = [
```
MaggieAppleton, line 406: `callsToAction` doesn't feel like quite the right name here. Something like `contentTypes` or `contentTypeLinks` would be more accurate. Small nitpick, but clear naming helps keep this readable.

src/components/app/header/index.tsx
```diff
@@ -345,3 +346,155 @@ const IconX = () => (
...
+                className={classNames(
+                  open
+                    ? 'hover:bg-gray-100 dark:hover:bg-gray-800'
+                    : 'active:bg-gray-200 dark:hover:text-white',
+                  'ml-2 h-5 w-5',
+                )}
+                aria-hidden="true"
+              />
+            </Popover.Button>
+            {/* @ts-expect-error */}
```
MaggieAppleton, line 440: Can we remove this comment? (Is it needed?)

MaggieAppleton (changes requested):
Pointed out some minor edits and spots to clean up, but overall this looks great! Nearly good to go

src/components/app/header/index.tsx
NicollGuarnizo: My bad, I have fixed it

src/components/app/header/index.tsx
NicollGuarnizo: Change done! I have renamed it as `contentSectionLinks`

src/components/app/header/index.tsx
NicollGuarnizo: Yes, it's needed. Typescript throws a warning that the flyout menu it's too complex too represent, with that comment we are ignoring the warning.

MaggieAppleton (approved):
All looks good!

PR #758, opened 2021-07-08
tweak cueplayer for initial release

joelhooks:

### DIFF
--- a/src/components/player/cue-bar.tsx
+++ b/src/components/player/cue-bar.tsx
@@ -107,19 +107,23 @@ const NoteCue: React.FC<any> = ({
 }) => {
   const {setPlayerPrefs, getPlayerPrefs} = useEggheadPlayerPrefs()
   const [visible, setVisible] = React.useState(false)
+  const [clickedOpen, setClickedOpen] = React.useState(false)
   const {muteNotes} = getPlayerPrefs()
 
   useCue(cue, actions)
 
-  const open = () => {
+  const clickOpen = () => {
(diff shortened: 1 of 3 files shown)


src/components/player/cue-bar.tsx
```diff
@@ -107,19 +107,23 @@ const NoteCue: React.FC<any> = ({
...
 
   useCue(cue, actions)
 
-  const open = () => {
+  const clickOpen = () => {
     setVisible(true)
+    setClickedOpen(true)
     // if we seek to the correct time, the note is displayed
     actions.seek(cue.startTime)
+    actions.pause()
```
joelhooks, line 120: playing with it, I wanted it to pause when I explicitly click a cue so I can read at my leisure.

also important to note when the cue is open because of a CLICK or because of auto display,

src/components/player/cue-bar.tsx
```diff
@@ -129,8 +133,17 @@ const NoteCue: React.FC<any> = ({
 
   React.useEffect(() => {
     const isVisible = !muteNotes && cueActive && !seeking && playerReadyEnough
-    setVisible(isVisible)
-  }, [setPlayerPrefs, cueActive, seeking, muteNotes, playerReadyEnough])
+    if (!clickedOpen) {
```
joelhooks, line 136: only run this if the cue wasn't clicked open, otherwise the logic stands. this prevents the "muted notes NEVER display even when I click them" issue @lsminter reported

PR #759, opened 2021-07-12
Add Note input

DrShpongle:
What was done:
1) Simplified `OverlayWrapper` behavior (thanks to sudden discovery - "minimal aspect ratio height" https://codepen.io/trys/pen/XWryxoy)
2) Add Note overlay with form. I have created the new player state `addingNote` and wire form appearance to it. Does it make sense?
3) Applied accessibility attributes to the form elements.

https://share.getcloudapp.com/GGuW9n1O


src/components/player/player-sidebar.tsx
```diff
@@ -88,55 +89,76 @@ const NotesTab: React.FC = () => {
...
+            </div>
+          </SimpleBar>
         </div>
-      </SimpleBar>
+        <div className="flex-shrink-0 p-4">
+          <button
+            type="button"
+            onClick={(e) => {
+              e.preventDefault()
+              sendPlayerState('ADD_NOTE')
```
joelhooks, line 149: instead of passing `send` this way, we should have a callback like `onAddNote` that is passed in so that the parent etc can pass in a callback and do `send` at that level instead of passing in a more powerful `send` function like this.

so here it would be 

```
if(isFunction(onAddNote){
onAddNote()
}
```

or `onAddNote` could default to `noop` so it'd always be a function.

src/pages/lessons/[slug].tsx
```diff
@@ -561,9 +577,14 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
                   <RecommendNextStepOverlay lesson={lesson} />
                 </OverlayWrapper>
               )}
+              {playerState.matches('addingNote') && (
+                <OverlayWrapper>
+                  <AddNoteOverlay sendPlayerState={send} />
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:582]:
```
<AddNoteOverlay onAddNote={() => send('ADD_NOTE')} />
```

Make sense? I like this more because now we aren't sending in all the power of state send method and handling the logical concerns of what happens when we click add note here instead of internally in the sidebar.

src/pages/lessons/[slug].tsx
```diff
@@ -561,9 +577,14 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
...
+                </OverlayWrapper>
+              )}
             </div>
             <div className="lg:col-span-3 side-bar">
-              <PlayerSidebar videoResource={lesson} />
+              <PlayerSidebar videoResource={lesson} sendPlayerState={send} />
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:587]:
```
<PlayerSideBar videoResource={lesson} onAddNote={() => send('ADD_NOTE')} />
```

Similar idea here.

src/components/pages/lessons/overlay/add-note-overlay.tsx
```diff
@@ -0,0 +1,93 @@
...
+  }, [])
+  return (
+    <div
+      className="w-[34rem] h-[20rem] rounded-md bg-white p-4 flex flex-col"
+      id="add-note-overlay"
+    >
+      <div className="flex-shrink-0 flex justify-end">
+        <button
+          type="button"
+          onClick={() => sendPlayerState('VIEW')}
```
joelhooks, line 23: this might be `onClose` versus `onAddNote` described below, but idea is the same. Send in a click handler function, handle click logic in player level to change state there not passing xstate around.

src/pages/lessons/[slug].tsx
Comment (DrShpongle) [reply src/pages/lessons/[slug].tsx:582]: Yeah, it makes sense 👍

Comment (DrShpongle) [reply src/pages/lessons/[slug].tsx:582]: Yeah, it makes sense 👍

PR #763, opened 2021-07-13
pl/Move dark mode toggle from navbar to footer

ParkerGits:
[Notes/reasons on moving the dark mode toggle here.](https://roamresearch.com/#/app/egghead/page/NFSVQUOU4)

New Header
new footer

I put the dark mode toggle on the same level as the copyright as is done on [the Vercel Dashboard](https://vercel.com/dashboard) **(bottom right of footer)**. Perhaps we should consider making it more visible/descriptive as they have done with the `<select>` element.

Vercel Dashboard

giphy

### DIFF
--- a/src/components/app/footer.tsx
+++ b/src/components/app/footer.tsx
@@ -144,52 +144,60 @@ const DarkModeToggle = () => {
   const {subscriber, cioIdentify} = useCio()
   const {theme, setTheme} = useTheme()
   React.useEffect(() => setMounted(true), [])
-
+  const handleClick = () => {
+    const nextTheme = theme === 'dark' ? 'light' : 'dark'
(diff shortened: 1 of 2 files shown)


MaggieAppleton:
I agree making it a more explicit `select` menu or a toggle element would help communicate to users that it's a clickable element. At the moment it could be mistaken for a static icon.

Adding text that says either "dark mode" or "light mode" next to the icon will help.
I think creating a toggle element like this communicates the fact it can be switched between the two modes:

image

ParkerGits:
new toggle button
light mode
dark mode
Open to feedback!

MaggieAppleton (approved):
This looks great 👍

PR #764, opened 2021-07-13
fixes single term as topic for search

joelhooks:
when typing a single term like `react` or navigating to the search page with a query like `q=react` it should load the associated topic page and not just search results.

this does not affect the loading of instructor pages, but those queries are much less frequent and generally more complex to load this way.

### DIFF
--- a/src/components/search/index.tsx
+++ b/src/components/search/index.tsx
@@ -10,7 +10,7 @@ import {
   ClearRefinements,
   ScrollTo,
 } from 'react-instantsearch-dom'
-import {get, isEqual, isEmpty} from 'lodash'
+import {get, isEqual, isEmpty, first} from 'lodash'
 import {useToggle, useClickAway} from 'react-use'
 
 import config from 'lib/config'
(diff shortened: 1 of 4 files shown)


src/components/search/index.tsx
```diff
@@ -52,16 +52,10 @@ const Search: FunctionComponent<SearchProps> = ({
   }
 
   const noTopicsSelected = (searchState: any) => {
-    return get(searchState, 'refinementList._tags', []).length === 0
-  }
-
-  const onlyTheseTagsSelected = (tags: string[], searchState: any) => {
```
joelhooks, line 58: this was unused

src/pages/q/[[...all]].tsx
```diff
@@ -121,7 +123,7 @@ const SearchIndex: any = ({
       const href: string = createUrl(searchState)
       setNoIndex(queryParamsPresent(href))
 
-      router.push(`/q/[[all]]`, href, {
+      router.push(href, undefined, {
```
Comment (joelhooks) [inline src/pages/q/[[...all]].tsx:126]: this has been broken for quite some time. probably since we upgraded next.

vojtaholik (approved):

PR #766, opened 2021-07-14
pl/Fix dark mode button switch position bug 

ParkerGits:
wrong position
For default dark mode users, the dark mode button was starting in the wrong position. This small change updates the position of the button switch once the component is mounted, thus fixing the problem.
fixed
bug squashing


ParkerGits:
this is a small fix but I really don't want to be the intern who force pushes to production 😄

MaggieAppleton (approved):
nice catch ⚾️

PR #767, opened 2021-07-14
do not apply search query as selected topic/tag facet

joelhooks:

### DIFF
--- a/src/utils/search/topic-extractor.ts
+++ b/src/utils/search/topic-extractor.ts
@@ -1,0 +1,18 @@
+import {SearchState} from 'react-instantsearch-core'
+import {first} from 'lodash'
+
+export const topicExtractor = (searchState: SearchState) => {
+  const selectedTopics = searchState.refinementList?._tags || []
+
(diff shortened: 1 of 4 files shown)


src/utils/search/topic-extractor.ts
```diff
@@ -1,8 +1,11 @@
 import {SearchState} from 'react-instantsearch-core'
-import {first} from 'lodash'
+import {first, isEmpty} from 'lodash'
 
 export const topicExtractor = (searchState: SearchState) => {
-  const selectedTopics = searchState.refinementList?._tags || []
```
joelhooks, line 5: `_tags` is always an empty array so what this was doing was using that array and pushing the query topic onto it. Now if the `tags` array has stuff in it we just return it because we don't want to mess with the "query as a topic" and instead just move on. If no tags are explicitly defined we push the single query onto the "selectedTopics" array to display that topic landing page

MaggieAppleton (approved):
works beautifully

Issue #768, opened 2021-07-14
Set search filter menu to visible by default


MaggieAppleton:
We have a very useful and powerful filter menu that learners can use to find relevant content on the site. Sadly, many learners don't use it to find relevant content. This is partially due to the fact we hide the filter menu by default.

Standard view for a search term (filter menu is closed):

![image](https://user-images.githubusercontent.com/5599295/125659547-3204f376-8082-4547-a6a0-d230aaf88b60.png)

If we look at our mixpanel data, we can see out of all searches only 10% of learners make use of search filters:

![image](https://user-images.githubusercontent.com/5599295/125659611-9a14512d-70d8-4059-b821-e1c8c5d0ee6f.png)

![image](https://user-images.githubusercontent.com/5599295/125659598-2e6e6389-3e5a-42eb-9769-2901d29e34d9.png)

It's possible learners simply don't need filtering to find what they're looking for.

However, we've gotten feedback from learners that indicate they don't know the filtering functions exist.
Relevant feedback from learners:
- "it'd be lovely if there was a way to filter courses that are free"
- "It's hard the experience to look for courses, ie free courses or "hot" courses"
- "The complete lack of sorting/filtering ability on almost every page makes it feel impossible to find anything."
- "I went looking for something to help me do fast complex webgl in react. But searching is a nightmare"

The simple solution here is to **have the filter menu open by default** when a learner first uses the search bar. The learner can choose to close the filter menu if they feel it's getting in the way, but at least they know it exists so they can use it in the future.

Once a learner closes the filter menu, it should stay closed until they choose to open it again. The state should persist across sessions.

At the moment, the filter menu also **disappears every time a learner selects a tag in the menu**. This can be disorienting as the interface unexpectedly jumps while the page refreshes. It also makes it difficult to quickly select multiple filters. The filter controls disappear the instant you try to select them.

Current behaviour:

![125483784-dec93456-aee2-4c66-911d-f50c80638136](https://user-images.githubusercontent.com/5599295/125660309-50439913-b73c-4ca1-b1bc-329b1fe9fe87.gif)

As a general rule, the learner should be in control of the open/close state of the menu. It shouldn't automatically close without them choosing to close it.

MaggieAppleton:
Additional note: this is a stop-gap solution for the moment.

We're currently exploring the possibility of moving the filters to a left-hand sidebar in the future so that they're always visible. TBD

[...]

PR #769, opened 2021-07-14
pl/set search filter menu visible by default and have it persist across sessions

ParkerGits:
( #768 )
The filter is now open by default and the state of the filter is stored in localStorage so that it persists across sessions. Filter no longer toggles off on click away. 

searching

### DIFF
--- a/src/components/search/index.tsx
+++ b/src/components/search/index.tsx
@@ -120,11 +120,26 @@ const Search: FunctionComponent<SearchProps> = ({
         {...rest}
       >
         <Configure hitsPerPage={config.searchResultCount} />
+        <ScrollTo scrollOn="query" />
         <div className="sm:pb-16 pb-8 space-y-8 bg-gray-50 dark:bg-gray-900 -mx-5">
-          <div
-            className="max-w-screen-xl md:-mt-5 -mt-3 pt-5 mx-auto"
-            ref={refinementRef}
-          >
+          {!isEmpty(instructor) && (
+            <div className="mb-10 pb-8 xl:px-0 px-5 mx-auto dark:bg-gray-900">
+              {shouldDisplayLandingPageForInstructor(instructor.slug) && (
+                <InstructorCuratedPage instructor={instructor} />
+              )}
+            </div>
+          )}
+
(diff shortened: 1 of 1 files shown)


ParkerGits:
filter

joelhooks:
I feel pretty strongly that this bar needs to move below the topic page content if this is gonna be open by default because it is hideous.

ParkerGits:
I moved the search bar below the topic page content, but felt unhappy about this functionality:
no scrollto

Setting the page to scroll to the top of the topic page content on query seems like a decent remedy:
with scrollto

Having to scroll to the bottom of, say, the React topic page content after searching "react" doesn't seem too bad. Could it be more clear that the search bar has moved down?

joelhooks:
I think it's "kind of janky" but less so than having that dropdown open at the top of our lovely Topic Pages before the content.


What I'd love to see is the filters moved to a side column next to the search results while keeping the bar prominently up at the top of the page and "solid like granite" in terms of its positioning. Moving user inputs around after they finish typing "react" is not great!


Here's how it looks on egghead-rails:





Feels like we could get that simple layout done relatively quickly and maybe skip this decision about moving inputs around the page while opening up the filters so learners know they have that power without needing to dig into a buried UI.

MaggieAppleton:
Making the filters always visible in a side column is definitely the next step. It requires a non-trivial redesign of the search page though.

Having the filter menu be open by default is not ideal, but it was meant as a temporary solution while we design and implement a more robust solution.

If we think having the topic menu open at the top of the page is too invasive, I vote we skip this temporary fix, and wait to fix the problem in a more cohesive way. I'm writing specs for a better solution today, and will have a pitch written up before Wednesday.

dealingwith:
If we need to abandon this specific direction, this PR should be closed

MaggieAppleton:
Yup – to follow up on this, we've decided to step back and holistically redesign the search functionality rather than implement a temporary patch to the filtering issue.

Thanks for working on this anyway @ParkerGits! Appreciate the effort. We'll make sure to get everyone on the same page before scoping out the next round of improvements to search.

PR #780, opened 2021-07-26
State Management curated page

laurosilvacom:
## [Live Preview](https://egghead-io-nextjs-git-ls-state-management-eggheadio1.vercel.app/learn/state-management)

CleanShot 2021-07-26 at 08 43 01@2x


src/pages/learn/state-management/index.tsx
```diff
@@ -0,0 +1,200 @@
...
+    title,
+  	description,
+  	"name": subTitle
+  },
+  'theBigIdeas': resources[slug.current == 'the-big-ideas'][0]{
+    description,
+    title,
+    resources[] {
+    "name": subTitle,
+      ...
```
zacjones93, line 49: oh nice, didn't know you could spread like this

src/pages/learn/state-management/index.tsx
```diff
@@ -0,0 +1,200 @@
...
+      </div>
+    </div>
+  )
+}
+
+type JumbotronProps = {
+  resource: CardResource
+}
+
+const Jumbotron: FunctionComponent<JumbotronProps> = ({resource}) => {
```
zacjones93, line 139: Is there a reason we define another Jumbotron here vs using the on in `/components`?

src/pages/learn/state-management/index.tsx
```diff
@@ -0,0 +1,200 @@
...
+  const data = await sanityClient.fetch(StateManagementQuery)
+
+  return {
+    props: {
+      data,
+    },
+  }
+}
+
+const RecoilSection = ({resource}: any) => {
```
zacjones93, line 70: These sections are great candidates to promote to a shared components folder

zacjones93 (approved):
This looks like a really good start to the state management page. Lots of room to build on!

The recoil section could use some gardening on telling us what problems it solves that React isn't good at but that can be done on the content side of things

src/pages/learn/state-management/index.tsx
laurosilvacom: Had a couple of issues passing the appropriate values. But that's definitely something to refactor the next time we work on this page.

PR #785, opened 2021-07-31
cue notes: render code blocks

vojtaholik:

PR #792, opened 2021-08-03
re-added vojtas code to fix code blocks

### LINKED ISSUE #793: Notes aren't displaying code blocks.
The notes are not displaying code blocks in the same TimeStamp'ed block that has text in it. 

Here is the raw markdown.
Screen Shot 2021-08-03 at 2 59 53 PM

Here is what is getting displayed. 
Screen Shot 2021-08-03 at 2 59 26 PM

Vojta addressed this issue in [this PR](https://github.com/eggheadio/egghead-next/pull/785) but due to notes causing lessons to crash. It turns out that the linked PR did not cause the crash so the changes that Vojta made needs to be re-implemented.

lsminter:
I re-added Vojta's code from this PR https://github.com/eggheadio/egghead-next/pull/785. 

This PR fixes #793.

The code blocks are now working again. 

Screen Shot 2021-08-03 at 3 30 08 PM

### DIFF
--- a/src/pages/api/__tests__/github-load-notes.test.ts
+++ b/src/pages/api/__tests__/github-load-notes.test.ts
@@ -11,5 +11,37 @@ This is a **note**.
 `
   const parsedNotes = parseMdxNotesFile(markdownNotes)
 
-  await expect(parsedNotes).resolves.toEqual('not the answer')
+  await expect(parsedNotes).resolves.toEqual(`WEBVTT
+
+note
+00:00:44.000 --> 00:00:50.000
+This is a **note**.
+
+`)
+})
+
+test('properly parses markdown codeblock and text in the same timestamp', async () => {
+  const markdownNotes = `
+<TimeStamp start="0:44" end="0:50">
+
+This is a **note**.
+
+\`\`\`js
+This is a code block!
+\`\`\`
+
+</TimeStamp>
+`
+
+  const parsedNotes = parseMdxNotesFile(markdownNotes)
+  await expect(parsedNotes).resolves.toBe(`WEBVTT
+
+note
+00:00:44.000 --> 00:00:50.000
+This is a **note**.
+ \`\`\`js
+This is a code block!
+\`\`\`
+
+`)
 })
--- a/src/pages/api/__tests__/github-load-notes.test.ts
+++ b/src/pages/api/__tests__/github-load-notes.test.ts
@@ -1,0 +1,15 @@
+import {parseMdxNotesFile} from '../github-load-notes'
+
+// this had to use an async/await because the mdx compiler
+// is set up to be async and return a promise
+test('properly parses a markdown text string', async () => {
+  const markdownNotes = `<TimeStamp start="0:44" end="0:50">
+
+This is a **note**.
+
+</TimeStamp>
+`
+  const parsedNotes = parseMdxNotesFile(markdownNotes)
+
+  await expect(parsedNotes).resolves.toEqual('not the answer')
+})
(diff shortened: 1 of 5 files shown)


src/pages/api/github-load-notes.ts
```diff
@@ -64,7 +64,9 @@ export async function loadNotesFromUrl(url: string) {
         }
       }, {})
       note.type = 'root'
-      const contents = toMarkdown(note)
+      const contents = note.children
+        .map((node: any) => toMarkdown(node))
+        .join(' ')
```
lsminter, line 69: Like Vojta said, only first node here was being recognised.

joelhooks: @lsminter, I'd like to see a test that covers this circumstance. can you add that please.

src/pages/api/__tests__/github-load-notes.test.ts
```diff
@@ -0,0 +1,15 @@
...
+test('properly parses a markdown text string', async () => {
+  const markdownNotes = `<TimeStamp start="0:44" end="0:50">
+
+This is a **note**.
+
+</TimeStamp>
+`
+  const parsedNotes = parseMdxNotesFile(markdownNotes)
+
+  await expect(parsedNotes).resolves.toEqual('not the answer')
```
joelhooks, line 14: because our `parseMdXNotesFile` method returns a `Promise` we can use await so that our test can understand to wait for the promise to resolve before checking the value

src/pages/api/__tests__/github-load-notes.test.ts
```diff
@@ -0,0 +1,15 @@
+import {parseMdxNotesFile} from '../github-load-notes'
+
+// this had to use an async/await because the mdx compiler
+// is set up to be async and return a promise
+test('properly parses a markdown text string', async () => {
+  const markdownNotes = `<TimeStamp start="0:44" end="0:50">
```
joelhooks, line 6: our markdown files are "just strings" so you can simulate conditions here and check the results below"

src/pages/api/github-load-notes.ts
```diff
@@ -17,11 +17,7 @@ const loadGithubNotes = async (req: NextApiRequest, res: NextApiResponse) => {
 
 export default loadGithubNotes
 
-export async function loadNotesFromUrl(url: string) {
-  const result = await fetch(url)
-
-  const text = await result.text()
-
+export function parseMdxNotesFile(text: string) {
```
joelhooks, line 20: made a new function here to make it simpler to test.

src/pages/api/github-load-notes.ts
```diff
@@ -88,3 +86,11 @@ ${note.text}\n\n`
     return vtt
   })
 }
+
+export async function loadNotesFromUrl(url: string) {
+  const result = await fetch(url)
+
+  const text = await result.text()
+
+  return parseMdxNotesFile(text)
```
joelhooks, line 95: now we are just calling the new method and return the value. This doesn't functionally change our code, just means we can test without worrying about loading a note from a URL.

we KNOW `fetch` works and testing that doesn't provide value to our local system.

there are reasons to test this sort of thing, but for now, we can focus on our implementation and not be concerned with testing `fetch` integration.

we'd say `parseMdxNotesFile` has been _isolated for testing_.

src/pages/api/__tests__/github-load-notes.test.ts
```diff
@@ -11,5 +11,37 @@ This is a **note**.
...
+<TimeStamp start="0:44" end="0:50">
+
+This is a **note**.
+
+\`\`\`js
+This is a code block!
+\`\`\`
+
+</TimeStamp>
+`
```
lsminter, line 34: The issue was having text then a code block in the same TimeStamp. This tests that issue.

joelhooks (approved):

PR #798, opened 2021-08-04
Portfolio Foundations email sequence kickoff

tayiorbeii:
This PR implements the kickoff of the Portfolio Foundations email sequence, and landing page copy to encourage signup at `/portfolio-foundations`

foundation


joelhooks:
I'd consider removing the "project club" cta since it is a dead-end (perhaps 302 forwarding the route entirely?) and replacing it with this (or just replacing the language and destination)

The previous landing page still exists as well, and should likely be forwarded (or otherwise updated accordingly)

https://github.com/eggheadio/egghead-next/blob/main/src/components/app/header/index.tsx#L54-L56

I'd also consider `developer-portfolio-foundations` as the URL for just a little uptick in SEO differentiation.

tayiorbeii:
> I'd consider removing the "project club" cta since it is a dead-end (perhaps 302 forwarding the route entirely?) and replacing it with this (or just replacing the language and destination)

@joelhooks  Created a new "Portfolio Foundations" CTA and swapped it out

> The previous landing page still exists as well, and should likely be forwarded (or otherwise updated accordingly)

In the last email of the Portfolio Foundations sequence, we link to the Club landing page since building out a portfolio project is the next step after building the portfolio, and the club is a great way to do it.


> I'd also consider `developer-portfolio-foundations` as the URL for just a little uptick in SEO differentiation.

Got this changed!

joelhooks:
@tayiorbeii we might need to make the logic more robust or add a time delay to how that switch statement works so they don't drop directly into portfolio club. It could look at CK or CIO properties to determine when is the right time to show the right CTA if you want to get slick with it.

tayiorbeii:
> @tayiorbeii we might need to make the logic more robust or add a time delay to how that switch statement works so they don't drop directly into portfolio club. It could look at CK or CIO properties to determine when is the right time to show the right CTA if you want to get slick with it.

Makes sense-- Removed the call to signup for a club from the last email, and left in the "let us know what your next step will be"

MaggieAppleton (approved):
Looks good 👍

jb-gf:
You think you are right about the future of the most popular connection 📶 [yeah last night @th toll]()

PR #799, opened 2021-08-05
Fix Notes tab scrolling

DrShpongle:
I did some rework related to scroll inside Notes tab inside player sidebar.

Now:

1) it doesn't fail when enter cue with Lessons tab active:
01-on-lessons-tab

2) it doesn't fail when new lesson starts with Notes tab active:
02-on-notes-tab

3) Autoscroll to active note works well:
03-autoscroll-on-play

The thing I was not able to figure out is that scrolling to active note on click on some cue works after second click only:
04-scrolls-on-double-click

### DIFF
--- a/src/components/player/cue-bar.tsx
+++ b/src/components/player/cue-bar.tsx
@@ -111,26 +111,6 @@ const NoteCue: React.FC<any> = ({
   const [clickedOpen, setClickedOpen] = React.useState(false)
   const {muteNotes, activeSidebarTab} = getPlayerPrefs()
 
-  useCue(cue, actions)
-
-  const clickOpen = () => {
-    if (activeSidebarTab === 1) {
-      scrollToActiveNote()
-    }
-    setVisible(true)
-    setClickedOpen(true)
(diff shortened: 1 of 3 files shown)


src/components/player/cue-bar.tsx
```diff
@@ -120,7 +129,10 @@ const NoteCue: React.FC<any> = ({
     actions.seek(cue.startTime)
     actions.pause()
     track('opened cue', {cue: cue.text})
-    !muteNotes && setPlayerPrefs({activeSidebarTab: 1})
+    // !muteNotes && setPlayerPrefs({activeSidebarTab: 1})
```
DrShpongle, line 132: commented this line out because it conflicts with the check I do (`if (activeSidebarTab === 1)`)

joelhooks: the idea here is that we are "very forward" with the notes and autodisplay in the sidebar if they aren't muted, but if that needs to be removed that is fine too

joelhooks (approved):
looks great! 🌮

PR #812, opened 2021-08-11
Add Purchase and Play polling to Thank You page

jbranchaud:
This PR adds _Purchase and Play_ functionality to egghead-next.

When a visitor completes the Stripe checkout flow for a pro membership, they will be immediately logged into egghead so that they can view paid content.

This is achieved by:
- Polling the backend server for an auth token with their Stripe Checkout Session ID
- Once an auth token is received, refresh the viewer context

Edges that are handled by the polling:
- A user is already signed in, don't poll.
- The checkout session ID is missing, don't poll.
- An auth token is received from the backend, stop polling.
- We've unsuccessfully polled 5 times, stop polling.

Here is what this looks like from the user's perspective when they are redirect from the Stripe Checkout Session to the Thank You/Membership page:

_(Note: this first video is outdated, keep scrolling to the second video to see the latest and greatest.)_

https://user-images.githubusercontent.com/694063/129088642-ca9a70b5-6a46-4b0f-a633-8a51ea6a593d.mp4

Room for improvement:

The Thank You page was originally built to handle two cases, which this PR doesn't seek to alter. Either:
- (A) you were already signed in, you can start viewing now or
- (B) you weren't signed in, you need to head over to your inbox and find the Email Confirmation link.

With the polling happening quickly in the background, what happens is that you see the (B) screen momentarily and then it flashes over to the (A) screen. This is Good Enough™ for a first pass. A second pass would involve a bit of UX cleanup: show a loading button that turns into 'Explore the egghead Catalogue' (once the auth token is received) or directs them to their inbox if we can't find it.

## Update

I've taken another pass at this to better handle the situation described above in _Room for Improvement_. The component now has a handle on the different states it can be in:
1. Already authenticated
2. In the process of fetching a fresh auth token
3. An auth token has been fetched
4. Unable to fetch an auth token

We can tailor the UI and the copy to these scenarios. I've taken a first stab at that, but I think this could now use a more thorough and skilled design pass. The copy needs work as well.

https://user-images.githubusercontent.com/694063/129979486-dc21498d-2e35-4e87-8d98-d4543b98cca2.mp4

---


src/components/feedback-input/index.tsx
```diff
@@ -4,7 +4,7 @@ import axios from 'utils/configured-axios'
 import * as Yup from 'yup'
 import isEmpty from 'lodash/isEmpty'
 import {motion} from 'framer-motion'
-import {useInterval, useMedia} from 'react-use'
+import {useInterval} from 'react-use'
```
jbranchaud, line 7: Unrelated cleanup of unused import.

src/context/viewer-context.tsx
```diff
@@ -165,14 +168,26 @@ function useAuthedViewer() {
...
+  const handleAccessTokenAuthentication = async (
+    authToken: string,
+    expiresInSeconds: string,
+  ) => {
+    const authenticatedUser = await auth.handleAccessTokenAuthentication(
+      authToken,
+      expiresInSeconds,
+    )
+    setViewer(authenticatedUser)
+  }
```
jbranchaud, line 180: This wraps `Auth`'s `handleAccessTokenAuthentication` so that the user data it returns gets set as the viewer in the viewer context. This was necessary to get both the main view and the header to recognized the user corresponding to the received auth token.

jbranchaud:
I'm putting a `NO MERGE` on this until the server-side implementation is deployed.

src/components/pages/confirm/membership/index.tsx
```diff
@@ -185,28 +187,64 @@ export const ConfirmMembership: React.FC<ConfirmMembershipProps> = ({
         </>
       ) : (
         <>
+          {currentState.matches('pending') && <p>Setting up your account...</p>}
```
jbranchaud, line 190: Requesting ideas for how to style and word this.

The idea is that while the client polls in the background for an auth token from the server, this page will display a message explaining that we are in the process of setting up their account.

src/components/pages/confirm/membership/index.tsx
```diff
@@ -185,28 +187,64 @@ export const ConfirmMembership: React.FC<ConfirmMembershipProps> = ({
...
-                  <strong className="dark:text-yellow-300 text-rose-500">
-                    ${session.amount} for an egghead membership
-                  </strong>{' '}
-                  and sent an email to{' '}
-                  <strong className="dark:text-yellow-300 text-rose-500">
-                    {session.email}
-                  </strong>{' '}
-                  so you can log in and access your membership.
-                </p>
+                {currentState.matches('pollingExpired') && (
```
jbranchaud, line 195: This is the fallback messaging. If we are unable to request an auth token for them, then they will just have to do it the old fashioned way via the emailed magic link.

src/components/pages/confirm/membership/index.tsx
```diff
@@ -185,28 +187,64 @@ export const ConfirmMembership: React.FC<ConfirmMembershipProps> = ({
...
+                      </strong>{' '}
+                      and sent an email to{' '}
+                      <strong className="dark:text-yellow-300 text-rose-500">
+                        {session.email}
+                      </strong>{' '}
+                      so you can log in and access your membership.
+                    </p>
+                  </>
+                )}
+                {currentState.matches('authTokenRetrieved') && (
```
jbranchaud, line 217: This is the happy path. We retrieved and stored their auth token, they are now ready to immediately view paid content 🎉

src/machines/auth-token-polling-machine.ts
```diff
@@ -0,0 +1,100 @@
...
+            invoke: {
+              src: 'requestAuthToken',
+              onDone: {
+                target: 'verifyRetrieval',
+                actions: ['assignAuthToken'],
+              },
+              onError: {
+                target: 'scheduleNextPoll',
+              },
+            },
```
jbranchaud, line 38: This is machine doesn't get sent any events. Instead, it starts in the `{pending: "polling"}` state which invokes the `requestAuthToken` service.

From there it will either:
- `onError` --> schedule the next poll after 2000ms, or transition to a final failed state.
- `onDone` --> check the auth token was set, apply any more delay to keep UI smooth, and then transition to a final state.

dealingwith:
@vojtaholik is going to do a design pass

Also, this branch now needs to be rebased or otherwise caught up to main.

joelhooks:
kodiak merges with main (barring conflicts) efficiently

vojtaholik:
updated designs:
updated designs

src/components/pages/confirm/membership/index.tsx
```diff
@@ -39,10 +44,12 @@ const PrimaryMessage: React.FC = ({children}) => {
   return <div className="dark:text-gray-200 text-gray-800">{children}</div>
 }
 
+const tweet = `https://twitter.com/intent/tweet/?text=Just joined @eggheadio to level up my development skills.`
```
vojtaholik, line 47: I added a little cta to tweet about egghead. @dealingwith copy suggestions welcome!

jbranchaud:
> updated designs:
> updated designs

This is so good! It's just how I was hoping it would end up. 🎉  \cc @vojtaholik

src/components/pages/confirm/membership/index.tsx
```diff
@@ -141,36 +159,41 @@ const LastResource = () => {
...
       {children}
     </div>
   )
 }
 
+const StartLearning: React.FC = () => {
+  return (
+    <Link href="/q">
+      <a className="text-white bg-blue-500 border-0 py-3 px-5 rounded-md hover:bg-blue-600">
+        Browse Catalogue
```
joelhooks, line 172:
```suggestion
        Browse Catalog
```
🇺🇸

joelhooks (approved):
can we extract this? I'd like to get it implemented for partner products! https://github.com/skillrecordings/turbo-playground

Issue #808, opened 2021-08-10
Add authors and publish dates to /articles landing page


MaggieAppleton:
Our current articles landing page (located at  `/blog` - a naming inconsistency issue we should resolve in a separate issue) currently only shows the cover image and title of an article.

![image](https://user-images.githubusercontent.com/5599295/128875775-50dc0d33-9ebb-4b51-8e52-9a88983a3a72.png)

To give readers more context about what they're clicking on, and authors more credit for their work we should visibly add:

- Author name + photo
- Publish date stated relative to current date (yesterday, 2 weeks ago).
	- If it's more than a month old then switch to regular dates: March 20th, 2021. If that turns out to be complex to implement then regular dates are fine.
- 1-2 sentence description. Should be a simple one-liner pulled from Sanity.

At the moment the article cards are formatted on the main articles page: https://github.com/eggheadio/egghead-next/blob/main/src/pages/blog/index.tsx

We could consider pulling them out into their own card component **but we do not need to go down the road of designing an ultimate, universal card component here.** That is a lovely, meaty, future goal that will derail us instead of iteratively improving the articles page today.

### Revised individual card layout with author, date, description:

![image](https://user-images.githubusercontent.com/5599295/128881206-30b3f0c5-45a5-40a6-a1f3-a62c383a2db1.png)

**If there's no description available, only display author and publish date**

![image](https://user-images.githubusercontent.com/5599295/128881276-71b5c592-2baa-43b8-8433-d7b1c07a7462.png)

### Whole page mockup
Page layout can stay the same, we're just adding metadata below the title:
![image](https://user-images.githubusercontent.com/5599295/128881149-08a1d38b-87fa-4500-ba7e-f5148c21d8ec.png)

---

⚡️ **Figma files with design mockups: https://www.figma.com/file/cWvKOJ0XyRte8kyP3LvBAT/Talks-Articles-Podcasts?node-id=101%3A764** ⚡️

🙋‍♀️ Feedback welcome. Are there any technical limitations that would block this work, or design considerations I've overlooked?

laurosilvacom:
This makes sense! 

I think @NicollGuarnizo would be a great candidate to work on this. She's doing similar work for Escuela Frontend.

NicollGuarnizo:
Yeah, I'll work on this :)

NicollGuarnizo:
**This is how it's looking so far:** 

[...]

PR #829, opened 2021-08-16
Adding authors and publish dates to /articles landing page

NicollGuarnizo:
This PR is related to this [issue](https://github.com/eggheadio/egghead-next/issues/808)

image

I have created a new field in Sanity to add the description for each article, and has a rule of 150 max characters to avoid any visual noise, as it was discussed in the issue. 

**Questions to Think About:**
The only article that has a description was for my internal testing. 

- Should we review the articles already published to see which one deserves a description? or we just want to start using this new feature from now on. :thinking:


MaggieAppleton (approved):

MaggieAppleton:
This all looks good. I think we can ship it now even though we don't have any descriptions on articles.

I'll post to the slack gardening channel to see if we can get someone do do a pass through existing articles and add short descriptions

PR #831, opened 2021-08-17
Batch of Sorting Hat Reducer Unit Tests

ParkerGits:
Wrote tests for each action on the Sorting Hat Reducer. This required me to move the `cioIdentify` function to another file so that I could mock its implementation. Also needed to updated moduleDirectories in the Jest config to get the tests working. 

this testing newbie appreciates feedback 😎

### DIFF
--- a/src/components/survey/__tests__/sorting-hat-reducer.test.js
+++ b/src/components/survey/__tests__/sorting-hat-reducer.test.js
@@ -1,0 +1,219 @@
+/**
+ * @jest-environment jsdom
+ */
+
+// could eventually test that cioIdentify and track are being called appropriately
+
+import {cioIdentify} from '../cio-identify'
+import {sortingHatReducer, sortingHatInitialState} from '../sorting-hat-reducer'
+import {track} from 'utils/analytics'
+
+jest.mock('../cio-identify', () => ({
+  cioIdentify: jest.fn(() => null),
+}))
+
+jest.mock('utils/analytics', () => ({
+  track: jest.fn(),
+}))
+
+const defaultFirstQuestion =
+  sortingHatInitialState.data[sortingHatInitialState.currentQuestionKey]
+
+describe('empty action', () => {
+  test('should return the initial state', () => {
+    expect(sortingHatReducer(sortingHatInitialState, {})).toStrictEqual(
+      sortingHatInitialState,
+    )
+  })
+})
+
+describe('load action', () => {
+  test('should close sorting hat for load action with undefined subscriber after finished loading', () => {
+    expect(
+      sortingHatReducer(sortingHatInitialState, {
+        type: 'load',
+        loadingSubscriber: 'false',
+      }),
+    ).toStrictEqual({
+      ...sortingHatInitialState,
+      question: defaultFirstQuestion,
+      closed: true,
+    })
+  })
+
+  test('should add question property to state on load action while loading subscriber', () => {
+    expect(
+      sortingHatReducer(sortingHatInitialState, {
+        type: 'load',
+        loadingSubscriber: 'true',
+      }),
+    ).toStrictEqual({...sortingHatInitialState, question: defaultFirstQuestion})
+  })
+
+  test('should close survey for subscriber with survey complete', () => {
+    const subscriber = {
+      id: 'testCompleteSurvey',
+      email: 'test@egghead.io',
+      attributes: {sorting_hat_finished_at: 'today'},
+    }
+    expect(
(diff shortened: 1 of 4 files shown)


joelhooks:
I generally prefer to just add the tests next to the file instead of in a `__tests__` folder.

src/components/survey/sorting-hat-reducer.ts
```diff
@@ -2,6 +2,7 @@ import sortingHatData, {SurveyQuestion} from 'data/sorting-hat'
 import {CIOSubscriber} from 'hooks/use-cio'
 import {track} from 'utils/analytics'
 import {isEmpty} from 'lodash'
+import {cioIdentify} from './cio-identify'
```
joelhooks, line 5: consider moving this into top level "utils"

src/components/survey/__tests__/sorting-hat-reducer.test.js
```diff
@@ -0,0 +1,219 @@
...
+
+const defaultFirstQuestion =
+  sortingHatInitialState.data[sortingHatInitialState.currentQuestionKey]
+
+describe('empty action', () => {
+  test('should return the initial state', () => {
+    expect(sortingHatReducer(sortingHatInitialState, {})).toStrictEqual(
+      sortingHatInitialState,
+    )
+  })
```
joelhooks, line 27: why do you prefer `toStrictEqual` over `toEqual`?

One thing I like to do is explicitly name the item that is the subject of the expect.

```
test('should return the initial state', () => {
    const initialState = sortingHatReducer(sortingHatInitialState, {})
    
    expect(initialState).toStrictEqual(sortingHatInitialState)
  })
```

I also use whitespace to give the test additional readability.

Tests are an area where I feel readability aesthtics are **vital**.

src/components/survey/__tests__/sorting-hat-reducer.test.js
```diff
@@ -0,0 +1,219 @@
...
+})
+
+describe('load action', () => {
+  test('should close sorting hat for load action with undefined subscriber after finished loading', () => {
+    expect(
+      sortingHatReducer(sortingHatInitialState, {
```
joelhooks, line 33: this is similar to above, and I won' belabor it further, but this makes me unsure as to WHAT we are expecting

src/components/survey/__tests__/sorting-hat-reducer.test.js
```diff
@@ -0,0 +1,219 @@
...
+        type: 'load',
+        loadingSubscriber: 'false',
+      }),
+    ).toStrictEqual({
+      ...sortingHatInitialState,
+      question: defaultFirstQuestion,
+      closed: true,
```
joelhooks, line 40: and then here we are expecting it to strictly equal some object that gets formed inline.

What is this? Giving it a name and assigning it to a `const` would bring a lot of clarity here.

src/components/survey/__tests__/sorting-hat-reducer.test.js
```diff
@@ -0,0 +1,219 @@
...
+      subscriber: {
+        ...subscriber,
+        sorting_hat_started_at: Math.round(Date.now() / 1000),
+      },
+    })
+    expect(track).toHaveBeenCalledWith(
+      'answered survey question',
+      expect.anything(),
+    )
+    expect(track).toHaveBeenLastCalledWith('started survey', expect.anything())
```
joelhooks, line 126: 3 expects is 3 tests and should be treated as such, even if they have the same setup. You can use `beforeEach` in a `describe` for the block to reduce repetition.

src/components/survey/__tests__/sorting-hat-reducer.test.js
```diff
@@ -0,0 +1,219 @@
...
+    const finalQuestionState = sortingHatReducer(secondQuestionState, {
+      type: 'answered',
+      answer: 'best_practice',
+    })
+    const numCioIdentifyCallsBefore = cioIdentify.mock.calls.length
+    sortingHatReducer(finalQuestionState, {
+      type: 'closed',
+    })
+    const numCioIdentifyCallsAfter = cioIdentify.mock.calls.length
+    expect(numCioIdentifyCallsAfter).toBe(numCioIdentifyCallsBefore + 1)
```
joelhooks, line 188: some whitespace bewteen everything here would be welcome addition

joelhooks (approved):
my comments were related to style primarily and can be addressed as needed in a future PR

Issue #805, opened 2021-08-09
Add sorting to search results


MaggieAppleton:
There's currently no way to sort search results. This makes it harder for learners to find relevant content. Learners should be able to sort search results by last updated date, highest review score, and most popular.

We should add a 'sort by' dropdown menu in a bar directly above the search results with options for 'most popular,' 'highest rated,' and 'recently updated'. The default sorting value should be 'most popular'.

- **Most popular** equates to # of views on a course or lesson.
- **Highest rated** equates to highest average rating from learner reviews.
- **Recently updated** equates the last time the course was reviewed, updated, or released.

We should also display the number of search results for a given search above the results.

### Mockups

**Search result count on left and 'sort by' dropdown on right:**

![image](https://user-images.githubusercontent.com/5599295/128748144-4fdf8850-627a-48f4-8206-1b5601901045.png)

**Dropdown menu expanded:**
Current sorting selection should have bold text with a small blue checkmark to indicate it's active.
The order of the items in the dropdown menu shouldn't change.

![image](https://user-images.githubusercontent.com/5599295/128748174-9bd01d4a-a4e6-432c-82a2-f9f7112c0ec7.png)

**Sorting dropdown on a curated landing page**
Sorting dropdown appears below curated content and directly above results

![image](https://user-images.githubusercontent.com/5599295/128748831-ec16587b-ce4a-4bc2-858b-059bcec8349a.png)

**Sorting dropdown on mobile**
Place sorting dropdown directly beneath the search bar. Hide the search results number.

![image](https://user-images.githubusercontent.com/5599295/128751302-4573b790-f31e-4369-a776-02ce79d32d9e.png)

---

### Implementation Details

Dropdown menu can be built using Tailwind UI's dropdown component: https://tailwindui.com/components/application-ui/elements/dropdowns

Sorting should be doable with Algolia: https://www.algolia.com/doc/guides/managing-results/refine-results/sorting/

Unsure of the technical hurdles to get sorting to play nicely with filtering. Ideally, if someone has already selected filters (eg. "react" and "lessons"), the sorting feature should only sort results that are lessons tagged React. Changing the 'sort by' option should not reset filters.

Any details or considerations missing here?

joelhooks:
I think people also want to sort by "published" date which we can easily add to the search properties.

Another interesting field to sort on imo is duration. That speaks a bit to the "find something short to watch" and may surface lessons and such.

laurosilvacom:
[...]

PR #835, opened 2021-08-19
Tweak search results with SortBy algolia widget

DrShpongle:
This is a draft PR related to the https://github.com/eggheadio/egghead-next/issues/805

This is how it looks:
Resources for Web Developers in 2021 egghead io 2021-08-19 at 2 57 58 PM

Also fixed some weird spacings (paddings and negative margins) on general search layout and minor changes on CSS curated page (will update other curated pages if those changes accepted).

What about `SortBy` algolia widget we need to create related indices on Algolia side. Is there a way to have a test Algolia instance to play with in dev?


src/components/search/index.tsx
```diff
@@ -228,25 +174,111 @@ const Search: FunctionComponent<SearchProps> = ({
...
+                    <CuratedTopicPage topic={topic} />
+                  )}
+              </div>
+            )}
+
+            <div className="dark:bg-gray-900 bg-gray-50 mt-6">
+              <ScrollTo scrollOn="page" />
+
+              {/* SortBy widget goes below. Related indicies should be added on Algolia side before */}
+              {/* <div className="flex justify-between items-end pb-4 mb-6 border-b border-gray-200">
```
DrShpongle, line 219: commented out for now

laurosilvacom:
@DrShpongle We don't have applications per environment in Algolia. Would using the different staging indices be helpful?

CleanShot 2021-08-19 at 11 51 40@2x

joelhooks:
we definitely have algolia partitioned by environment. There is content_dev and content_prod

MaggieAppleton:
Anything we can do to move this forward? Are we stuck on the Algolia integration?

DrShpongle:
Yes, I have stuck with Algolia integration. If I understand correct, all `SortBy` widget does is switch the indices. So we have to create the indices for each of criteria (published, duration, popular, rated, updated), but I'm confused with how to rank it by these values and he we can pass for example rating or publish date to algolia.

DrShpongle:
closed in favor of https://github.com/eggheadio/egghead-next/pull/870

PR #839, opened 2021-08-24
Updating Sorting Hat tests according to feedback

ParkerGits:
@joelhooks gave me good feedback in #831 -- made those changes here

> why do you prefer `toStrictEqual` over `toEqual`?

It seems appropriate to expect that the returned state doesn't have any unexpected `undefined` properties. Though, I could see how `toStrictEqual` might not be worth the hit in readability/aesthetic.

### DIFF
--- a/src/components/survey/sorting-hat-reducer.test.js
+++ b/src/components/survey/sorting-hat-reducer.test.js
@@ -2,11 +2,11 @@
  * @jest-environment jsdom
  */
 
-import {cioIdentify} from '../cio-identify'
-import {sortingHatReducer, sortingHatInitialState} from '../sorting-hat-reducer'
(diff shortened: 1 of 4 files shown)


joelhooks:
@ParkerGits with a strict equal it’s more about specificity to me. is that much specificity needed? Maybe, but likely not where a simpler toEqual may be enough

src/components/survey/sorting-hat-reducer.test.js
```diff
@@ -0,0 +1,307 @@
+/**
+ * @jest-environment jsdom
```
joelhooks, line 2: this can be set in the co fog globally and to andy should be for this app

joelhooks (approved):

PR #844, opened 2021-08-25
add note control on player control bar

joelhooks:
this isn't perfect visually, but it at least isn't hidden by default and brings the context of note creation **into the player** more instead of buried to the side. 

Image 2021-08-24 at 6 30 03 PM

man taking notes

### DIFF
--- a/src/components/pages/lessons/overlay/add-note-overlay.tsx
+++ b/src/components/pages/lessons/overlay/add-note-overlay.tsx
@@ -6,7 +6,7 @@ import readingTime from 'reading-time'
 import {track} from '../../../../utils/analytics'
 
 const AddNoteOverlay: React.FC<{
-  onClose: any
+  onClose: (newNote: any) => void
   resourceId: string
   currentTime: number
 }> = ({onClose, resourceId, currentTime}) => {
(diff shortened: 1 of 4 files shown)


MaggieAppleton:
Tricky for find a good place for it, but a small pop-up button in the lower right with a comment icon and a clear label would be easier to see than having it in the video controls.

Either just the icon, or icon + label at 30-50% opacity would be more visible:

image

image

Ideally when users first open the page the button would be 100% opacity and the button would be extended to show the full label. A slightly pulse wouldn't hurt either. Once the button has made itself known it can transition to only being the icon in a smaller circle so it doesn't get in the way of the video.

On hover it should transition to full opacity and become a clickable button:

image

Figma: https://www.figma.com/file/eo7NVaUTiwNPbBwxLMSn2o/video-player?node-id=723%3A172

MaggieAppleton:
Slight alternative that moves it further to the side of the video. Dock the icon pop up to the edge and have it transition to full opacity and expand on hover.

image

image

joelhooks:
those look interesting. not sure about obstructing the video with UI, but definitely happy to see some further experiments.

Issue #842, opened 2021-08-24
Reimplement the "In Progress" section on the egghead homepage


MaggieAppleton:
Back in April of 2021 we implemented an "in progress" component that helped learners pick up where they left off.

It showed logged in users the last 3 courses they watched at the top of their home page.

Original pitch: #551 

![image](https://user-images.githubusercontent.com/5599295/130646602-2742f25d-1b64-4ef5-b05b-b0b4e966c16d.png)

When we deployed the component, we found a number of users were experiencing bugs and quickly rolled it back: #558 

The issue related to the `SeriesProgress` value in our GraphQL requests conflicting with progress data coming in from client websites (epic react, testing js): https://github.com/eggheadio/egghead-rails/issues/4476

> If the user has watched any lessons from a WBaaS site (pure react, testing js, epic react etc) this query will break because we are adding the x-site-client header to the graphql request.

> We need to tenant the SeriesProgress model to resolve this. There are 2_551_357 SeriesProgress objects so we have to be careful with this data migration.

@theianjones began working on this with help from @jbranchaud: https://github.com/eggheadio/egghead-rails/pull/4487

It's unclear whether the issue with `SeriesProgress` has been resolved.

Open question: have we fixed the conflicting data between our client websites and egghead? Is it safe to try and reimplement this component?

jbranchaud:
It looks like for the vast majority of Series Progress records, we now have the relevant `site` recorded. If this wasn't working before because we didn't know what records belong with what Site Tenant, then it should now have the data it needs.

Here is a query showing that most `series_progresses` records have a `site` recorded.

![CleanShot 2021-08-24 at 13 26 17@2x](https://user-images.githubusercontent.com/694063/130670362-60574b7c-c30d-43d2-a5a1-2a9bf846db39.png)

DrShpongle:
Is it actual still? cc @laurosilvacom

PR #845, opened 2021-08-25
Rebirth of the In Progress component

MaggieAppleton:
A reimplementation of PR #551

We ran into some technical issues with the first attempt that should be resolved now. Context: https://github.com/eggheadio/egghead-next/issues/842

Component will appear when users are logged in and have current courses in progress:

image

image

### 📹 Preview of interaction: https://share.getcloudapp.com/qGuJl8W0

Since the last attempt at doing this caused issues for a number of users, how can be stress test this to make sure it's not going to break the experience for anyone?

### DIFF
--- a/src/components/pages/home/in-progress-section/index.tsx
+++ b/src/components/pages/home/in-progress-section/index.tsx
@@ -18,7 +18,7 @@ const InProgressSection: FunctionComponent<InProgressSectionProps> = ({
   coursesInProgress,
 }) => {
   return (
-    <section className="pt-4 pb-10">
+    <section className="pt-4 pb-12">
       <div className="flex justify-between align-text-top">
         <h2 className="md:text-xl text-lg mb-4 text-left">
           {isEmpty(viewer.name) ? (
--- a/src/components/pages/home/index.tsx
+++ b/src/components/pages/home/index.tsx
@@ -2,10 +2,12 @@ import React, {FunctionComponent} from 'react'
 import {Card} from 'components/card'
 import Link from 'next/link'
 import Image from 'next/image'
-import {map, get, isEmpty} from 'lodash'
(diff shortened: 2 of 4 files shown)


jbranchaud:
This looks great!

> Since the last attempt at doing this caused issues for a number of users, how can we stress test this to make sure it's not going to break the experience for anyone?

This is a good question. At some point I think we'd benefit from introducing feature flags so that we have more control over when something is enabled and for who. As for how we could test this feature _today_, I have two thoughts:

1. We could do a local/dev smoke test of it. If the issue is primarily to ensure that partner site series progresses aren't showing up on egghead-next, then we create a user with recent series progresses on a partner site and then sign in. Check that all looks good and partner site progresses are excluded.
2. We could deploy as is and make sure to test it promptly with a user who will have recent series progresses for a partner site.

MaggieAppleton:
### New update

I mocked up a loading component and an empty state for learners who don't have any courses in progress.

These are currently toggled on/off by temporary variables at the top of the file – `showLoader` and `showPlaceholder`. Logic for conditional rendering needs to be added.

📹 Video walkthrough: https://share.getcloudapp.com/v1u0z61m

#### Loading view: (To prevent layout shift)

image

#### Empty state:

image

image

#### Responsive view:

Frame 58

---

### Remaining Todos

- [ ] Mockup dismissal function on in-progress cards (eg. a small cross in the top corner of resources that allows learners to remove them from this section)
- [ ] Setup conditional logic to render the spinner while data loads, and empty state if the learner has no courses.

---

### Out of scope improvements

I'm still on board with making the results that show up here more sophisticated and personalised to the user. Eg. not just courses in progress; also relevant CTAs, recommended content, etc.

This version feels like a good first step in that direction.

PR #849, opened 2021-08-27
re-enable autoplay

joelhooks:
weird eyes


src/pages/lessons/[slug].tsx
```diff
@@ -342,18 +339,6 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
     }
   }, [currentPlayerState])
 
-  React.useEffect(() => {
-    const handleRouteChange = () => {
-      if (!autoplay) {
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:347]: this `!autoplay` was the culprit here and we were never entering the load state. Moved this below into the effect that monitors for a slug change (same idea) and send the load and initial lesson then

src/pages/lessons/[slug].tsx
```diff
@@ -379,6 +364,12 @@ const Lesson: FunctionComponent<LessonProps> = ({initialLesson}) => {
       )
     }
 
+    send({
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:367]: we have an `initial lesson` and can put player in loading state now

DrShpongle (approved):

PR #859, opened 2021-09-02
Rearrange switch case to promote Portfolio Foundations

tayiorbeii:
I forgot that `case` order matters when writing `switch` statements 🤦 

Moving the Portfolio Foundations sequence up so we can get some people seeing the sequence that Maggie & I worked so hard on 😎 

donkey kong ladder climb


MaggieAppleton (commented):
Looks good to me. Ship it 😎

MaggieAppleton (approved):

PR #864, opened 2021-09-08
Turn disabled buttons into informative, actionable modals on the course page

MaggieAppleton:
## Background Context

Our course pages offer three action items; bookmark, download, and RSS feed. You need to be logged in to bookmark a course and you need to be a pro member to download or access the RSS feed. These buttons are currently visible but disabled (greyed out, non-clickable) for users who don't have the required access level.

The disabled buttons don't tell users _why_ the buttons are inaccessible to them. It's unclear how they might get access to bookmarks, downloads, or RSS.

image

**Research and WIP trail in Roam**: https://roamresearch.com/#/app/egghead/page/tRjxuF2bx

## Implemented Solution

This PR turns our disabled buttons into more informative and actionable elements. It keeps the buttons visible and clickable for all users. If a user doesn't have the right access level, it presents the buttons with a lock icon:

image

Clicking each button triggers a modal dialogue that tells them why the feature is unavailable and then gives with an actionable CTA to gain access.

### Locked bookmarks overlay
Prompts users to sign in with email or Github. Offers an option to create an account if they don't have one (link at the bottom of the modal -- this should swap the copywriting on the modal to say "Create your free account" with the same input forms. Not yet implemented)

image

### Locked downloads overlay
Prompts users to become a member, or learn more about membership. "Become a member" button should sent people to the Stripe checkout flow. "Learn more..." link should direct people to the `/pricing` page (not yet implemented).

image

### Locked RSS overlay

image

## Updated bookmarks button

I also went ahead and adjusted the visual styling on the bookmarks button so that it's more obvious when a course is bookmarked.

**Previous styling on a bookmarked course**
image

**Updated styling on a bookmarked course**
Screenshot - 2021-09-08 14 46 01

⚡️ Live preview: https://egghead-io-nextjs-lymh42exu-eggheadio1.vercel.app/courses/the-beginner-s-guide-to-react

## Remaining Tasks

- [ ] Make buttons and links functional. At the moment they're all just plain UI elements w/o handler functions
- [ ] Create logic to swap copywriting on the sign in / create an account modal – happens if users click the "create a free account" link at the bottom of the modal.
- [ ] Add tracking functionality to all buttons and links so we can see usage stats in mixpanel
- [ ] Run through accessibility checks and test usability for each variety of user (visitor, free account, pro)


src/components/layouts/collection-page-layout.tsx
```diff
@@ -1094,4 +1204,92 @@ const transformSanityTopics = (topics: any) => {
...
+      <button className={buttonStyles && buttonStyles} onClick={openDialog}>
+        <LockIcon className="h-4 w-4 mr-2" /> {buttonText}
+      </button>
+      <DialogOverlay
+        isOpen={showDialog}
+        onDismiss={closeDialog}
+        className="bg-black/50 backdrop-blur-sm flex justify-center items-center p-4 xs:px-6 xs:py-8 z-50"
+      >
+        <DialogContent
+          aria-label="sign in to bookmark"
```
joelhooks, line 1228: a11y: if we are gonna use an aria-label, we need to make sure it's correct

src/components/layouts/collection-page-layout.tsx
```diff
@@ -1094,4 +1204,92 @@ const transformSanityTopics = (topics: any) => {
   return topics.items
 }
 
+const DialogButton = ({title, buttonStyles, buttonText, children}: any) => {
```
joelhooks, line 1207: suggestion [code style]: this is a monster above and I think that the DialogButton is likely generally useful so I'd recommend extracting it out into its own component file

src/components/layouts/collection-page-layout.tsx
```diff
@@ -1094,4 +1204,92 @@ const transformSanityTopics = (topics: any) => {
...
+            </button>
+          </div>
+        </DialogContent>
+      </DialogOverlay>
+    </>
+  )
+}
+
+const PlanFeatures = () => {
+  const DEFAULT_FEATURES = [
```
joelhooks, line 1258: suggestion [single source]: this data is also used in the "pick a plan" component with a different presentation. I think it makes sense for them both to say the same thing and use the same set of features.

https://github.com/eggheadio/egghead-next/blob/2b7c352b4ccb2b8d641b0ba45274e3fed6057c72/src/components/pricing/select-plan/index.tsx

The `SelectPlan` component is a more complex presentation, but it's the same data and seeing it copied here made me think we might just want to have the data aspect in one place.

tailwind.config.js
```diff
@@ -133,6 +133,9 @@ module.exports = {
         '3xl': '1920px',
         print: {raw: 'print'},
       },
+      width: {
```
joelhooks, line 136: opinion: seeing tailwind global config updated to affect change in (what I'm assuming is) a one-off problem is concerning and makes me think that perhaps we should sometimes "just use css" if tailwind's config isn't doing the work we need it to do

src/components/layouts/collection-page-layout.tsx
```diff
@@ -1094,4 +1204,92 @@ const transformSanityTopics = (topics: any) => {
...
+        <LockIcon className="h-4 w-4 mr-2" /> {buttonText}
+      </button>
+      <DialogOverlay
+        isOpen={showDialog}
+        onDismiss={closeDialog}
+        className="bg-black/50 backdrop-blur-sm flex justify-center items-center p-4 xs:px-6 xs:py-8 z-50"
+      >
+        <DialogContent
+          aria-label="sign in to bookmark"
+          className="bg-white dark:bg-gray-900 shadow-lg rounded-lg max-w-screen-sm border dark:border-gray-800 relative p-3 xs:p-6 w-112"
```
joelhooks, line 1229: suggestion: instead of a custom one-off tailwind class to achieve this we might consider just hard coding the width value as a style property unless this will be a common need across the entire app

related: https://github.com/eggheadio/egghead-next/pull/864/files#r708555190

src/components/layouts/collection-page-layout.tsx
```diff
@@ -530,24 +535,84 @@ const CollectionPageLayout: React.FunctionComponent<CoursePageLayoutProps> = ({
...
-                  dark:hover:bg-gray-700 border-gray-300 bg-white dark:bg-gray-800 dark:border-gray-600 px-4 py-2 border transition-colors text-sm xs:text-base ease-in-out opacity-30"
+                  <DialogButton
+                    buttonText="Bookmark"
+                    title="Sign in to bookmark"
+                    buttonStyles="text-gray-600 dark:text-gray-300 flex flex-row items-center rounded hover:bg-gray-100 
+                      dark:hover:bg-gray-700 border-gray-300 bg-white dark:bg-gray-800 dark:border-gray-600 px-4 py-2 border transition-colors text-sm xs:text-base ease-in-out opacity-90 shadow-sm"
                   >
-                    <BookmarkIcon className="w-4 h-4 mr-1" /> Bookmark
-                  </div>
+                    <div className="flex flex-col space-y-4">
```
joelhooks, line 559: suggestion [reuse]: We have a common login form that should probably be repurposed for this dialog

https://github.com/eggheadio/egghead-next/blob/03c48cb041dc7e0c83d9fcc2c31881e649dc09e4/src/pages/login.tsx

This form is exported from a "page" which is unorthodox, but is also used as a component in the "login cta overlay" that appears on lessons

https://github.com/eggheadio/egghead-next/blob/f120bf7a8a1390b3efddf6796bc698cb12bc5cdc/src/components/pages/lessons/overlay/email-capture-cta-overlay.tsx#L24-L41

The form is modestly customizable but would benefit from extraction into a component versus a page. This is somewhat arbitrary as a page is "just a component" but makes sense as we reuse it.

This might require some adjustments to the form component itself and/or the design presented here

src/components/layouts/collection-page-layout.tsx
```diff
@@ -559,20 +624,39 @@ const CollectionPageLayout: React.FunctionComponent<CoursePageLayoutProps> = ({
...
-                    <FolderDownloadIcon className="w-4 h-4 mr-1" /> Download
-                  </div>
+                    <div className="flex flex-col space-y-4">
+                      <p className="max-w-10 text-center text-gray-700">
+                        As a member you can download any of our courses and
+                        watch them offline. As a member you'll also get:
+                      </p>
+                      <PlanFeatures />
+
+                      <button className="font-semibold w-full mt-6 inline-flex justify-center items-center px-4 py-3 rounded-md bg-blue-600 text-white transition-all hover:bg-blue-700 ease-in-out duration-200">
```
joelhooks, line 647: this should just be a link styled as a button that navigates the learner to the `/pricing` route.

src/components/layouts/collection-page-layout.tsx
```diff
@@ -559,20 +624,39 @@ const CollectionPageLayout: React.FunctionComponent<CoursePageLayoutProps> = ({
...
+                        Become a Member
+                      </button>
+                      <Link href="/pricing">
```
joelhooks, line 650: suggestion: this link (if used now) should likely go to the `/about` page, but that page isn't great and I'd just drop this until it's better

src/components/layouts/collection-page-layout.tsx
```diff
@@ -584,19 +668,45 @@ const CollectionPageLayout: React.FunctionComponent<CoursePageLayoutProps> = ({
...
+                    onClick={() => {
+                      track(`clicked disabled rss feed link`, {
+                        course: course.slug,
+                      })
+                      console.log('hello')
+                    }}
                   >
-                    <RSSIcon className="w-4 h-4 mr-1" /> RSS
-                  </div>
+                    <DialogButton
```
joelhooks, line 686: suggestion [reuse]: this is the same component as the download button above with some text differences so I would recommend creating a component that allowed those strings to be supplied as properties

src/components/layouts/collection-page-layout.tsx
Creeland: Should I move it to components/buttons/index.tsx?

joelhooks:
My general thought process would be "how often will this be reused?" and in my opinion it's probably not THE button or even related to the two buttons that are in that file you link to. I'm not stoked on the "buttons" in that file to begin with (they are actually links styled as buttons which are different semantically in html)


I'd extract these to `src/components/pages/courses` for now

src/components/layouts/collection-page-layout.tsx
Creeland: image
Here is what the modal looks like using the pricing features list

src/components/layouts/collection-page-layout.tsx
```diff
@@ -530,24 +542,45 @@ const CollectionPageLayout: React.FunctionComponent<CoursePageLayoutProps> = ({
...
-                  dark:hover:bg-gray-700 border-gray-300 bg-white dark:bg-gray-800 dark:border-gray-600 px-4 py-2 border transition-colors text-sm xs:text-base ease-in-out opacity-30"
+                  <DialogButton
+                    buttonText="Bookmark"
+                    title="Sign in (or up) to bookmark"
+                    buttonStyles="text-gray-600 dark:text-gray-300 flex flex-row items-center rounded hover:bg-gray-100 
+                      dark:hover:bg-gray-700 border-gray-300 bg-white dark:bg-gray-800 dark:border-gray-600 px-4 py-2 border transition-colors text-sm xs:text-base ease-in-out opacity-90 shadow-sm"
                   >
-                    <BookmarkIcon className="w-4 h-4 mr-1" /> Bookmark
-                  </div>
+                    <LoginForm
```
Creeland, line 566: Design for the various form states needs adjusting. LoginForm component hardcodes `h2`s

src/components/layouts/collection-page-layout.tsx
```diff
@@ -584,19 +620,29 @@ const CollectionPageLayout: React.FunctionComponent<CoursePageLayoutProps> = ({
...
                 ) : (
-                  <div
-                    className="text-gray-600 dark:text-gray-300 flex flex-row items-center rounded hover:bg-gray-200 
-                  dark:hover:bg-gray-700 border-gray-300 bg-white dark:bg-gray-800 dark:border-gray-600 px-4 py-2 border transition-colors text-sm xs:text-base ease-in-out opacity-30"
+                  <a
+                    onClick={() => {
+                      track(`clicked disabled rss feed link`, {
+                        course: course.slug,
+                      })
+                      console.log('hello')
```
joelhooks, line 635:
```suggestion
```

src/components/pages/courses/dialog-button.tsx
```diff
@@ -0,0 +1,65 @@
...
+            {title && (
+              <h4
+                id={labelId}
+                className="text-lg sm:text-xl mb-4 font-semibold text-center px-4 leading-tight"
+              >
+                {title}
+              </h4>
+            )}
+            <div className="flex flex-col space-y-4">
+              {children && children}
```
joelhooks, line 44:
```suggestion
              {children}
```

dunno why that was here but it's not needed

src/components/pages/courses/membership-dialog-button.tsx
```diff
@@ -0,0 +1,28 @@
...
+  return (
+    <DialogButton
+      buttonText={buttonText}
+      title={title}
+      buttonStyles="text-gray-600 dark:text-gray-300 flex flex-row items-center rounded hover:bg-gray-100 
+        dark:hover:bg-gray-700 border-gray-300 bg-white dark:bg-gray-800 dark:border-gray-600 px-4 py-2 border transition-colors text-sm xs:text-base ease-in-out opacity-90 shadow-sm"
+    >
+      <p className="max-w-10 text-center text-gray-700">{children}</p>
+      <p className="max-w-10 text-center text-gray-700">PRO members get:</p>
+      <FeaturesList />
```
joelhooks, line 16: this is more "robust" than the bullets that were here previously and I'd like to see some screenshots

Creeland:
I've recorded a loom walking through the features while logged in and out 👉  https://www.loom.com/share/99a363b16b1240559701afccf50f1621

image
Auth Dialog

image
Download Dialog with features list

image
RSS Dialog with features list

joelhooks (approved):
I think the visual styles of the modals can use refinement since it lost a bit in translation, but that can be a future PR as required.

This looks great! Thanks y'all.

MaggieAppleton:
🚀 Whoo! Thanks for doing great work on this @Creeland.

Some of the styling needed another pass. Spun up a new PR with refinements: https://github.com/eggheadio/egghead-next/pull/873

PR #868, opened 2021-09-13
Created standardised button components

MaggieAppleton:
## Context

The button styles across our site are declared individually in every location. This leads to lots of repetitive code and inconsistencies in the way our buttons look and behave. 

We've discussed this in the past, but decided it's not high on our priority list to go through all the existing code to update old buttons.

However, that shouldn't prevent us from improving button consistency in the future. It's much easier to have single components like `<PrimaryButton />` and `<SecondaryButton />` we can import and know the styles will be consistent across instances. It's also faster to import these rather than hunt for one of multiple button styles + functionality sets to copypasta every time we need a new button.

We can use these new components on future pages and features without any obligation to backfill the codebase with them.

This PR adds two buttons to the codebase – `<PrimaryButton />` and `<SecondaryButton />`

Both take props for `label`, `url`, `onClick` (optional), and `className` (optional)

The secondary button also takes an optional `quiet` prop that makes it slightly less visually dominant.

Preview:

Build the portfolio you need to be a badass web developer  egghead io 2021-09-13 at 12 19 59 pm

Build the portfolio you need to be a badass web developer  egghead io 2021-09-13 at 12 19 49 pm

Hover states:
Screen Recording 2021-09-13 at 12 18 51 pm

## TODO

- [x] Check TypeScript components for errors / bugs (I wrote them and I'm a TS n00b)

### DIFF
--- a/src/components/buttons/index.tsx
+++ b/src/components/buttons/index.tsx
@@ -1,0 +1,21 @@
+import React from 'react'
+
+export function PrimaryButton({className = '', children}) {
+  return (
+    <button
+      className={`${className} inline-flex justify-center items-center px-4 py-2 rounded-md bg-blue-600 text-white font-medium transition-all hover:bg-blue-800 ease-in-out duration-200`}
+    >
+      {children}
+    </button>
+  )
+}
+
+export function SecondaryButton({className = '', children}) {
+  return (
+    <button
+      className={`${className} inline-flex justify-center items-center px-4 py-2 rounded-md bg-blue-600 text-white font-medium transition-all hover:bg-blue-800 ease-in-out duration-200`}
+    >
+      {children}
+    </button>
+  )
+}
--- a/src/components/buttons/index.tsx
+++ b/src/components/buttons/index.tsx
@@ -39,10 +39,10 @@ export const SecondaryButton: React.FC<ButtonProps> = ({
     <Link href={url}>
       <a
         onClick={onClick}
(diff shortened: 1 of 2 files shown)


NicollGuarnizo:
Looks good 🔥

I just deleted an import that wasn't being used.

NicollGuarnizo (approved):

src/components/buttons/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+}
+
+export const PrimaryButton: React.FC<ButtonProps> = ({
+  url,
+  className,
+  label,
+  onClick = noop,
+}) => {
+  return (
+    <Link href={url}>
```
joelhooks, line 20: suggestion: if we are going to really dig in to standardize a single common button is that we should be focused on a11y and semantics. There are some distinctive differences between button buttons and link buttons that should be accounted for.

Any clickable element that opens a new page is best represented by a link/anchor. This might also handle an `onClick` in our app, but that is reserved for analytics and not interaction.

An element that performs an action when clicked should be represented by the `button` element or have the appropriate aria role assigned to it.

Marcy Sutton has a great article with more in-depth details about this: https://marcysutton.com/links-vs-buttons-in-modern-web-applications

This is a well-considered article with React code that demonstrates a very solid solution/abstraction for this issue: https://kittygiraudel.com/2020/01/17/accessible-links-and-buttons-with-react/

PR #870, opened 2021-09-14
algolia search sort component

joelhooks:
filter out the crap

### DIFF
--- a/src/components/search/index.tsx
+++ b/src/components/search/index.tsx
@@ -230,8 +230,7 @@ const Search: FunctionComponent<SearchProps> = ({
                       label: 'Most Popular',
                     },
                     {value: 'reviews', label: 'Highest Rated'},
-                    {value: 'updated', label: 'Recently Updated'},
-                    {value: 'created', label: 'Recently Published'},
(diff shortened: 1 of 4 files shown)


src/components/search/index.tsx
```diff
@@ -116,6 +117,19 @@ const Search: FunctionComponent<SearchProps> = ({
         searchState={searchState}
         {...rest}
       >
+        <SortBy
```
joelhooks, line 120: this drops the component completely unstyled in the top of the search section

src/lib/search-url-builder.ts
```diff
@@ -72,10 +72,12 @@ export const createUrl = (searchState: {
   query?: any
   refinementList?: any
   page?: number
+  sortBy?: string
```
joelhooks, line 75: these changes make this "just work" as a query param and let's us persist state via the url

DrShpongle:
Great!
We have some responsive styles issues on the search page (and curated pages). I think it can be done in next PR's so I hit "approve" on this one. It's the draft PR, so it will be merged after you convert it to regular PR.

PR #876, opened 2021-09-17
Merge homepage sanity data queries 

zacjones93:
This PR consolidates the queries that we've been building into a single `homePageFeatures` query. It queries for a `landing-page` resource in sanity that contains a list of features that we want to display. 

Image 2021-09-17 at 3 49 46 PM


src/components/pages/home/index.tsx
```diff
@@ -20,18 +20,47 @@ const Home: FunctionComponent<any> = ({homePageData}) => {
...
+
+  const typescriptFeatures = find(homePageData?.homePageFeatures?.features, [
+    'slug.current',
+    'typescript-features',
+  ])
+
+  const awsFeatures = find(homePageData?.homePageFeatures?.features, [
+    'slug.current',
+    'aws-features',
+  ])
```
zacjones93, line 63: One thing I don't like is that we still need to know what data is coming in on the front end. 

If the resource indicated what component it needed to be rendered we could simplify this by mapping over the data and switching on components that we have available for features

src/components/pages/home/index.tsx
```diff
@@ -59,11 +88,11 @@ const Home: FunctionComponent<any> = ({homePageData}) => {
         </section>
 
         <section className="mt-20 sm:mt-24">
-          <FeatureRow resource={reactFeatures} />
+          {reactFeatures && <FeatureRow resource={reactFeatures} />}
```
zacjones93, line 91: safely rendering features..

src/pages/index.tsx
```diff
@@ -64,163 +64,38 @@ export const whatsNewQuery = groq`*[_type == 'resource' && slug.current == "what
...
+      related[]->{
+      title,
+      'description': summary,
+      path,
+      'byline': meta,
+      image,
+      byline
+    }
+  }
+}`
```
zacjones93, line 91: Home page feature data is all largely the same which this query covers querying for all the data that the individual queries we fetching

src/pages/index.tsx
```diff
@@ -64,163 +64,38 @@ export const whatsNewQuery = groq`*[_type == 'resource' && slug.current == "what
...
   'featureWhatsNew': ${whatsNewQuery},
+  'homePageFeatures': ${homePageFeatures},
   'featureDeveloperPortfolio': ${developerPortfolioQuery},
-  'reactFeatures': ${reactFeaturesQuery},
-  'javascriptFeatures': ${javascriptFeaturesQuery},
-  'cssFeatures': ${cssFeaturesQuery},
-  'reduxFeatures': ${reduxFeaturesQuery},
-  'typescriptFeatures': ${typescriptFeaturesQuery},
-  'kcdFeatures': ${kcdFeaturesQuery},
-  'awsFeatures': ${awsFeaturesQuery},
```
zacjones93, line 223: delete lots of queries!

src/components/pages/home/index.tsx
```diff
@@ -268,7 +297,13 @@ const WhatsNew: FunctionComponent<any> = ({resource, location = 'home'}) => {
...
 }
 
-const FeatureRow: FunctionComponent<any> = ({resource, location = 'home'}) => {
+const FeatureRow: FunctionComponent<any> = ({
+  resource,
+  location = 'home',
+}: {
+  resource: CardResource
+  location: string
+}) => {
```
zacjones93, line 306: Add CardResource type to the resource received

src/types.ts
```diff
@@ -12,8 +12,10 @@ export type CardResource = Resource & {
   id?: string
   image: string | {src: string; alt: string}
   name: string
+  subTitle?: string
   byline: string
   resources?: CardResource[]
+  related?: CardResource[]
```
zacjones93, line 18: sometimes we need `subTitle` and `related` fields from sanity

src/components/pages/home/index.tsx
```diff
@@ -149,19 +178,19 @@ const Home: FunctionComponent<any> = ({homePageData}) => {
...
         </section>
 
         <section className="mt-20 sm:mt-24">
-          <FeatureRow resource={reduxFeatures} />
+          {reduxFeatures && <FeatureRow resource={reduxFeatures} />}
         </section>
 
         <section className="mt-20 sm:mt-24">
-          <FeatureRow resource={typescriptFeatures} />
+          {typescriptFeatures && <FeatureRow resource={typescriptFeatures} />}
```
Creeland, line 193: I wonder if you could take this and 
```
 const typescriptFeatures = find(homePageData?.homePageFeatures?.features, [
    'slug.current',
    'typescript-features',
  ])
```

and combine them into a component that takes in a string like `'typescript-features'` as props

zacjones93: finding the data by slug is something I want to address in a later PR. It would be nice if the front end didn't care at all what actual data it received

src/components/pages/home/index.tsx
lsminter: Are we planning on always keeping these as `react features` or `aws features`? Just trying to think of a naming convention that will last the years, (or if this a topic that really matters that much).

src/pages/index.tsx
```diff
@@ -64,163 +64,37 @@ export const whatsNewQuery = groq`*[_type == 'resource' && slug.current == "what
...
-  path,
-  name,
-  resources[]->{
+export const homePageFeatures = groq`*[_type == 'resource' && slug.current == "home-page"][0]{
+  'features': resources[]-> {
     title,
-    'description': summary,
+    slug,
+    subTitle,
     path,
```
NicollGuarnizo, line 72: Currently, we are displaying "Browse more {resource.name}", with these changes it looks like we are only displaying "Browse more" -- Is that the intention? If not, adding just `name` in the query would fix that @zacjones93

zacjones93: good catch!

src/components/pages/home/index.tsx
zacjones93: In a future PR I'm hoping to remove the need for naming these variables but it definitely makes sense to have a naming scheme on the sanity side of things

src/components/pages/home/index.tsx
```diff
@@ -20,18 +20,47 @@ const Home: FunctionComponent<any> = ({homePageData}) => {
...
   const featureWhatsNew: any = get(homePageData, 'featureWhatsNew')
-  const reactFeatures: any = get(homePageData, 'reactFeatures')
-  const javascriptFeatures: any = get(homePageData, 'javascriptFeatures')
-  const cssFeatures: any = get(homePageData, 'cssFeatures')
-  const reduxFeatures: any = get(homePageData, 'reduxFeatures')
-  const typescriptFeatures: any = get(homePageData, 'typescriptFeatures')
-  const kcdFeatures: any = get(homePageData, 'kcdFeatures')
-  const awsFeatures: any = get(homePageData, 'awsFeatures')
+
+  const reactFeatures = find(homePageData?.homePageFeatures?.features, [
```
joelhooks, line 30: looking at all these I'm thinking about a function that would look like:

```typescript
const findFeature = (featureName: string) => {
  return find(homePageData?.homePageFeatures?.features, [
    'slug.current',
    featureName
  ]
}
```
This would greatly reduce the repetition so:

```typescript
const reactFeatures = findFeature('react-features')
```

NicollGuarnizo (approved):

PR #879, opened 2021-09-20
Add empty state to search

MaggieAppleton:
## Context

We've been implementing improvements to our search experience over the last few weeks:  https://github.com/eggheadio/egghead-next/pull/871 https://github.com/eggheadio/egghead-next/pull/870

At the moment when learners search for a term that has 0 results, they just get a blank space where results would usually appear.

image

This is potentially confusing as the learner hasn't received clear feedback that their search returned no results. They may think the page is still loading, or an error has occurred.

### 📹 [Loom video walkthrough](https://www.loom.com/share/5f8fc8d8723f4d9190b7f16e9832a60c)

To improve the experience, I've added an empty state that clearly tells the learner we don't have any resources that match their search term.

image

We should also offer the learner a few alternative actions that _would_ return search results. I've included links to four of our most popular topics in here; React, JavaScript, CSS, and TypeScript.

### Dark mode
image

### Mobile
image

## Todo

Some of the functionality on this feature is still broken. Empty state isn't conditionally rendered, and links to topic suggestions aren't working.

- [x] Add conditional rendering to the empty state component so that it only shows up when no search results are returned
- [x] Check that topic suggestion buttons are a11y friendly (currently div's wrapped in a Link and a tag)
- [x] Fix topic suggestion buttons so that they redirect learners to the relevant search page (currently doesn't work without a manual refresh)


DrShpongle:
"no results" eggo is super cute 😍
everything looks good to me, going to add some code to this branch

DrShpongle:
Pushed needed updates.

joelhooks (approved):

PR #880, opened 2021-09-21
move 'sorting hat' to show it's a survey instead

joelhooks:
sorting hat


src/pages/ask.tsx
```diff
@@ -1,10 +1,11 @@
-import SortingHat from 'components/survey/sorting-hat'
+import Survey from 'components/survey/survey'
 import * as React from 'react'
+import {sortingHatInitialState} from '../components/survey/survey-reducer'
 
 const Ask: React.FunctionComponent = () => {
   return (
     <div>
-      <SortingHat></SortingHat>
+      <Survey initialSurveyState={sortingHatInitialState} />
```
joelhooks, line 8: The purpose of this PR is basically here where I'm demonstrating that this is a generic survey and NOT the `sorting hat` so we can pass in ANY `SurveyState` and it will build a survey based on the data

src/components/survey/survey-reducer.ts
```diff
@@ -7,7 +7,33 @@ import {cioIdentify} from 'utils/cio-identify'
 const DEFAULT_FIRST_QUESTION = `biggest_path`
 const DEFAULT_FINAL_QUESTION = `thanks`
 
-export type SortingHatState = {
+export type Survey = {
+  [key: string]: SurveyQuestion | string
```
joelhooks, line 11: the `| string` here is because a `Survey` has a version. Not stoked on this for typing but it allows for arbitrary keys of type `SurveyQuestion` so we can build out datasets for surveys

src/components/survey/survey-reducer.ts
```diff
@@ -7,7 +7,33 @@ import {cioIdentify} from 'utils/cio-identify'
...
+}
+
+export type MultipleChoiceAnswer = {
+  answer: string
+  label: string
+  always_last?: boolean
+}
+
+export type SurveyQuestion = {
```
joelhooks, line 20: these properties cover quite a few types of surveys

src/components/survey/survey-reducer.ts
```diff
@@ -7,7 +7,39 @@ import {cioIdentify} from 'utils/cio-identify'
...
+  heading: string
+  subheading: string
+  type:
```
joelhooks, line 23: I considered creating types for all of these... types, but we can talk more about that

src/components/survey/survey.tsx
```diff
@@ -82,12 +79,14 @@ const SortingHat: React.FunctionComponent<{
         {state.question.type === 'cta-email' && (
           <div>
             <QuestionHeading question={state.question} />
-            <Image
-              className="rounded-full"
-              width={128}
-              height={128}
-              src={state.question.image}
-            />
+            {state.question.image && (
```
joelhooks, line 82: these were causing errors if `undefined`

src/pages/ask.tsx
ParkerGits: survey: used to ask any sequence of questions in a branching fashion
sorting hat: specific implementation of a survey defined by its unique SurveyState
we can implement a "CTA survey" to personalize the active CTA, just as we implemented the sorting hat to reach a final message/state.

ParkerGits (approved):
🧙

PR #886, opened 2021-09-22
Pull site og image from sanity if latest course has image set

zacjones93:
Currently the sites main og image is hard coded in the front end. This allows us to update the og image when a course is released by adding a `main-og-image` to the course. 

When that course is set as the primary course (the one displayed in the jumbotron) its site og image will be used. If there is none, a default will be used.

```
<meta property="og:image" content="https://res.cloudinary.com/dg3gyk0gu/image/upload/v1631646052/egghead-next-pages/jotai/main-og-image--jotai-atoms.png">
```

**latest course:**
jotai og image

**default:**
default og image


src/pages/index.tsx
```diff
@@ -2,20 +2,27 @@ import React, {FunctionComponent} from 'react'
...
 import {sanityClient} from 'utils/sanity-client'
 import staticHomePageData from 'components/pages/home/homepage-data'
 import {digitalGardeningQuery} from './learn/digital-gardening'
 import {developerPortfolioQuery} from './learn/developer-portfolio'
 
 const IndexPage: FunctionComponent = ({homePageData}: any) => {
+  let courseOgImage = get(
+    homePageData,
+    'featureWhatsNew.primary.resources[0].mainOgImage',
+  )
```
zacjones93, line 15: check for an `mainOgImage` being set on the primary resource.

src/pages/index.tsx
```diff
@@ -2,20 +2,27 @@ import React, {FunctionComponent} from 'react'
...
     <>
       <NextSeo
         canonical={process.env.NEXT_PUBLIC_DEPLOYMENT_URL}
         openGraph={{
           images: [
             {
-              url: 'https://res.cloudinary.com/dg3gyk0gu/image/upload/v1628710683/next.egghead.io/resources/build-a-real-time-data-syncing-chat-application-with-supabase-and-next-js/main-ogImage--nextjs--supabase.png',
+              url: courseOgImage
+                ? courseOgImage
+                : 'https://res.cloudinary.com/dg3gyk0gu/image/upload/v1632239045/og-image-assets/egghead-og-image.png',
```
zacjones93, line 25: if `courseOgImage` is set, use it else use a default og image

src/pages/index.tsx
```diff
@@ -43,6 +50,7 @@ export const whatsNewQuery = groq`*[_type == 'resource' && slug.current == "what
     	image,
       'background': images[label == 'banner-image-blank'][0].url,
       'featureCardBackground': images[label == 'feature-card-background'][0].url,
+      'mainOgImage': images[label == 'main-og-image'][0].url,
```
zacjones93, line 53: gotta query for the data!

PR #887, opened 2021-09-22
Resources by Chris Biscardi Curation

Creeland:
image

### DIFF
--- a/src/components/card/vertical-resource-collection-card.tsx
+++ b/src/components/card/vertical-resource-collection-card.tsx
@@ -20,7 +20,8 @@ const VerticalResourceCollectionCard: React.FC<{
   location?: string
   describe?: boolean
   className?: string
-}> = ({resource, location, className = '', ...props}) => {
+  titleColor?: string
+}> = ({resource, location, className = '', titleColor, ...props}) => {
   className = `${className} border-none flex flex-col sm:py-8 py-6`
 
   return (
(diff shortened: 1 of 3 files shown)


laurosilvacom:
To prevent the layout shifts we need to stop implementing this pattern or figure out a more performant way to load the card background


https://user-images.githubusercontent.com/57044804/134417138-8dcaf50d-cf97-4607-871b-d7a0b9ac7d2e.mp4

joelhooks:
@laurosilvacom this is a css (probably) related layout shift issue that also occurs on topic pages with CTAs.

fix needs to be prioritized, but it isn't related to this PR

NicollGuarnizo:
If you use the current card displaying in his landing page, there is not a shift issue. 


https://user-images.githubusercontent.com/43156646/134453663-7209b474-c832-4c1e-85ad-8507d0024769.mov

joelhooks:
That’s a lot better but there is definitely still significant amounts of layout shift. 


The reason this is important is that Google actually punishes sites that behave this way with lower placement in search results

Creeland:
Refactoring to use `Image` instead of `img` solved the image shift issue but there's still this text that gets shifted underneath the CTA for some reason. I'll figure it out tomorrow @joelhooks @NicollGuarnizo 
https://user-images.githubusercontent.com/518406/134460428-345bdf33-12fe-4340-a7ea-daac7ed6927c.mov

NicollGuarnizo:
I did some minor changes in the margins - 

Before: 
Screen Shot 2021-09-23 at 9 48 53 AM

Now:
Screen Shot 2021-09-23 at 9 50 05 AM

As a side note I created an issue for the weird shift layouts, I will try to figure out as well. [Issue](https://github.com/eggheadio/egghead-next/issues/890)

lsminter:
Something I had to implement with the Next.js landing page is a sort of fail-safe. The Sanity data doesn't always load the first time loading a page. I believe it has something to do with how we search for the data. I was working on it with Zac before his vacation but we weren't able to figure it out. The fail-safe was basically just some hardcoded data in an `if` statement, the `SearchInstructorEssential` component and the `CTA` if the data didn't load, `else` load the sanity data. This is just so that when someone navigates to a certain subject, they aren't just displayed with a blank page.

NicollGuarnizo:
Lucas suggestion will be a good implementation @Creeland 

Other than that, the page looks good🔥

PR #889, opened 2021-09-22
use a div instead of a fragment for layout to avoid double renders

joelhooks:
double render

The footer was rendering twice, which has happened before. Through process of elimination I found that changing the `<>` fragment element to a `<div>` in the layout caused this to return to normal.

An interesting thing to note here is that the rendered double was outside of the `_next` component

outside of _next structure

double vision

### DIFF
--- a/src/pages/q/[[...all]].tsx
+++ b/src/pages/q/[[...all]].tsx
@@ -154,14 +154,13 @@ const SearchIndex: any = ({
 // this fixes the issue with a double footer rendering. 🥴
 SearchIndex.getLayout = (Page: any, pageProps: any) => {
   return (
-    <>
+    <div>
       <Header />
       <Main className="bg-gray-50">
         <Page {...pageProps} />
       </Main>
       <Footer />
-      <div />
-    </>
+    </div>
   )
 }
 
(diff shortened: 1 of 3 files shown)


src/pages/q/[[...all]].tsx
```diff
@@ -154,14 +154,13 @@ const SearchIndex: any = ({
 // this fixes the issue with a double footer rendering. 🥴
 SearchIndex.getLayout = (Page: any, pageProps: any) => {
   return (
-    <>
+    <div>
```
Comment (DrShpongle) [inline src/pages/q/[[...all]].tsx:157]: 👍

src/pages/q/[[...all]].tsx
```diff
@@ -154,14 +154,13 @@ const SearchIndex: any = ({
...
       <Header />
       <Main className="bg-gray-50">
         <Page {...pageProps} />
       </Main>
       <Footer />
-      <div />
-    </>
+    </div>
```
Comment (lsminter) [inline src/pages/q/[[...all]].tsx:163]: Double checked and it works perfectly! Going to merge.

lsminter (approved):
Tripple checked locally and it works perfectly!

PR #891, opened 2021-09-23
Fixing layout instructor pages

NicollGuarnizo:
This PR fixes the current layout issue in all instructor pages and leaves the same layout  for future instructor pages. 

tenor-153317489


lsminter (commented):
I double checked each spot she fixed and it's looking great on my end!

dealingwith (approved):
LGTM

PR #895, opened 2021-09-27
Revert "small fix to the default cta to fill the height of the component"

NicollGuarnizo:
Reverts eggheadio/egghead-next#894


NicollGuarnizo:
Reverting those changes since it doesn't work on small descriptions.

lsminter (approved):

PR #899, opened 2021-09-29
New home landing page

MaggieAppleton:
## Intro and Context

When we moved from old egghead to egghead-next back in January we left behind our long form sales page for new visitors to the site.

At the moment when people arrive at egghead.io they all arrive at the same content-heavy homepage. For new visitors this doesn't tell them what egghead is, what they can expect to find here, whether we're selling anything, what we're offering, and how they might buy it if they so desire.

This PR will add a long form sales page to the site as a homepage for new visitors. It uses the copywriting currently available on [egghead.io/about](https://egghead.io/about) (with some spelling, grammar, and clarity corrections).

When a new visitor arrives at egghead.io who isn't logged in and/or doesn't haven't have an account with us, they should see this page. If they _are_ logged in they should immediately be redirected to egghead.io/learn where they can go browse our wall of content (the current egghead.io homepage).

The design is incredibly basic at the moment, and this page should receive more visual design attention in the future. I've used one of the course illustrations as a placeholder in the header, and added some icons to help break the page up. However, it's still very text heavy and visually monotonous. Commissioning illustrators and/or visual designers to take 2nd and 3rd passes at it will help.

📹 **Video walkthrough**: https://share.getcloudapp.com/6quYqGzE

## Dev Todos

- [ ]


laurosilvacom:
We can scope down this PR further by removing the checkout widget component that includes email address validation. This is new functionality, and it's something that needs to be addressed separately and in the context of creating composable components and separating style from behavior. 

My suggestion is to reuse the existing checkout component as it is and adjust the existing design as needed.  

CleanShot 2021-10-05 at 23 46 41@2x

PR #901, opened 2021-10-01
Lm/lazar nikolov instructor page

lsminter:
Created a custom instructor page for Lazar. He only has one course so I got that added as the CTA. 

Screen Shot 2021-10-01 at 12 38 54 PM


NicollGuarnizo (approved):

laurosilvacom:
Will the background color match the home page banner?

NicollGuarnizo:
Yes, all the social assets have been created with the same background color. The home page banner will be updated on Monday after the email announcement.

PR #940, opened 2021-11-01
Add inline subscription flow to blocked lesson CTA

jbranchaud:
There are two versions of this flow:
1. a user is signed in, but not pro, so they see the monthly pricing option and the checkout button
2. a user is not signed in, so they see the monthly pricing option, an email text field, and the checkout button

Both flows support PPP pricing when the user is in an eligible area.

Per: https://github.com/eggheadio/egghead-next/discussions/893

What still needs doing:
- [x] This still needs a lot of styling work. I could use an assist from someone with more design chops than me (@DrShpongle?). I'm happy to hop on a call to pair on this piece of it too. This styling work should be added in before this PR is merged.
- [ ] As a follow up PR, I still need to add in the logic to return the user from Stripe Checkout directly back to the lesson they were trying to view. Also involved in this is wiring up 'Purchase and Play'.

This video shows the flow where a PPP-eligible user who doesn't have an account (or isn't signed in) is able to choose their pricing and advance to the Stripe Checkout flow.

https://user-images.githubusercontent.com/694063/139704285-e05630fb-5173-4654-889c-bce3095c1e1a.mp4


src/components/pages/lessons/overlay/go-pro-cta-overlay.tsx
```diff
@@ -2,30 +2,170 @@ import React, {FunctionComponent} from 'react'
...
+      </button>
+      {pppCouponAvailable && pppCouponEligible && (
+        <div className="mt-4 pb-5 max-w-screen-md mx-auto">
+          <SmallParityCouponMessage
+            coupon={parityCoupon as Coupon}
+            countryName={countryName as string}
+            onApply={onApplyParityCoupon}
+            onDismiss={onDismissParityCoupon}
+            isPPP={pppCouponIsApplied}
+          />
```
jbranchaud, line 166: I created a slimmed down version of the Parity Coupon Message, relative to the one that displays on the pricing page.

src/hooks/use-commerce-machine.ts
```diff
@@ -32,11 +32,19 @@ const extractPricesFromPricingData = (pricingData: PricingData): Prices => {
...
 }
 
-export const useCommerceMachine = () => {
+type Options = {
+  initialPlan: 'annualPrice' | 'monthlyPrice' | 'quarterlyPrice'
+}
+
+export const useCommerceMachine = (
+  options: Options = {initialPlan: 'annualPrice'},
+) => {
```
jbranchaud, line 41: This makes the initial plan selected by the Commerce Machine hook configurable. It defaults to annual, but for this CTA overlay, we can override it as monthly.

src/components/pages/lessons/overlay/go-pro-cta-overlay.tsx
```diff
@@ -2,30 +2,170 @@ import React, {FunctionComponent} from 'react'
...
 
 type JoinCTAProps = {
   lesson: LessonResource
 }
 
 const GoProCtaOverlay: FunctionComponent<JoinCTAProps> = ({lesson}) => {
+  const {viewer, authToken} = useViewer()
+  const {state, send, priceId, quantity, availableCoupons, currentPlan} =
+    useCommerceMachine({initialPlan: 'monthlyPrice'})
+
```
jbranchaud, line 27: A lot of the variables and derived data below is very similar to what we are doing on the pricing page, but there are minor differences to fit this CTA overlay. In particular, the `onClickCheckout` handler in particular.

src/components/pages/lessons/overlay/go-pro-cta-overlay.tsx
```diff
@@ -2,30 +2,170 @@ import React, {FunctionComponent} from 'react'
...
+        track('cta overlay checkout: existing pro account found', {
+          email: emailForCheckout,
+        })
+
+        // email is already associated with a pro account, return early instead
+        // of sending the user to a Stripe Checkout Session.
+        return
+      }
+    }
```
jbranchaud, line 97: Similar to the _Step 2: Email_ form that we inject between the Pricing Page and Stripe Checkout, we do a Pro Access Check on the user-provided email.

src/components/pages/lessons/overlay/go-pro-cta-overlay.tsx
```diff
@@ -2,30 +2,170 @@ import React, {FunctionComponent} from 'react'
...
-          onClick={() => {
-            track('clicked join cta on blocked lesson', {
-              lesson: lesson.slug,
-            })
+      <h1 className="sm:text-2xl text-xl leading-tighter tracking-tight font-light text-center">
+        Ready to take your career to the next level?
+      </h1>
+      <PlanPrice pricesLoading={pricesLoading} plan={currentPlan} />
+      {isEmpty(viewer) && (
+        <input
```
jbranchaud, line 133: We only display the Email Input if there isn't a viewer (i.e. the visitor isn't signed in)

vojtaholik:
updated design:
screenshot@2x

vojtaholik (approved):

jbranchaud:
Thanks, @vojtaholik. This looks great!

PR #945, opened 2021-11-02
middleware for signup page

joelhooks:
the middle


src/utils/auth.ts
```diff
@@ -14,7 +15,6 @@ const AUTH_CLIENT_ID = process.env.NEXT_PUBLIC_CLIENT_ID
 const AUTH_REDIRECT_URL = process.env.NEXT_PUBLIC_REDIRECT_URI
 
 export const USER_KEY = 'eh_user_2020_11_22'
-export const ACCESS_TOKEN_KEY = 'eh_token_2020_11_22'
```
joelhooks, line 17: this was throwing a `window` error importing this file in edge function environment so I moved it out

PR #953, opened 2021-11-04
Put Purchase and Play logic in custom hook

jbranchaud:
This moves the Purchase and Play logic into a custom hook. That does a couple things:
- it helps isolate some dependencies, making it easier to reason about
- it makes it so that the hook can be used in other post-checkout contexts (e.g. lesson CTA overlay)


joelhooks (approved):
I think we should consider tests for the logic here now that we are starting to refine/abstract

PR #956, opened 2021-11-05
Exclude empty coupon param from URL

jbranchaud:
Before this, if there wasn't a coupon being applied, an empty `coupon`
param would still appear in the URL.

CleanShot 2021-11-05 at 10 46 35@2x


joelhooks (approved):
nice catch

PR #962, opened 2021-11-09
Simplify machine tests with onTransition

jbranchaud:
Remove a bunch of weird `await sleep(0)` and excess expectations by using `onTransition` and `done()`.

Per: https://github.com/skillrecordings/products/blob/main/packages/commerce/src/machines/commerce/commerce-machine.test.ts

wow

### DIFF
--- a/src/machines/__tests__/auth-token-polling-machine.test.ts
+++ b/src/machines/__tests__/auth-token-polling-machine.test.ts
@@ -2,12 +2,6 @@ import {authTokenPollingMachine} from '../auth-token-polling-machine'
 import {interpret} from 'xstate'
 import {SimulatedClock} from 'utils/test/simulated-clock'
 
-function sleep(time: number) {
-  return new Promise((resolve) => {
-    return setTimeout(resolve, time)
-  })
-}
-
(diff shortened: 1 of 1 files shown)


src/machines/__tests__/auth-token-polling-machine.test.ts
```diff
@@ -42,25 +36,32 @@ test('it invokes the requestAuthToken service', () => {
...
       },
     }),
-  )
+    {clock},
+  ).onTransition((state) => {
+    if (state.matches({pending: 'verifyRetrieval'})) {
+      clock.increment(4000)
+    }
+
+    if (state.matches('authTokenRetrieved')) {
```
joelhooks, line 54: One of my hardline rules is "no logic in tests"

https://www.slideshare.net/joelhooks/unit-testing-guidelines

Screen Shot 2021-11-09 at 3 28 55 PM

If this condition isn't true the test passes, which doesn't seem like a situation we'd want.

I don't fully understand the clock overhead yet

joelhooks (approved):
overall this is a big improvement

PR #964, opened 2021-11-10
Remove need for SimClock from one test

jbranchaud:
Experimenting with one way of eliminating a conditional and the bleed of implementation details from machine tests.

By making the polling delay configurable and then overriding it to `0ms` in the test, I don't need to `increment` a simulated clock.

simulated clock

### DIFF
--- a/src/machines/__tests__/auth-token-polling-machine.test.ts
+++ b/src/machines/__tests__/auth-token-polling-machine.test.ts
@@ -80,27 +80,24 @@ test('it schedules the next poll, then done', (done) => {
 })
 
 test('polls several times for the value', (done) => {
-  const clock = new SimulatedClock()
-
   // mock a couple failed requests followed by a successful request
   const requestAuthToken = jest
     .fn()
-    .mockRejectedValueOnce(null)
-    .mockRejectedValueOnce(null)
-    .mockResolvedValueOnce({authToken: 'auth123'})
+    .mockRejectedValueOnce(null) // 1st request fails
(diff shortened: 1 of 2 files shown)


src/machines/auth-token-polling-machine.ts
```diff
@@ -108,5 +108,8 @@ export const authTokenPollingMachine = createMachine<Context, Event>(
         return typeof context.authToken !== 'string'
       },
     },
+    delays: {
+      WAIT_BETWEEN_POLLS: 2000,
+    },
```
jbranchaud, line 113: This does two things:
- gives a name to a magic number which provides some additional context
- makes the delay configurable by the interpreter of the machine

The latter gives us more flexibility in a testing context.

src/machines/__tests__/auth-token-polling-machine.test.ts
```diff
@@ -80,27 +80,24 @@ test('it schedules the next poll, then done', (done) => {
...
+    .mockResolvedValueOnce({authToken: 'auth123'}) // 3rd request succeeds!
 
   const pollingService = interpret(
     authTokenPollingMachine.withConfig({
       services: {
         requestAuthToken,
       },
+      delays: {
+        WAIT_BETWEEN_POLLS: 0,
+      },
```
jbranchaud, line 97: Overriding the polling delay to `0ms` means the test can quickly simulate the machine's behavior without using the `SimulatedClock` to _advance_ time.

joelhooks (approved):

PR #968, opened 2021-11-12
Add Purchase and Play to Blocked Lesson CTA

jbranchaud:
This PR introduces most of the logic and components for Purchase and Play on the blocked lesson CTA (https://github.com/eggheadio/egghead-next/discussions/893).

If you are not signed in (or not Pro), go to a Pro-only lesson. You'll see the CTA for inline checkout. Complete checkout and you should be returned directly to the lesson. An overlay with a Thank You message will appear as the client attempts to load your new account (auth token). Once it is loaded a button will appear inviting you to proceed to view the Pro lesson you now have access to.

What is still needed at this point:
- [x] improved styles catered to the lesson overlay

I would love an assist on the styling from @DrShpongle if you're available. Otherwise, I can take a swing at it next week.

- [x] ~proper machine transition to viewing the lesson~ for the time being, we are triggering a reload of the page

@joelhooks I'm not that familiar with the Player Machine yet. I tried adding a transition from `subscribing` to `viewing`, but it gets stuck with the loading spinner. Maybe there is more to successfully shifting to that state. Any insights?

https://user-images.githubusercontent.com/694063/141511742-66483286-f22c-4ba2-b2a8-63e6bceaf95c.mp4

lesson


joelhooks:
a punt might be to hard refresh the page. i feel like it may need to go back to “loading” but won’t be able to look today

vojtaholik:
In terms of UX I think it would be just fine to use the existing thank you page with a CTA to continue where they left off. Bit of context switching but I don't think we have to avoid that in this case.

jbranchaud:
> In terms of UX I think it would be just fine to use the existing thank you page with a CTA to continue where they left off. Bit of context switching but I don't think we need to avoid it in this case. And we get to refresh the page "for free".

I'm generally completely in favor of these kinds of simplifications, but, as I understand it, the "near-instant access to the course" is one of the core elements of this pitch (https://github.com/eggheadio/egghead-next/discussions/893). Dropping the user right back in the context of the course after checkout is something we wanted to try out.

Fwiw, we can toggle back and forth between these two approaches. If we want to experiment with them, all it takes is switching out the the `success_url` that we send to Stripe.

vojtaholik:
Sounds good, thanks for clarification. I guess my only concern is that they have plenty of options to click away while it's processing their account.

joelhooks:
refreshing the page gives me this error, IMO if the query param exists it should function accordingly 
Screen Shot 2021-11-15 at 11 21 05 AM

jbranchaud:
> refreshing the page gives me this error, IMO if the query param exists it should function accordingly Screen Shot 2021-11-15 at 11 21 05 AM

When I refresh, since the 'purchase and play' has logged me in, it loads the video player. Do you have any more specifics about how I might reproduce that error? @joelhooks

joelhooks:
if a session_id was present and a session wasn't loaded it was resulting in an undefined error so I'm passing in the id instead, loading the session in an effect "locally" and guarding against undefined

https://github.com/eggheadio/egghead-next/pull/968/commits/89723d68fd6d5dbf0510823b045fbb7aa7c18327

src/pages/lessons/[slug].tsx
```diff
@@ -557,7 +557,15 @@ const Lesson: React.FC<LessonProps> = ({initialLesson}) => {
...
               {playerState.matches('subscribing') && (
                 <OverlayWrapper>
-                  <GoProCtaOverlay lesson={lesson} />
+                  <GoProCtaOverlay
+                    lesson={lesson}
+                    viewLesson={() => {
+                      router.push({
+                        pathname: router.pathname,
+                        query: {slug: lesson.slug},
+                      })
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:566]:
```suggestion
                      router.reload()
```

jbranchaud:
I've added a Thank You component specialized to the situation where we are unable to look up their session.

This component does a couple things:
    - Thanks them for their purchase
    - Directs them to their inbox for their confirmation email
    - Links them to support because they will likely need help from support if they are seeing this component

CleanShot 2021-11-16 at 13 16 35@2x

src/components/pages/lessons/overlay/go-pro-cta-overlay.tsx
```diff
@@ -280,4 +288,22 @@ const GoProCtaOverlay: FunctionComponent<JoinCTAProps> = ({lesson}) => {
...
-export default GoProCtaOverlay
+const OverlayParent: FunctionComponent<JoinCTAProps> = ({lesson}) => {
+  const {query, reload} = useRouter()
+
+  const {session_id} = query
+
+  if (session_id) {
+    return (
+      <ConfirmMembership
+        lesson={{lesson}}
```
jbranchaud, line 299: I think this was turning into `{{lesson: lesson}}` and instead needs to be `{lesson}`.

```suggestion
        lesson={lesson}
```

joelhooks (approved):

PR #973, opened 2021-11-17
fix rootElement

DrShpongle:
this PR's goal is to pass the containerRef from outside and set it properly on player's side

related PR: https://github.com/joelhooks/cueplayer-react/pull/9


DrShpongle:
Have to update cueplayer-react vesion here

joelhooks:
published cueplayer-react  `0.3.23`

DrShpongle:
updated player

joelhooks (approved):

PR #974, opened 2021-11-17
Show Thank You to signed in user, remove session_id from URL

jbranchaud:
_This plays on top of https://github.com/eggheadio/egghead-next/pull/968, but there are some details in the changeset that I wanted to look at separately in the PR review._

This PR does two things:
1. It allows a signed in user who has just completed Stripe Checkout to see the Thank You overlay. Before this, the `useEffect` state transition logic was immediately loading the lesson after checkout.
2. It removes the `session_id` from the URL after we've fetched their session and displayed the Thank You overlay.

radishes

### DIFF
--- a/src/components/pages/lessons/overlay/confirm-membership.tsx
+++ b/src/components/pages/lessons/overlay/confirm-membership.tsx
@@ -6,6 +6,7 @@ import axios from 'axios'
 import {track} from 'utils/analytics'
 import {useViewer} from 'context/viewer-context'
 import {LessonResource} from '../../../../types'
+import {useRouter} from 'next/router'
 
 type HeaderProps = {
   heading: React.ReactElement
@@ -243,6 +244,7 @@ const ConfirmMembership: React.FC<ConfirmMembershipProps> = ({
   sessionId,
   viewLesson,
 }) => {
+  const router = useRouter()
   const [alreadyAuthenticated, currentState] = usePurchaseAndPlay()
   const [session, setSession] = React.useState<any>()
   const {viewer, refreshUser} = useViewer()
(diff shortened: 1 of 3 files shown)


src/components/pages/lessons/overlay/confirm-membership.tsx
```diff
@@ -250,15 +252,29 @@ const ConfirmMembership: React.FC<ConfirmMembershipProps> = ({
...
+  const {session_id, ...nonSessionQueryParams} = router.query
+
+  const cleanUrlAndViewLesson = () => {
+    router.replace({
+      pathname: router.pathname,
+      query: nonSessionQueryParams,
+    })
+
+    setTimeout(viewLesson, 750)
+  }
```
jbranchaud, line 264: This removes `session_id` query param from the URL and then triggers `viewLesson`. In my testing if `viewLesson` was called immediately after `router.replace`, the updated URL hadn't propagated yet causing the machine state to not update properly. This `timeout` gives it a chance to propagate.

This is a reasonable near-term solution, but it is admittedly a bit hacky. So we can keep this in mind and, at some point, look at more involved alternatives like moving this logic into the machine.

src/pages/lessons/[slug].tsx
```diff
@@ -335,17 +337,30 @@ const Lesson: React.FC<LessonProps> = ({initialLesson}) => {
...
 
-        if (isEmpty(viewer) && free_forever) {
-          if (viewLimitNotReached && mediaPresent) {
+        // TODO: Detangle this nested series of `if` statements to make the
+        // logic more immediately easy to reason about.
+
+        if (session_id) {
+          // If the URL contains the session ID, even if there is a viewer, put
+          // them in the `subscribing` state.
+          send('SUBSCRIBE')
```
Comment (jbranchaud) [inline src/pages/lessons/[slug].tsx:346]: This initial check allows a signed-in user who completes checkout to be shown the Thank You overlay instead of immediately loading the lesson for them.

src/pages/lessons/[slug].tsx
```diff
@@ -393,9 +408,14 @@ const Lesson: React.FC<LessonProps> = ({initialLesson}) => {
...
         break
     }
-  }, [currentPlayerState])
+  }, [currentPlayerState, session_id])
 
   React.useEffect(() => {
+    // only execute the contents of this effect if the machine is in the
+    // process of loading. Any other Player Machine state change should bail
+    // early.
+    if (currentPlayerState !== 'loading') return
```
Comment (jbranchaud) [inline src/pages/lessons/[slug].tsx:417]: allowing this effect to be triggered whenever the player state changes to loading, this allows the player to load when transitioning from 'subscribing' to 'loading'.

src/pages/lessons/[slug].tsx
```diff
@@ -560,7 +580,11 @@ const Lesson: React.FC<LessonProps> = ({initialLesson}) => {
                   <GoProCtaOverlay
                     lesson={lesson}
                     viewLesson={() => {
-                      router.reload()
+                      send({
+                        type: 'LOAD',
+                        lesson: initialLesson,
+                        viewer,
+                      })
```
Comment (jbranchaud) [inline src/pages/lessons/[slug].tsx:587]: Get the user into the `viewing` state with a transition instead of a page reload.

joelhooks:
looks like some type issues and conflicts going on here

joelhooks (approved):

PR #979, opened 2021-11-22
Fix markup on Login screen

DrShpongle:
**before:**
before

**after:**
after

### DIFF
--- a/src/pages/login.tsx
+++ b/src/pages/login.tsx
@@ -54,7 +54,7 @@ const LoginForm: FunctionComponent<LoginFormProps> = ({
       className={
         className
           ? className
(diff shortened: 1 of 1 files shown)


src/pages/login.tsx
```diff
@@ -79,7 +79,7 @@ const LoginForm: FunctionComponent<LoginFormProps> = ({
             children
           ) : (
             <>
-              <h2 className="text-center text-3xl font-bold w-2/3 mx-auto leading-10">
+              <h2 className="w-2/3 mx-auto text-3xl font-bold leading-10 text-center">
```
vojtaholik, line 82:
```suggestion
              <h2 className="w-2/3 mx-auto lg:text-3xl sm:text-2xl text-xl font-bold leading-10 text-center">
```

vojtaholik (approved):

PR #982, opened 2021-11-23
use pusher to get active learner count for a course

joelhooks:
pusher


src/hooks/use-course-presence.tsx
```diff
@@ -0,0 +1,45 @@
...
+  useEffect(() => {
+    if (!slug) return
+
+    const pusher = new Pusher(process.env.NEXT_PUBLIC_PUSHER_KEY || '', {
+      cluster: process.env.NEXT_PUBLIC_PUSHER_CLUSTER,
+      authEndpoint: '/api/pusher/auth',
+    })
+
+    if (contactId) {
+      const channelName = `private-${slug}-${contactId}`
```
joelhooks, line 23: each user gets a private channel that only they belong to. This is per the Pusher guidelines for channels with more than 100 members

https://support.pusher.com/hc/en-us/articles/360019620253-How-can-I-implement-large-presence-channels-on-Channels-

Otherwise, you'd use their "presence channels" but those have a 100 user limit.

In this case, we can return a list of channels to get a count but aren't able to display info on the actual users yet. To do that we'd need to figure out a scheme on our side that utilized a list of users id's to generate images/list of present users.

Most likely we'd just want to display a max of 6 or so avatars and a count, so that should be relatively simple to load data-wise. Was thinking maybe an avatar image endpoint similar to og:image but would accept a Contact ID and return their avatar. Could be cached...

PR #983, opened 2021-11-23
display the current learners avatars

joelhooks:
This is a component that displays the currently active learners' avatars in a component.

I used tailwind UI for the avatar tokens just to get it in here, but it's obviously out of place here.

Where should these go in the app? 

Individual video/lesson pages?

The Beginner's Guide to React egghead io 2021-11-22 at 11 49 43 PM

current events

### DIFF
--- a/src/components/layouts/collection-page-layout.tsx
+++ b/src/components/layouts/collection-page-layout.tsx
@@ -31,9 +31,6 @@ import DialogButton from '../pages/courses/dialog-button'
 import MembershipDialogButton from '../pages/courses/membership-dialog-button'
 
 import LoginForm from 'pages/login'
-import {useCoursePresence} from '../../hooks/use-course-presence'
-import {useViewer} from '../../context/viewer-context'
-import ActiveLearners from '../pages/courses/active-learners'
 
 type CoursePageLayoutProps = {
   lessons: any
@@ -501,8 +498,6 @@ const CollectionPageLayout: React.FunctionComponent<CoursePageLayoutProps> = ({
                     )}
                   </div>
                 </div>
-
-                <ActiveLearners courseSlug={course.slug} />
 
                 <div className="flex flex-row space-x-3 text-sm opacity-80 md:items-start">
                   {created_at && (
--- a/src/components/layouts/collection-page-layout.tsx
+++ b/src/components/layouts/collection-page-layout.tsx
@@ -32,6 +32,8 @@ import MembershipDialogButton from '../pages/courses/membership-dialog-button'
 
 import LoginForm from 'pages/login'
 import {useCoursePresence} from '../../hooks/use-course-presence'
+import {useViewer} from '../../context/viewer-context'
+import ActiveLearners from '../pages/courses/active-learners'
 
 type CoursePageLayoutProps = {
   lessons: any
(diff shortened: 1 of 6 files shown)


src/components/layouts/collection-page-layout.tsx
```diff
@@ -506,13 +502,7 @@ const CollectionPageLayout: React.FunctionComponent<CoursePageLayoutProps> = ({
...
                 </div>
 
-                <div className="flex flex-row space-x-3 text-sm opacity-80 md:items-start">
-                  {count > 1 && (
-                    <span>
-                      {count} learners are watching this course right now.
-                    </span>
-                  )}
-                </div>
+                <ActiveLearners courseSlug={course.slug} />
```
joelhooks, line 505: put this in an external component because this layout is very large and difficult to navigate

src/components/pages/courses/active-learners.tsx
```diff
@@ -0,0 +1,36 @@
+import * as React from 'react'
+import {useViewer} from '../../../context/viewer-context'
+import {useCoursePresence} from '../../../hooks/use-course-presence'
+
+const ActiveLearners: React.FC<{courseSlug: string}> = ({courseSlug}) => {
+  const {viewer} = useViewer()
+  const activeLearners = useCoursePresence(courseSlug)
+  const filteredActiveLearners = activeLearners.filter((learner) => {
+    return learner.avatar_url && learner.avatar_url !== viewer?.avatar_url
```
joelhooks, line 9: filtering the logged-in user, but they could be in there if it made sense.

src/hooks/use-course-presence.tsx
```diff
@@ -27,12 +27,12 @@ export const useCoursePresence = (slug: string) => {
...
     async function checkChannels() {
       const channels = await axios.get(`/api/pusher/channels/${slug}`)
-      let newCount = keys(channels.data).length
-      if (!contactId) newCount++
-      setCount(newCount)
+      setLearners(channels.data)
     }
 
-    const intervalId = setInterval(checkChannels, 750)
+    const intervalId = setInterval(checkChannels, 20000)
```
joelhooks, line 33: refresh every 20 seconds or so

src/pages/api/pusher/channels/[course].ts
```diff
@@ -30,7 +32,14 @@ export default async function handler(
   if (channelData.status === 200) {
     const body = await channelData.json()
     const channelInfo = body.channels
-    res.json(channelInfo)
+
+    const contact_ids = keys(channelInfo).map((channel) => {
+      return last(channel.split('@@')) || ''
```
Comment (joelhooks) [inline src/pages/api/pusher/channels/[course].ts:37]: the `@@` is a separator used in naming the private channels on Pusher where the second half of it is the contact guid for the learner

they "generally suggest something shaped like this" in this article

https://support.pusher.com/hc/en-us/articles/360019620253-How-can-I-implement-large-presence-channels-on-Channels-

they have a more straightforward presence feature but it's limited to 100 connections... maybe it's just wishful thinking on my part ;P

joelhooks:
removed the widget so the functionality is there but the UI is not

vojtaholik (approved):

DrShpongle:
Can't imagine more "right" place for this other than where you placed it. This should be wired to the Course page imo - I don't think this info will be usefull on the the Lesson page (probably it makes sense to have something like that on the home page, even on signed-off landing, something like **X people wathcing Y courses right now**). Regarding this component on the Course page it probably can be moved inside tooltip and everybody who wants to look at the users can click it.

PR #985, opened 2021-11-23
Add types for the topics array

jbranchaud:
topics


src/components/pages/confirm/membership/index.tsx
```diff
@@ -286,7 +286,14 @@ export const ConfirmMembership: React.FC<ConfirmMembershipProps> = ({
   )
 }
 
-const topics: any = [
+type Topic = {
```
joelhooks, line 289: this could probably get promoted to the top-level types

joelhooks (approved):

PR #986, opened 2021-11-23
Add state and data handling for default coupon

jbranchaud:
We were 95% of the way there with site-wide/default coupons. The only thing that was missing was to set the default coupon as the `couponToApply` so that it gets passed to the Stripe Checkout Session.

Without this PR, the pricing page displays the discounted price from the default coupon, but the Stripe Checkout Session shows a full-price subscription.

big sale

### DIFF
--- a/src/machines/__tests__/commerce-machine.test.ts
+++ b/src/machines/__tests__/commerce-machine.test.ts
@@ -43,6 +43,67 @@ const pricingResponseWithPPPAvailable = {
   ],
 }
 
+const pricingResponseWithDefaultApplied = {
+  mode: 'individual',
+  quantity: 1,
+  applied_coupon: {
+    coupon_discount: 0.5,
+    coupon_code: 'QJVIXHB6',
+    coupon_expires_at: 1639814399,
+    default: true,
+    coupon_region_restricted: false,
+  },
+  coupon_code_errors: [],
+  available_coupons: {
+    default: {
+      coupon_discount: 0.5,
+      coupon_code: 'QJVIXHB6',
+      coupon_expires_at: 1639814399,
+      default: true,
+      coupon_region_restricted: false,
+    },
+    ppp: {
+      coupon_discount: 0.6,
+      coupon_code: 'Y48OGYPR',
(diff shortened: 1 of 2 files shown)


src/machines/__tests__/commerce-machine.test.ts
```diff
@@ -142,6 +203,27 @@ test('it can apply PPP coupon when available', async () => {
...
+
+  commerceService.start()
+
+  await sleep(0)
+
+  // expect(commerceService.state.context.priceId).toEqual(undefined)
+  expect(commerceService.state.context.couponToApply?.couponCode).toEqual(
+    'QJVIXHB6',
+  )
+})
```
jbranchaud, line 225: These tests still need to be refactored, but this works for now and ensures that the appropriate context gets set.

src/machines/commerce-machine.ts
```diff
@@ -133,6 +140,7 @@ export const commerceMachine = createMachine<
               },
             },
           },
+          withDefaultCoupon: {},
```
jbranchaud, line 143: Adds a very basic new state. There are no transitions or specific behaviors for this state, but the parent events to SWITCH or CONFIRM price will trigger any necessary transitions.

src/machines/commerce-machine.ts
```diff
@@ -163,7 +171,42 @@ export const commerceMachine = createMachine<
         (_, event) => {
           if (event.type !== 'done.invoke.fetchPricingDataService') return {}
 
-          return {pricingData: event.data}
+          const extractAppliedDefaultCoupon = (
```
jbranchaud, line 174: If the pricing data has a default (site-wide) coupon that has been applied, this will add that to the machine context as `couponToApply` so that it gets passed to the Stripe Checkout Session.

src/machines/commerce-machine.ts
jbranchaud: This `withDefaultCoupon` state can also be referenced in the component if we want to conditionally render any coupon messaging. /cc @vojtaholik

joelhooks (approved):

PR #987, opened 2021-11-24
Move lesson loading logic into machine

jbranchaud:
This PR moves the lesson loading/re-loading logic into the `loading` state of the machine as an [Invoked Callback](https://xstate.js.org/docs/guides/communication.html#the-invoke-property). This way the machine knows how to load the lesson whenever it gets put in the loading state. It eliminates the need for one `useEffect` and it allows another `useEffect` to be trimmed down to just a `send` call.

This PR also does a refactor of lesson/viewer assignments. It creates a single named `action` to do that and then each of the states that needs to perform that on entry can reference that named action. This reduces some duplicated logic.

video loading


src/machines/lesson-player-machine.ts
```diff
@@ -333,6 +319,37 @@ export const playerMachine = Machine<
...
+
+        return {lesson: event.data}
+      }),
+    },
+    services: {
+      loadLesson: () => {
+        throw new Error(
+          "The LessonPlayerMachine's interpreter must implement `loadLesson`.",
+        )
+      },
```
jbranchaud, line 352: The `loadLesson` service, used by the `loading` state, needs to be included in the machine setup wherever this is being interpreted.

src/pages/lessons/[slug].tsx
```diff
@@ -155,19 +155,41 @@ const Lesson: React.FC<LessonProps> = ({initialLesson}) => {
...
+          )
+        }
+
+        console.debug('loading video with auth')
+        const loadedLesson = await loadLesson(initialLesson.slug)
+        console.debug('authed video loaded', {video: loadedLesson})
+
+        return loadedLesson
+      },
+    },
```
Comment (jbranchaud) [inline src/pages/lessons/[slug].tsx:185]: The implementation of `loadLesson` -- this moves the logic from a couple of useEffects into this single spot that the machine can invoked internally.

src/pages/lessons/[slug].tsx
```diff
@@ -411,83 +433,16 @@ const Lesson: React.FC<LessonProps> = ({initialLesson}) => {
...
-
-    run()
   }, [initialLesson.slug])
 
-  // TODO: Temporary duplication of the above useEffect to ensure the lesson is
-  // loaded when transitioning from the 'subscribing'/Thank You overlay to
-  // watching the lesson. If this loading process was moved into the machine as
-  // an Invoked Callback, then state transitions could be employed to handle
-  // what the useEffects are doing.
-  React.useEffect(() => {
```
Comment (jbranchaud) [inline src/pages/lessons/[slug].tsx:452]: This entire useEffect just happens in the machine as part of a transition into the `loading` state.

src/machines/lesson-player-machine.ts
```diff
@@ -333,6 +319,37 @@ export const playerMachine = Machine<
...
+
+        if (lessonData) {
+          assignment['lesson'] = lessonData
+        }
+        if (viewerData) {
+          assignment['viewer'] = viewerData
+        }
+
+        return assignment
+      }),
```
jbranchaud, line 340: The `lesson` and `viewer` values in some cases were being overridden with `undefined`. This implementation ensures that they are only getting set if there is something to set them to.

src/machines/lesson-player-machine.ts
```diff
@@ -58,28 +59,19 @@ export const playerMachine = Machine<
...
-          LOADED: 'loaded',
+        entry: ['assignLessonAndViewer'],
+        invoke: {
+          id: 'fetchLessonDataService',
+          src: 'loadLesson',
+          onDone: {
+            target: 'loaded',
+            actions: ['assignLessonData'],
+          },
+          onError: {},
```
jbranchaud, line 70: The invoked service!

joelhooks (approved):
🎉

PR #994, opened 2021-11-27
fix course mdx widget

vojtaholik:
this fixes `/learn/react` page

engine repair

### DIFF
--- a/src/lib/courses.ts
+++ b/src/lib/courses.ts
@@ -4,10 +4,13 @@ import groq from 'groq'
 
 const courseQuery = groq`
 *[_type == 'resource' && externalId == $courseId][0]{
+  "slug": slug.current,
   title,
   challengeRating,
   description,
   summary,
+  path,
+  image,
   essentialQuestions[]->{
     question
    },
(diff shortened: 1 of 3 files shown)


src/lib/courses.ts
```diff
@@ -66,9 +69,9 @@ const courseQuery = groq`
 }
 `
 
-export async function loadCourse(id: string) {
+export async function loadCourse(id: number) {
   const params = {
-    courseId: id,
+    courseId: Number(id),
```
vojtaholik, line 74: this was the culprit, it needs to be a number, but useSWR is passing a string.

src/pages/learn/react/index.mdx
```diff
@@ -37,7 +37,7 @@ The short story with JSX is that you absolutely need to simply embrace it for wh
 
 If you feel comfortable with your JavaScript skills and have agreed to just accept JSX into your life, we've got an excellent Community Resource for you to explore to get started:
 
-<Course slug="the-beginner-s-guide-to-react" />
+<Course id={432490} />
```
joelhooks, line 40: using numeric ids for this widget is unsustainable in a practical sense

they are very buried versus just grabbing them from the url

vojtaholik: I was following what `lib/courses` expects since that's the loader for this widget. I can make another one that filters by slug for this use case.

joelhooks: on the rails backend i’d/slug are interchangeable. i didn’t use git blame to discover when it changed but know for sure when i made this it was using the slug. 

i suspect the change is because on the Sanity side the ID is safe/stable but not totally sure

joelhooks (approved):

PR #1006, opened 2021-12-01
use isMember

joelhooks:
mbr


src/components/app/header/index.tsx
```diff
@@ -391,7 +393,7 @@ const Browse: React.FC<any> = ({viewer}) => {
             >
               <div className="overflow-hidden rounded-b-lg shadow-lg ring-1 ring-black ring-opacity-5">
                 <div className="relative bg-white dark:bg-gray-800">
-                  {!viewer?.is_pro && (
+                  {!isMember(viewer, subscriber?.attributes) && (
```
joelhooks, line 396: the "customer" is the attributes of the subscriber from useCIO

vojtaholik (approved):

PR #1008, opened 2021-12-02
fix the display of staff notes when available

joelhooks:
merge

### DIFF
--- a/src/lib/lessons.ts
+++ b/src/lib/lessons.ts
@@ -48,6 +48,7 @@ export async function loadLesson(slug: string, token?: string) {
         icon_url
         download_url
         created_at
+        staff_notes_url
         collection {
           ... on Playlist {
             title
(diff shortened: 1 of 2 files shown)


src/lib/lessons.ts
```diff
@@ -48,6 +48,7 @@ export async function loadLesson(slug: string, token?: string) {
         icon_url
         download_url
         created_at
+        staff_notes_url
```
joelhooks, line 51: prop was missing from full lesson load

src/pages/lessons/[slug].tsx
```diff
@@ -181,7 +181,7 @@ const Lesson: React.FC<LessonProps> = ({initialLesson}) => {
         const loadedLesson = await loadLesson(initialLesson.slug)
         console.debug('authed video loaded', {video: loadedLesson})
 
-        return loadedLesson
+        return {...initialLesson, ...loadedLesson}
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:184]: better to combine these IMO allowing the loaded lesson to win

jbranchaud (approved):

PR #1010, opened 2021-12-06
Add shared props interface for Parity Coupon Msg

jbranchaud:
Move duplicated props interface into a shared file for both of the Parity Coupon Message components.

interface

### DIFF
--- a/src/interfaces/parity_coupon_message.ts
+++ b/src/interfaces/parity_coupon_message.ts
@@ -1,0 +1,10 @@
+import {Coupon} from 'types'
+
+export interface ParityCouponMessageProps {
+  coupon: Coupon
+  countryName: string
+  onApply: () => void
+  onDismiss: () => void
+  isPPP?: boolean
+  isLoading?: boolean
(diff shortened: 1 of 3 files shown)


src/interfaces/parity_coupon_message.ts
```diff
@@ -0,0 +1,10 @@
+import {Coupon} from 'types'
+
+export interface ParityCouponMessageProps {
```
joelhooks, line 3: why not just add this to `types.ts` as a `type`?

we've long settled on being a `type` shop unless there's a compelling reason and I'm suspect of the need for a new directory that invites many interfaces to be created 🤔

joelhooks (approved):
approved this with the RFC label so you can merge it at will

src/interfaces/parity_coupon_message.ts
jbranchaud: Fixed! https://github.com/eggheadio/egghead-next/pull/1018

PR #1018, opened 2021-12-14
Remove interfaces dir, reference types instead

jbranchaud:
This was less painful to clean up than I thought.

On the left is the `Viewer` type declared in `src/types.ts`. On the right is the `Viewer` type declared in `src/interfaces/viewer.ts`. I removed the `interfaces` one in favor of the `types` one. The `interfaces` version was only referenced in two places and swapping them out didn't result in any type errors.

CleanShot 2021-12-14 at 16 50 59@2x

On the left is the `Resource` type declared in `src/types.ts`. On the right is the `Resource` type declared in `src/interfaces/resource.ts`. I removed the `interfaces` one in favor of the `types` one here as well. The `interfaces` one wasn't being referenced anywhere, so no impact.

CleanShot 2021-12-14 at 16 51 43@2x

I also moved `ParityCouponMessageProps` into `types.ts`.

The result is that I was able to get rid of the `interfaces/` directory entirely.

big props


joelhooks (approved):
types 4ev4r! 🌮

PR #1019, opened 2021-12-20
link to holiday releases from header, update root share card

vojtaholik:
screenshot

good boy

### DIFF
--- a/src/components/pages/home/index.tsx
+++ b/src/components/pages/home/index.tsx
@@ -1,30 +1,25 @@
 import * as React from 'react'
 import {HorizontalResourceCard} from 'components/card/new-horizontal-resource-card'
 import {VerticalResourceCard} from 'components/card/new-vertical-resource-card'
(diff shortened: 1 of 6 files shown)


src/components/pages/home/index.tsx
```diff
@@ -1,30 +1,25 @@
...
-      ) : (
-        <div className="md:container">
-          <Jumbotron data={jumbotron} />
-        </div>
-      )}
+      <div className="md:container">
+        {/* TODO: Switch back after holidays */}
+        {/* <Jumbotron data={jumbotron} /> */}
+        <HolidayReleaseJumbotron />
+      </div>
```
vojtaholik, line 22: this is sloppy but I didn't want to create another flag just for this. I also considered using Jumbotron component but that would require a lot of changes which would most likely remain unused apart of this use case.

joelhooks (approved):

PR #1020, opened 2021-12-23
Run pro status check for auth'd user on pricing

jbranchaud:
We are seeing users from time to time that land on the pricing page not
realizing they still have remaining time on their cancelled
subscription, who then try to subscribe anew. They end up on the Stripe
Checkout Session and are able to create a new subscription. We should
instead halt them at the pricing page and direct them to support for any
adjustments they need to make.

This PR adds the Pro Status Check for signed in users on the pricing widget so that they don't accidentally purchase a second subscription. It also uses vercel middleware to prevent an already subscribed user from landing on pricing. Instead it shuffles them over to the login page or their user page. It hides the pricing link from the footer as well.

https://user-images.githubusercontent.com/694063/147281430-7736cc51-d3b5-47ea-b5a2-8e8b68df65c3.mp4

pro access

### DIFF
--- a/src/components/pricing/pricing-widget/index.tsx
+++ b/src/components/pricing/pricing-widget/index.tsx
@@ -12,6 +12,8 @@ import SelectPlanNew from 'components/pricing/select-plan-new'
 import PoweredByStripe from 'components/pricing/powered-by-stripe'
 import ParityCouponMessage from 'components/pricing/parity-coupon-message'
 import isEmpty from 'lodash/isEmpty'
+import axios from 'axios'
+import toast from 'react-hot-toast'
 
 const PricingWidget: FunctionComponent<{}> = () => {
   const {viewer, authToken} = useViewer()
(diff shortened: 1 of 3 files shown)


src/components/pricing/pricing-widget/index.tsx
```diff
@@ -71,6 +73,26 @@ const PricingWidget: FunctionComponent<{}> = () => {
     })
 
     if (emailIsValid(viewer?.email)) {
+      const {hasProAccess} = await axios
```
joelhooks, line 76: one. i e to have would be to move calls like this into lib/users versus raw axios calls. same thing ultimately just jive since we probably make this call in multiple places

joelhooks: here’s another spot where we are making this check. on the pricing page we also ask for an email immediately and check what they enter there. 

https://github.com/eggheadio/egghead-next/blob/f84a9c2219f8e11d01a80da759d54385b79345c1/src/pages/pricing/email/index.tsx#L44

are the “widgets” not doing that prior to this pr?

(there are three spots in the app that make this call)

src/components/pricing/pricing-widget/index.tsx
```diff
@@ -71,6 +73,26 @@ const PricingWidget: FunctionComponent<{}> = () => {
...
+          duration: 6000,
+          icon: '❌',
+        })
+
+        // email is already associated with a pro account, return early instead
+        // of sending the user to a Stripe Checkout Session.
+        return
+      }
+
+      // the user doesn't have pro access, proceed to checkout
```
joelhooks, line 95: the line below was already here but track calls shouldn’t be blocking in general but i’m not looking at all the context so might be required here

joelhooks:
ok, so I looked at this on an actual non-pocket computer and think that visiting the pricing page for a logged in user we already know if they are `pro` or not (it's on the viewer object) and should just redirect them to their membership page instead of letting them click the button at all and drop a toast over there.

feels like this check was in place at some point but was lost in the shuffle

another approach is to just disable the button or replace the widget with a "you are already a member" block instead?

I think this status check can be more eager and should require a surprise. just send them to their profile/membership isntead

src/components/pricing/pricing-widget/index.tsx
jbranchaud: > move calls like this into lib/users versus raw axios calls.

for sure, I want to make another pass that consolidates these pro access checks and moves the axios call into `/lib`. we've "favored duplication over the wrong abstraction" enough and I think we can come up with something shared now.

joelhooks (approved):

PR #1025, opened 2022-01-05
garden hp

vojtaholik:
This PR is adding all courses released during holiday promo into their corresponding sections. It also brings back our jumbotron component highlighting last published course and moves CTA for [20 days](https://egghead.io/20-days) at the bottom of the page. I also added some more logic for handling cards grid based on number of resources in order to get them aligned nicely.

screenshot

gardening


zacjones93 (commented):
This looks great! The dynamic font size looks a little odd on mobile with the variety of course title lengths. That's a hard problem to fix. Is there a way we can have the titles all scale together?

Build the portfolio you need to be a badass web developer   egghe

zacjones93 (approved):

PR #1029, opened 2022-01-12
notes: include image

vojtaholik:
this adds image data point to video notes. related to https://github.com/skillrecordings/products/pull/150, won't have any visible effect at the moment.

new profile pic


src/pages/api/github-load-notes.ts
```diff
@@ -109,6 +113,7 @@ export function parseMdxNotesFile(text: string) {
       return {
         text: contents.trim(),
         type: 'staff',
+        image: eggheadLogo,
```
vojtaholik, line 116: for staff notes we default to egghead logo
image

DrShpongle (approved):
👍

PR #1032, opened 2022-01-19
Lm/remove pricing in footer

lsminter:
Joel removed pricing if they don't need to see it and it probably caused an index off by one error. I'm removing the Pricing path to hopefully fix this issue.


src/components/app/footer.tsx
```diff
@@ -37,11 +37,6 @@ const content = [
 ]
 
 const about = [
-  {
-    path: '/pricing',
-    label: 'Pricing',
-    nonMemberRequired: true,
-  },
```
lsminter, line 44: Going to add this back. We want this, but we need to make sure it shows up only for users that NEED it.

lsminter:
I believe the issue comes from this PR: https://github.com/eggheadio/egghead-next/pull/1020.

src/components/app/footer.tsx
joelhooks: this is not the appropriate solution which is found in how the links are displayed and not that pricing exists at all. The solution can be found closer to the jsx. Please do not remove links to our path to purchase.

joelhooks: https://github.com/eggheadio/egghead-next/blob/main/src/components/app/footer.tsx#L106-L131

the solution is in here or the method doing the filtering, not the data

lsminter:
Issue was fixed in https://github.com/eggheadio/egghead-next/pull/1038

PR #1041, opened 2022-02-02
update subscription status text color for dark mode on user page

zacjones93:
We keep the subscription status background the same on dark and light mode but the text changes

## Before
Image 2022-02-02 at 11 24 44 AM

## After
Image 2022-02-02 at 11 24 28 AM

tiny

### DIFF
--- a/src/components/users/subscription-details.tsx
+++ b/src/components/users/subscription-details.tsx
@@ -43,11 +43,11 @@ const SubscriptionDetails: React.FunctionComponent<SubscriptionDetailsProps> =
         {!loading && subscriptionData && (
           <div className="sm:px-6 lg:px-0 lg:col-span-9">
             <section className="mb-32">
-              <div className="p-4 w-full">
-                <div className="border border-accents-1	w-full p rounded-md m-auto my-8 max-w-max-content">
+              <div className="w-full p-4">
+                <div className="w-full m-auto my-8 border rounded-md border-accents-1 p max-w-max-content">
                   {subscriptionName ? (
                     <div className="px-5 py-4">
-                      <h3 className="text-2xl mb-1 font-medium">
+                      <h3 className="mb-1 text-2xl font-medium">
                         ⭐️ You're an <strong>egghead Member!</strong>
                       </h3>
                       <p className="text-accents-5">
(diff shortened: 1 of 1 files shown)


src/components/users/subscription-details.tsx
```diff
@@ -43,11 +43,11 @@ const SubscriptionDetails: React.FunctionComponent<SubscriptionDetailsProps> =
         {!loading && subscriptionData && (
           <div className="sm:px-6 lg:px-0 lg:col-span-9">
             <section className="mb-32">
-              <div className="p-4 w-full">
-                <div className="border border-accents-1	w-full p rounded-md m-auto my-8 max-w-max-content">
+              <div className="w-full p-4">
+                <div className="w-full m-auto my-8 border rounded-md border-accents-1 p max-w-max-content">
```
zacjones93, line 47: @DrShpongle do I need to adjust my tailwind extension? Almost every file I touch my extension sorts the classes. Not sure if it's correct

src/components/users/subscription-details.tsx
```diff
@@ -68,7 +68,7 @@ const SubscriptionDetails: React.FunctionComponent<SubscriptionDetailsProps> =
                             </div>
                             {subscriptionData?.subscription
                               ?.cancel_at_period_end && (
-                              <div className="rounded text-xs px-2 py-1 flex justify-center items-center bg-gray-100">
+                              <div className="flex items-center justify-center px-2 py-1 text-xs bg-gray-100 rounded text-gray-1000">
```
zacjones93, line 71: adds `text-gray-1000` to the label

src/components/users/subscription-details.tsx
joelhooks: or do we need to run it once on the entire code base?

vojtaholik: I think once and with [tailwind's new prettier plugin](https://tailwindcss.com/blog/automatic-class-sorting-with-prettier).

src/components/users/subscription-details.tsx
DrShpongle, line 71:
```suggestion
                              <div className="flex items-center justify-center px-2 py-1 text-xs bg-gray-100 rounded text-gray-1000 mt-2">
```

DrShpongle (approved):

PR #1042, opened 2022-02-02
Add script to index Sanity articles with Algolia

jbranchaud:
This is a spike and could use input from others on whether this is how we want to approach this.

This script is something that can be manually run or added to the build to ensure that Sanity articles get synced and indexed with Algolia.

This initial spike is only grabbing a subset of the articles and writing them to `content_staging` instead of `content_production`.

CleanShot 2022-02-02 at 16 59 20@2x

indexing


index-sanity-articles.js
```diff
@@ -0,0 +1,97 @@
...
+   * in Algolia.
+   *
+   * Removing `'body': body,` from the GROQ query for now because it is
+   * exceeding Algolia record size limits of 20kb.
+   */
+
+  // TODO: notice this is only grabbing the first 3 article at the moment in
+  // order to test it out. Eventually those second set of square brackets
+  // should be empty so that ALL articles are pulled in.
+  let articlesQuery = groq`*[_type == "post" && publishedAt < now() && !(_id in path("drafts.**"))][0..3]{
```
jbranchaud, line 44: This is still a spike, so I'm only grabbing the first 3 article while iterating on it.

index-sanity-articles.js
```diff
@@ -0,0 +1,97 @@
...
+  const fullTextSearch = {
+    appId: process.env.NEXT_PUBLIC_ALGOLIA_APP || '',
+    searchApiKey: process.env.NEXT_PUBLIC_ALGOLIA_KEY || '',
+  }
+
+  // TODO: Note that this is referencing the `content_staging` Algolia index.
+  // Once this is ready for prime-time, it should instead write to
+  // `content_production`.
+  const ALGOLIA_INDEX_NAME = 'content_staging'
+  // process.env.NEXT_PUBLIC_ALGOLIA_INDEX_NAME || 'content_production'
```
jbranchaud, line 81: This will eventually need to be swapped back to the production index.

index-sanity-articles.js
```diff
@@ -0,0 +1,97 @@
...
+  )
+
+  const index = searchClient.initIndex(ALGOLIA_INDEX_NAME)
+
+  index
+    .saveObjects(articles)
+    .then(({objectIDs}) => {
+      console.log(objectIDs)
+    })
+    .catch((e) => console.log(e))
```
jbranchaud, line 95: I'm not yet sure if `saveObjects` is the right function to use here long-term. Since there are few enough articles, it may be easiest to just stick with it. I think the tradeoff is that it is aggressive and may use more of the allotted API quota than is necessary. Again, that may be negligible for the small number of articles we are dealing with.

index-sanity-articles.js
```diff
@@ -0,0 +1,97 @@
...
+   *     ...
+   *   },
+   *
+   * Also, ignore `categories` for now, they are one of two values (Technical
+   * Case Study and Instructor Case Study) and I don't think they will be useful
+   * in Algolia.
+   *
+   * Removing `'body': body,` from the GROQ query for now because it is
+   * exceeding Algolia record size limits of 20kb.
+   */
```
jbranchaud, line 39: Here are some notes about attributes on articles that are available from Sanity, but that are not incorporated into the GROQ query and why.

joelhooks (approved):

PR #1043, opened 2022-02-07
fix: correct typos in "Writing for egghead" page

arthurdenner:
I was reading the page and noticed a few small things.


zacjones93 (approved):
Thanks for the PR @arthurdenner! 🎉

PR #1046, opened 2022-02-08
[WIP] Lesson upload form

Creeland:
// Currently, onFinish callback in the Upload component is creating a clojure with formik.values.lessons.
// So when that state is updated, it's spreading an empty array with each update. Causing only the latest upload to be saved

This whole PR is still a WIP as of https://github.com/eggheadio/egghead-next/pull/1046/commits/8e6e997cb10d0fe6376651743d0fbd8be9f8bd3c

It is very close to an end-to-end proof of concept. The `lesson` document type isn't showing up in Sanity Studio yet. Once it is, we should be able to tweak and test what we have here. Upload a video, edit the title, and send directly to Sanity.

lessons

### DIFF
--- a/src/pages/api/sanity/lessons/create.ts
+++ b/src/pages/api/sanity/lessons/create.ts
@@ -1,0 +1,97 @@
+import axios from 'axios'
+
+import {NextApiRequest, NextApiResponse} from 'next'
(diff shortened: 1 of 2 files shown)


Creeland:
Commits cherry picked into #1051

PR #1047, opened 2022-02-09
Add Podcast seasons and episodes to the schema

jbranchaud:
In collaboration with @zacjones93 

@zacjones93, look this over and let me know what you think. You can also boot up Sanity Studio locally to see how it looks "live".

This adds the concept of a _Podcast Episode_ to Sanity Studio as a transfer target for the podcasts that currently live in `egghead-rails`. It also adds _Podcast Season_ which is a way of organizing a group of episodes into a bucket with additional metadata.

podcast

### DIFF
--- a/studio/schemas/documents/podcastEpisode.js
+++ b/studio/schemas/documents/podcastEpisode.js
@@ -1,0 +1,145 @@
+/*
+ * Here is an example podcast as stored in egghead-rails:
+
+#<Podcast:0x00007f805f2e74a8
+ id: 390,
+ title: "Accessible Cross-Browser CSS Form Styling with Stephanie Eckles",
+ summary: "Taylor and Stephanie talk about accessibility, the sticking points, and Stephanie's experience with learning about forms. ",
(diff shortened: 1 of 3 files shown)


studio/schemas/documents/podcastEpisode.js
```diff
@@ -0,0 +1,145 @@
+/*
+ * Here is an example podcast as stored in egghead-rails:
+
+#<Podcast:0x00007f805f2e74a8
```
jbranchaud, line 4: We can delete this when we're done, just wanted it as a reference point for what values we may want to include in the migration.

studio/schemas/documents/podcastEpisode.js
```diff
@@ -0,0 +1,145 @@
...
+ square_cover_file_size: nil,
+ square_cover_updated_at: nil,
+ square_cover_processing: nil,
+ show_guid: "11504d19-d225-48ff-9ce0-ef8f15d8ec54",
+ row_order: 8388607>
+
+ Questions:
+ - do we need to represent transcripts in this Sanity data model?
+ - do we care about transferring over the `show_guid`?
+ - should the `id` be included as a reference back to egghead-rails DB?
```
jbranchaud, line 37: A couple questions about values that can be migrated

joelhooks (approved):

PR #1055, opened 2022-02-23
Add is_publisher to viewer type

jbranchaud:
This boolean value is now returned by the API and used in the client.

publisher

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -74,6 +74,7 @@ export type Viewer = {
   name: string
   is_pro: boolean
   is_instructor: boolean
+  is_publisher: boolean
   can_comment: boolean
   created_at: number
   discord_id: string


joelhooks:
we’ve traditionally avoided using boolean attribute flags on our user types for access relying instead on the availability of api urls that facilitate the activity itself. so the presence of “upload video url” or “publish lesson url” would be the property that we use to determine/show UI for that activity

this prevents the runaway use of boolean flags for access AND allows us to avoid URL construction in our client app. 

Is it possible to reconsider this approach and including these URLs on the viewer were available to enable this process?

joelhooks:
i propose this be removed in favor of using the roles and CASL here and on the api side as well

joelhooks:
I appreciate your response but these boolean checks are a slippery slope of “the ui just needs to…” and you end up with every role being represented and a :dumpster-fire: of RBAC (role based access control)

ABAC is attribute-based and a lot more flexible and we can avoid this sort of stacking of bools

technically speaking the server is sending viewer.roles array now too so we can have isPublisher etc if really needed but should probably use CASL (and maybe @casl/react) for this sort of thing going forward

PR #1056, opened 2022-02-24
feat(progress): spikes a quick look at loading course progress

joelhooks:
this is a quick look at using a caching loader library like react-query to fetch course progress.

only focused on the data loading aspect and we'd probably want to bake it into the card layer since it already understands `resource` but even this starts to be useful on many levels imo

I didn't check if use is logged in or not with this, but logging in on the preview deploy should make it function. react-query has a lot of cool features like suspense support, caching, loading state, etc...

Resources for Web Developers in 2022 egghead io 2022-02-23 at 20 34 09

progress

### DIFF
--- a/package.json
+++ b/package.json
@@ -119,6 +119,7 @@
     "react-hot-toast": "^2.1.0",
     "react-instantsearch-dom": "^6.12.1",
     "react-markdown": "^5.0.3",
+    "react-query": "^3.34.16",
     "react-s3-uploader": "^5.0.0",
     "react-scroll": "^1.8.3",
     "react-textfit": "^1.1.1",
--- a/src/components/card/search-hit-resource-card.tsx
+++ b/src/components/card/search-hit-resource-card.tsx
@@ -16,6 +16,8 @@ import {CardResource} from 'types'
 import {Textfit} from 'react-textfit'
 import ReactMarkdown from 'react-markdown'
 import useFitText from 'use-fit-text'
+import {useQuery} from 'react-query'
+import axios from '../../utils/configured-axios'
 
 const SearchHitResourceCard: React.FC<{
   resource: CardResource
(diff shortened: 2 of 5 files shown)


src/components/card/search-hit-resource-card.tsx
```diff
@@ -16,6 +16,8 @@ import {CardResource} from 'types'
 import {Textfit} from 'react-textfit'
 import ReactMarkdown from 'react-markdown'
 import useFitText from 'use-fit-text'
+import {useQuery} from 'react-query'
+import axios from '../../utils/configured-axios'
```
joelhooks, line 20: this looks for the bearer token

src/components/card/search-hit-resource-card.tsx
```diff
@@ -33,11 +35,32 @@ const SearchHitResourceCard: React.FC<{
...
+            ),
+          )
+        : {data: false}
+      return result.data?.progress
+    },
+    {
+      retry: false,
+    },
+  )
+
```
joelhooks, line 55: clunky, but works

it's particularly interesting that if you use the same key it caches very nicely so this would probably fit best in a `useResourceProgress` hook that could manage all the edge cases and such

package.json
```diff
@@ -119,6 +119,7 @@
     "react-hot-toast": "^2.1.0",
     "react-instantsearch-dom": "^6.12.1",
     "react-markdown": "^5.0.3",
+    "react-query": "^3.34.16",
```
jbranchaud, line 122: I don't know much about `react-query` or `swr`, but they seem like they are alternatives to one another. Any reason to pull in `react-query` here when `swr` is already available?

joelhooks: the caching “just works” and it’s more robust than swr in general.  we should be using it across the board imo. 

i wouldn’t really care either way as long as the caching was effective, but i went for what was simplest and i knew worked instead of trying to figure out if what was already there could do the job mostly for the sake of time. 

i plan on closing this PR not merging it.

jbranchaud: 👍 I missed that this was a proof of concept

PR #1060, opened 2022-03-04
Remove all ahoy tracking and its npm package

jbranchaud:
We are looking to remove Ahoy event tracking from the backend. This is a preliminary step that removes the uses of those endpoints from the client.

ahoy


joelhooks (approved):
🚮

PR #1061, opened 2022-03-08
feat: sign urls on egghead-next

joelhooks:
sign


src/pages/api/aws/sign-s3.ts
```diff
@@ -0,0 +1,49 @@
+import {NextApiRequest, NextApiResponse} from 'next'
+import AWS from 'aws-sdk'
```
joelhooks, line 2: this is the v2 of the aws sdk because the modern v3 "modular" sdk documentation made me very sad and I couldn't get this flow figured out

src/pages/api/aws/sign-s3.ts
```diff
@@ -0,0 +1,49 @@
...
+
+const signedUrl = async (req: NextApiRequest, res: NextApiResponse) => {
+  if (req.method === 'GET') {
+    const {objectName, contentType} = req.query
+
+    const filename = `${uuidv4()}/${objectName}`
+
+    const params = {
+      Bucket: options.bucket,
+      Key: `rip/${filename}`,
```
joelhooks, line 29:
```suggestion
      Key: `${process.env.NODE_ENV}/${filename}`,
```

src/pages/upload/index.tsx
```diff
@@ -41,7 +41,7 @@ const Upload: React.FC = () => {
         //if we set this to `false` we can list all the files and
         //call `uploaderRef.current.uploadFile()` when we are ready
         autoUpload={true}
-        signingUrl={viewer.s3_signing_url}
+        signingUrl={`/api/aws/sign-s3`}
```
joelhooks, line 44: this could be moved over to the API in `egghead-rails` instead of hard coded here

jbranchaud: What would be the benefit of having `egghead-rails` provide the signing route?

One drawback I see is that if we ever wanted to change the route, we'd have to deploy changes to two systems.

src/pages/api/aws/sign-s3.ts
```diff
@@ -0,0 +1,49 @@
...
+  accessKeyId: process.env.AWS_VIDEO_UPLOAD_ACCESS_KEY_ID || '',
+  secretAccessKey: process.env.AWS_VIDEO_UPLOAD_SECRET_ACCESS_KEY || '',
+}
+
+AWS.config.credentials = credentials
+
+const s3 = new AWS.S3(options)
+
+const signedUrl = async (req: NextApiRequest, res: NextApiResponse) => {
+  if (req.method === 'GET') {
```
jbranchaud, line 22: Any permissions checking we want to do here? Otherwise this endpoint is wide open to anyone who wants to upload to the S3 bucket.

jbranchaud (approved):
I left a question about permission checking that we can deal with here or in a follow up.

This looks great. It's nice that it just drops in in place of the `egghead-rails` approach.

src/pages/api/aws/sign-s3.ts
joelhooks: probably, I don't know exactly how that'd be accomplished so I just left it. feels like we are inching towards the "big sack of booleans" for permissions, but outside of giving people ideas here in this channel, I doubt the risk is significant because there is literally nothing to gain from shoving stuff into this bucket that I can tell.

https://www.npmjs.com/package/accesscontrol

https://dev.to/noclat/unify-acl-across-front-end-and-back-end-with-next-js-and-nextauth-js-5ban

This looks way better than the "is role" style to me and something we should probably consider.

PR #1063, opened 2022-03-11
Fix role checking logic to compare multiple roles

jbranchaud:
JavaScript's `includes` cannot compare multiple values on the right side
in the way we were trying to do.

```javascript
['user', 'publisher'].includes('editor', 'publisher')
//=> false
```

Instead, I've replaced that with a small utility function `hasRoles`
that handles the logic of comparing a single role or multiple roles to
the user's list of roles.

```javascript
hasRoles(user, ['editor', 'publisher'])
//=> true
```

ability check

### DIFF
--- a/src/server/ability.ts
+++ b/src/server/ability.ts
@@ -1,5 +1,6 @@
 import {AbilityBuilder, Ability} from '@casl/ability'
 import {loadCurrentUser} from 'lib/users'
+import {intersection, isString} from 'lodash'
 
 type Actions = 'manage' | 'upload'
 type Subjects = 'Video' | 'all'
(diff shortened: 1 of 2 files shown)


src/server/ability.ts
```diff
@@ -14,13 +15,28 @@ function defineAbilityFor(user: any) {
   // AbilityBuilder https://casl.js.org/v5/en/guide/define-rules#ability-builder-class
   const {can, build} = new AbilityBuilder<AppAbility>(Ability)
 
-  if (user.roles?.includes('admin')) {
+  if (hasRoles(user, 'admin')) {
```
joelhooks, line 18: `_.includes([1, 2, 3], 1, 2);` lo dash includes works like you’d expect and is probably the way to go IMO

jbranchaud: I forgot that lodash has an `includes`. It's not quite what I want though. That third argument is to specify an index.
CleanShot 2022-03-10 at 18 07 36@2x

I wanted to improve on the readability as well by not having to do:

```javascript
if (includes(user?.roles, 'publisher') || includes(user?.roles, 'editor')) {
  ...
}
```

The multi-role checks get kinda cumbersome to read imo.

joelhooks (approved):

PR #1064, opened 2022-03-11
feat(auth): use the viewer endpoint to load roles array

joelhooks:
rolls

### DIFF
--- a/src/lib/viewer.ts
+++ b/src/lib/viewer.ts
@@ -1,0 +1,16 @@
+import {getGraphQLClient} from '../utils/configured-graphql-client'
+import gql from 'graphql-tag'
+
+export async function loadCurrentViewerRoles(token: string) {
+  const query = gql`
+    query viewer {
+      viewer {
+        roles
+      }
+    }
+  `
+  const graphQLClient = getGraphQLClient(token)
+  const {viewer} = await graphQLClient.request(query)
+
+  return viewer ? viewer.roles : []
+}
--- a/src/server/ability.ts
+++ b/src/server/ability.ts
@@ -1,42 +1,42 @@
 import {AbilityBuilder, Ability} from '@casl/ability'
-import {loadCurrentUser} from 'lib/users'
 import {intersection, isString} from 'lodash'
+import {loadCurrentViewerRoles} from '../lib/viewer'
 
 type Actions = 'manage' | 'upload'
 type Subjects = 'Video' | 'all'
+export type Roles = 'admin' | 'editor' | 'publisher'
 type AppAbility = Ability<[Actions, Subjects]>
 
 export async function getAbilityFromToken(token: string) {
-  const user = await loadCurrentUser(token)
-  return defineAbilityFor(user)
+  const viewerRoles = await loadCurrentViewerRoles(token)
+  return defineAbilityFor(viewerRoles)
 }
 
-function defineAbilityFor(user: any) {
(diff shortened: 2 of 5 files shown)


src/lib/viewer.ts
```diff
@@ -0,0 +1,16 @@
+import {getGraphQLClient} from '../utils/configured-graphql-client'
+import gql from 'graphql-tag'
+
+export async function loadCurrentViewerRoles(token: string) {
+  const query = gql`
+    query viewer {
+      viewer {
```
joelhooks, line 7: this is an alias server side in gql for current user and generally in egghead-next we've tried to use "viewer" terminology over "user" with varied levels of success

the idea being that a "viewer" isn't necessarily the user but could be support viewing as the user, as a for instance

that's why we have a ViewerProvider etc

joelhooks: Figured gql just grabbing the roles array would be quicker than jbuilder any day of the week

src/server/ability.ts
```diff
@@ -1,42 +1,42 @@
 import {AbilityBuilder, Ability} from '@casl/ability'
-import {loadCurrentUser} from 'lib/users'
 import {intersection, isString} from 'lodash'
+import {loadCurrentViewerRoles} from '../lib/viewer'
 
 type Actions = 'manage' | 'upload'
 type Subjects = 'Video' | 'all'
+export type Roles = 'admin' | 'editor' | 'publisher'
```
joelhooks, line 7: this gives type checking which is nice, only drawback was having to cast in the tests

src/server/ability.ts
```diff
@@ -1,42 +1,42 @@
...
-  const user = await loadCurrentUser(token)
-  return defineAbilityFor(user)
+  const viewerRoles = await loadCurrentViewerRoles(token)
+  return defineAbilityFor(viewerRoles)
 }
 
-function defineAbilityFor(user: any) {
-  // AbilityBuilder https://casl.js.org/v5/en/guide/define-rules#ability-builder-class
+/**
+ * @see {@link https://casl.js.org/v5/en/guide/define-rules#ability-builder-class|AbilityBuilder}
```
joelhooks, line 16: jsdoc is pretty sweet

you can actually replace TS with it which is interesting

src/server/ability.ts
```diff
@@ -1,42 +1,42 @@
...
+export function includesRoles(
+  roles: Roles[] = [],
+  rolesToCheck: Roles | Roles[] = [],
+) {
   if (isString(rolesToCheck)) {
-    rolesToCheckArray = [rolesToCheck]
-  } else {
-    rolesToCheckArray = rolesToCheck
+    rolesToCheck = [rolesToCheck]
   }
```
joelhooks, line 39: there are some various arguments (ha) for not overwriting like this, but in this case the extremely marginal risk seems fine

src/server/ability.ts
```diff
@@ -1,42 +1,42 @@
...
     can('upload', 'Video')
   }
 
   return build()
 }
 
-export function hasRoles(user: any, rolesToCheck: string | string[]) {
-  // ensure rolesToCheck is an array
-  let rolesToCheckArray: string[]
+export function includesRoles(
```
joelhooks, line 33: I went with includes because it's similar to includes and since it's typed now it's pretty true

jbranchaud: I like this name better 💯

jbranchaud (approved):

PR #1065, opened 2022-03-11
fix(player): stop double load

vojtaholik:
fix it


src/pages/lessons/[slug].tsx
```diff
@@ -354,10 +354,6 @@ const Lesson: React.FC<LessonProps> = ({
 
     switch (currentLessonState) {
       case 'loaded':
-        videoService.send({
-          type: 'LOAD_RESOURCE',
-          resource: lesson,
-        })
```
Comment (vojtaholik) [inline src/pages/lessons/[slug].tsx:360]: this was being called twice

PR #1066, opened 2022-03-11
fix(player): prevent double loading

vojtaholik:
fix

### DIFF
--- a/src/pages/lessons/[slug].tsx
+++ b/src/pages/lessons/[slug].tsx
@@ -155,7 +155,7 @@ export const getServerSideProps: GetServerSideProps = async function ({
 
 type LessonProps = {
   state: any
-  initialLesson: LessonResource
+  initialLesson: VideoResource
   watchCount: number
   setWatchCount: (value: number) => void
 }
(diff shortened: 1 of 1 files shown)


src/pages/lessons/[slug].tsx
```diff
@@ -397,14 +396,16 @@ const Lesson: React.FC<LessonProps> = ({
         console.debug(
           `changed to viewing isFullscreen: ${isFullscreen} mediaPresent: ${mediaPresent}`,
         )
+        if (!initialMediaPresent && mediaPresent) {
+          videoService.send({
+            type: 'LOAD_RESOURCE',
+            resource: lesson,
+          })
+        }
```
Comment (vojtaholik) [inline src/pages/lessons/[slug].tsx:404]: only trigger LOAD_RESOURCE if the initial lesson doesn't include media urls and the full lesson does.

src/pages/lessons/[slug].tsx
```diff
@@ -567,11 +568,11 @@ const Lesson: React.FC<LessonProps> = ({
           >
             <div className={cx({hidden: !playerVisible})}>
               <Player
-                poster={lesson.thumb_url}
                 canAddNotes={isEmpty(viewer) ? false : !isFullscreen}
                 className="font-sans"
                 container={fullscreenWrapperRef.current || undefined}
                 controls={<DownloadControl lesson={lesson} />}
+                // poster={lesson.thumb_url}
```
Comment (vojtaholik) [inline src/pages/lessons/[slug].tsx:575]: if the thumbnail image was the first frame of the video then it would work, but this felt clunky.

Comment (joelhooks) [reply src/pages/lessons/[slug].tsx:575]: what would be cool is an og image style generator that gave a stylish consistent poster frame

joelhooks (approved):

PR #1068, opened 2022-03-14
Include Ability object on viewer context

jbranchaud:
Then use that ability in a React component to check if the current
viewer has authorization to use the `/upload` endpoint.

There is no need to pull in `@casl/react` at this point. It is easier to include the `ability` object on the `viewer` context. The `viewer` context is already broadly available, so these ability checks can happen in any component now.

john malkovich

### DIFF
--- a/src/context/viewer-context.tsx
+++ b/src/context/viewer-context.tsx
@@ -211,7 +211,7 @@ function useAuthedViewer() {
       refreshUser: auth.refreshUser,
       ability,
     }),
-    [viewer, loading],
+    [viewer, loading, authToken],
   )
 
   return values
--- a/src/context/viewer-context.tsx
+++ b/src/context/viewer-context.tsx
@@ -7,6 +7,7 @@ import isEmpty from 'lodash/isEmpty'
 import getAccessTokenFromCookie from '../utils/get-access-token-from-cookie'
 import useTokenSigner from '../hooks/use-token-signer'
 import useAffiliateAssigner from '../hooks/use-affiliate-assigner'
+import {getAbilityFromToken, canDoNothingAbility} from 'server/ability'
 
 export const auth = new Auth()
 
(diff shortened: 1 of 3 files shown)


src/pages/upload/index.tsx
```diff
@@ -30,10 +30,10 @@ const fileUploadReducer = (state: any, action: any) => {
 
 const Upload: React.FC = () => {
   const [state, dispatch] = React.useReducer(fileUploadReducer, {files: []})
-  const {viewer} = useViewer()
+  const {ability} = useViewer()
   const uploaderRef = React.useRef(null)
 
-  return viewer?.s3_signing_url ? (
+  return ability.can('upload', 'Video') ? (
```
jbranchaud, line 36: I like how this is coming together. Moving from an indirect presence check of a URL to an ability check of the action taken on this page. This makes it much clearer what is going on.

src/context/viewer-context.tsx
```diff
@@ -179,6 +183,18 @@ function useAuthedViewer() {
...
+    const fetchAbility = async (authToken: string) => {
+      const ability = await getAbilityFromToken(authToken)
+      setAbility(ability)
+    }
+
+    if (authToken) {
+      fetchAbility(authToken)
+    }
+  }, [authToken])
+
```
jbranchaud, line 197: This uses `src/server/ability.ts` to build the `ability` object. That object is then included on the `viewer` context.

src/context/viewer-context.tsx
```diff
@@ -189,10 +205,11 @@ function useAuthedViewer() {
...
       handleAccessTokenAuthentication: handleAccessTokenAuthentication,
       isAuthenticated: () => auth.isAuthenticated(),
-      authToken: auth.getAuthToken(),
+      authToken: authToken,
       requestSignInEmail: (email: any) => auth.requestSignInEmail(email),
       loading,
       refreshUser: auth.refreshUser,
+      ability,
     }),
     [viewer, loading],
```
joelhooks, line 214: you'll likely want to add the authToken to the deps array so that it will update the memoized value when/if it changes for various reasons

joelhooks (approved):

PR #1075, opened 2022-03-16
Proof of concept for an instructor dropdown on the Course Builder page

jbranchaud:
- Load instructors for assigning to course upload
- Add specialized role for selecting instructor

We still separately need to get instructor IDs synced in Sanity. This PR helps highlight the user-facing need for those IDs to be available.

https://user-images.githubusercontent.com/694063/158635825-68fbd139-a808-4c14-b3fe-75e5e2dedcc2.mp4

course upload

### DIFF
--- a/src/pages/upload/index.tsx
+++ b/src/pages/upload/index.tsx
@@ -6,17 +6,42 @@ import fileExtension from 'file-extension'
 import {find} from 'lodash'
 import {GetServerSideProps} from 'next'
 import {getAbilityFromToken} from 'server/ability'
+import groq from 'groq'
+import {sanityClient} from 'utils/sanity-client'
 
 const SIGNING_URL = `/api/aws/sign-s3`
 
 type FileUpload = {file: File; percent: number; message: string}
 
+type Instructor = {
+  externalId: string
+  person: {
+    _id: string
+    name: string
+    image: string
+  }
+}
+
 export const getServerSideProps: GetServerSideProps = async function ({req}) {
   const ability = await getAbilityFromToken(req.cookies[ACCESS_TOKEN_KEY])
 
   if (ability.can('upload', 'Video')) {
+    const instructorQuery = groq`
+      *[_type == 'collaborator' && role == 'instructor'][]{
+        'person': person-> {
+            _id,
+            name,
+            'image': image.url,
+          },
+        externalId,
+      }`
+
+    const instructors: Instructor[] = await sanityClient.fetch(instructorQuery)
+
     return {
-      props: {},
+      props: {
+        instructors,
+      },
     }
   } else {
     return {
@@ -48,7 +73,7 @@ const fileUploadReducer = (state: any, action: any) => {
   }
 }
 
-const Upload: React.FC = () => {
+const Upload: React.FC<{instructors: Instructor[]}> = ({instructors}) => {
   const [state, dispatch] = React.useReducer(fileUploadReducer, {files: []})
   const uploaderRef = React.useRef(null)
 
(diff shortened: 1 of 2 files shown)


src/pages/upload/index.tsx
```diff
@@ -6,17 +6,43 @@ import fileExtension from 'file-extension'
...
+      *[_type == 'collaborator' && role == 'instructor'][]{
+        'person': person-> {
+            _id,
+            name,
+            'image': image.url,
+          },
+        externalId,
+      }`
+
+    const instructors: Instructor[] = await sanityClient.fetch(instructorQuery)
```
jbranchaud, line 40: Load collaborators that are instructors from Sanity and pass into `/upload` as a prop.

src/pages/upload/index.tsx
```diff
@@ -102,6 +129,25 @@ const Upload: React.FC = () => {
           </div>
         )
       })}
+      {ability.can('upload', 'Video.WithInstructor') ? (
+        <>
+          <h2>Insructor</h2>
+          {/* we can use a more featureful select component here that allows for search and displaying an image thumbnail. This is a proof of concept. */}
+          <select>
+            {instructors.map(
```
jbranchaud, line 137: Allow editor/publisher to select the instructor for this Course.

I'm dropping in a plain old `<select>` at the moment, but I think it'd be more useful to move to something more featureful that allows for search and displaying an avatar.

src/pages/upload/index.tsx
```diff
@@ -102,6 +129,25 @@ const Upload: React.FC = () => {
           </div>
         )
       })}
+      {ability.can('upload', 'Video.WithInstructor') ? (
```
jbranchaud, line 132: In addition to publishers/editors, eventually _instructors_ will be able to use this Course Builder. Those with only the `instructor` role won't be able to select the instructor, hence the ability check.

src/server/ability.ts
```diff
@@ -25,6 +25,9 @@ function defineAbilityFor(viewerRoles: Roles[]) {
 
   if (includesRoles(viewerRoles, ['editor', 'publisher'])) {
     can('upload', 'Video')
+    // note: an instructor wouldn't have this ability, so they'd only get the
+    // choice to upload as themselves.
+    can('upload', 'Video.WithInstructor')
```
jbranchaud, line 30: Not sure how best to approach the naming here.

Looking at it from the github PR page, it seems like `Video.WithInstructor` isn't really what I want here. This is a course builder, so maybe something like `can('assignInstructor', 'Course')` or something like that.

Thoughts?

joelhooks: for the most part we want to be thinking about this from a CRUD perspective, and complex namespaced rules is a slippery slope to hell ;)

so you'd have something like `can('update', 'Video', ['instructor_id'])` under the `editor` roll, otherwise the instructor dropdown doesn't appear and it'd be hard coded to the current user.

I'd probably just put the boolean in the `disabled` field and preselect the current instructor 

in the future you'd probably have something like `can('update', 'Video', ['title', 'description'], {instructor_id: user.instructor.id})` etc

generally with this group this is mostly cosmetic and we don't need to be ultra secure about enforcing this server side where it'd be a little more complex

This is referred to as [Restricting Field Access in the docs](https://casl.js.org/v5/en/guide/restricting-fields)

joelhooks (approved):
approving this, but I think it needs some adjustment on the ability front that can be in a follow up

src/server/ability.ts
jbranchaud: This is spot on. The attribute-level scoping is a way better way to do this kind of thing.

> generally with this group this is mostly cosmetic and we don't need to be ultra secure about enforcing this server side where it'd be a little more complex

I like that with the Next.js `/api` directory, we can use `ability` in both client-side and server-side contexts.

PR #1082, opened 2022-03-24
Refactor Software Lib naming to Topic

jbranchaud:
> “The discussions that you have [with your teammates] about naming have benefits far beyond the work that you are currently doing. They help you and your team develop a common view of what the system is and what it can become.”
> –Working Effectively with Legacy Code

This refactoring is motivated by my efforts to sync all egghead-rails tags into Sanity. We are also in the midst of thinking about and rethinking the schema. This feels like a good chance to reuse existing schema objects as tagging becomes a primary concern of Sanity.

The shift from `Software Libarary` to `Topic` is to be more encompassing
of what can go in this bucket and to better align with the `topic`
context that we've started using in egghead-rails. Things like `a11y`
and `JavaScript` aren't really software libraries. They are however
topics of the different kinds of content provided by egghead.

Then, instead of Versioned Software Libraries, it is `TopicTaggings`.
This name does a couple things in addition to aligning it with the
updated naming of `Topic`.
1) It drops the "Versioned" qualifier since that isn't required and
doesn't make sense for a topic like `a11y`.
2) It suggests that it is a concept to be applied to (or tagged to) a
piece of content.

Open to discussion on this, hence the RFC tag.

topical

### DIFF
--- a/src/lib/courses.ts
+++ b/src/lib/courses.ts
@@ -34,9 +34,9 @@ const courseQuery = groq`
     'avatar_url': person->image.url,
     'slug': person->slug.current
   },
-	'dependencies': softwareLibraries[]{
+	'dependencies': topicTaggings[]{
     version,
-    ...library->{
+    ...topic->{
       description,
       "slug": slug.current,
       path,
--- a/studio/schemas/documents/topic.js
+++ b/studio/schemas/documents/topic.js
@@ -2,8 +2,8 @@ import {MdLibraryBooks as icon} from 'react-icons/md'
 import React from 'react'
 
 export default {
-  name: 'software-library',
-  title: 'software library (dependency)',
+  name: 'topic',
+  title: 'Topic (Libraries/Frameworks/Languages/Etc.)',
   type: 'document',
   icon,
   fields: [
@@ -36,7 +36,7 @@ export default {
       type: 'string',
     },
     {
-      title: 'Link to library',
+      title: 'Link to canonical resource',
       name: 'url',
       type: 'url',
     },
--- a/studio/schemas/objects/topic-tagging.js
+++ b/studio/schemas/objects/topic-tagging.js
@@ -1,27 +1,27 @@
 import React from 'react'
 
 export default {
-  name: 'versioned-software-library',
-  title: 'Versioned Software Library (Dependency)',
+  name: 'topic-tagging',
+  title: 'Topic Tagging',
   type: 'object',
   fields: [
     {
       name: 'version',
-      title: 'Version',
+      title: 'Version (if applicable)',
       type: 'string',
     },
     {
-      name: 'library',
-      title: 'Software Library (Dependency)',
+      name: 'topic',
+      title: 'Topic (Libraries/Frameworks/Languages/Etc.)',
       type: 'reference',
-      to: [{type: 'software-library'}],
+      to: [{type: 'topic'}],
     },
   ],
   preview: {
     select: {
       version: 'version',
-      name: 'library.name',
-      media: 'library.image.url',
+      name: 'topic.name',
+      media: 'topic.image.url',
     },
     prepare(selection) {
       const {version, name, media} = selection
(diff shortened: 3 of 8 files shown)


studio/schemas/documents/topic.js
```diff
@@ -2,8 +2,8 @@ import {MdLibraryBooks as icon} from 'react-icons/md'
 import React from 'react'
 
 export default {
-  name: 'software-library',
-  title: 'software library (dependency)',
+  name: 'topic',
+  title: 'Topic (Libraries/Frameworks/Languages/Etc.)',
```
jbranchaud, line 6: Renaming `software-library` --> `topic`

studio/schemas/objects/topic-tagging.js
```diff
@@ -1,27 +1,27 @@
 import React from 'react'
 
 export default {
-  name: 'versioned-software-library',
-  title: 'Versioned Software Library (Dependency)',
+  name: 'topic-tagging',
+  title: 'Topic Tagging',
```
jbranchaud, line 5: Renaming `versioned-software-library` --> `topic-tagging`

studio/schemas/objects/topic-tagging.js
```diff
@@ -1,27 +1,27 @@
...
       type: 'string',
     },
     {
-      name: 'library',
-      title: 'Software Library (Dependency)',
+      name: 'topic',
+      title: 'Topic (Libraries/Frameworks/Languages/Etc.)',
       type: 'reference',
-      to: [{type: 'software-library'}],
+      to: [{type: 'topic'}],
```
jbranchaud, line 17: This continues to use what is now named `topic`

src/lib/courses.ts
```diff
@@ -34,9 +34,9 @@ const courseQuery = groq`
     'avatar_url': person->image.url,
     'slug': person->slug.current
   },
-	'dependencies': softwareLibraries[]{
+	'dependencies': topicTaggings[]{
     version,
-    ...library->{
+    ...topic->{
```
jbranchaud, line 39: This is the only `groq` query referencing `softwareLibraries`. Updating it to match the internal schema.

This query ends up calling the resulting array of data `dependencies` which is a little but of a misnomer for things like `a11y`, but it works.

src/lib/courses.ts
```diff
@@ -34,9 +34,9 @@ const courseQuery = groq`
     'avatar_url': person->image.url,
     'slug': person->slug.current
   },
-	'dependencies': softwareLibraries[]{
+	'dependencies': topicTaggings[]{
```
Creeland, line 37: Any particular reason for `topicTaggings` instead of `topicTags`?

Creeland (commented):
I double checked and it looks like you got everything! I just had one question on naming.

joelhooks:
tags are the tags themselves and taggings are tags associated with the thing tagged


I haven’t looked into this deeply but that’s the canonical difference.

src/lib/courses.ts
joelhooks: a "tagging" would be a join table and not the tag itself so it would represent a relationship

I don't know what's appropriate here

src/lib/courses.ts
joelhooks: I think we need a set of `conceptTags` or similar from a taxonomy perspective.

studio/schemas/documents/topic.js
```diff
@@ -36,7 +36,7 @@ export default {
       type: 'string',
     },
     {
-      title: 'Link to library',
+      title: 'Link to canonical resource',
```
joelhooks, line 39:
```suggestion
      title: 'Link to resource',
```

joelhooks (approved):
this looks good to me

Noticed it's been held up for awhile, let me know if I can provide more feedback.

jbranchaud:
To actually merge this would, iirc, require migrating a bunch of existing Sanity data to the new namings. Closing for now since it is stale and not a quick-merge. We can always revisit this if it seems valuable.

PR #1086, opened 2022-04-01
Add hidden rails id field on lesson document

Creeland:
This is so we can properly compare a unique attribute between schema. 

Originally we were comparing slugs then realized that those can be changed on either end. 

stamp


studio/schemas/documents/lesson.js
```diff
@@ -103,5 +103,11 @@ export default {
         layout: 'radio',
       },
     },
+    {
+      title: 'egghead-rails lesson id',
+      name: 'egghead_rails_id',
```
jbranchaud, line 108: We should come up with a consistent naming scheme for this kinds of _mirrored against egghead-rails_ attributes.

In `studio/schemas/documents/collaborator.js`, we called the egghead-rails instructor ID attribute `eggheadInstructorId`. This one is `egghead_rails_id`. Both the casing and the namespacing are mismatched.

It is not easy to rename attributes once they are in production and have data, which is why I'm suggesting thinking this through now.

Thoughts?

jbranchaud: For instance, we could do `eggheadRails<Model><Attribute>`:
- `eggheadRailsInstructorId`
- `eggheadRailsLessonId`

That would be easy enough to update here. It would require a small, low-risk migration for the `collaborator` entity which would need to change from `eggheadInstructorId` to `eggheadRailsInstructorId`. On the plus side, that would be a low-risk opportunity to practice one of these kinds of renaming migrations -- I expect we'll have to do more and more of that kind of thing as we continue to rely on Sanity for all things CMS.

Creeland: `eggheadRails<Model><Attribute>` makes a lot of sense to me. I can make that change and give the naming migration a shot

jbranchaud (approved):

PR #1089, opened 2022-04-04
Sketch out initial schema design for Course

jbranchaud:
If we are going to represent courses in Sanity, here is what they could
look like for starters. This pairs with the work we've been doing on
adding a `lesson` type and migrating lesson data. This is a move away
from the `resource` model which has a ton of fields that aren't being
used as it tries to be a way to flexibly represent everything.

We may iterate on and add to this model, but I think these cover the
essential attributes that we need to get the Course Builder off the
ground.

add course


Creeland (approved):
Schema looks good. It's missing some of the stuff on "resource" that gets displayed on the course landing page but that actually might beneficial. 

I think it would make sense to have some kind of landing page type sanity document where you can add a course + the other stuff like related content, "what you'll learn" bullets, etc.

jbranchaud:
> I think it would make sense to have some kind of landing page type sanity document where you can add a course + the other stuff like related content, "what you'll learn" bullets, etc.

Can you say more about this? I think I'm probably missing some context.

vojtaholik:
> I think it would make sense to have some kind of landing page type sanity document where you can add a course + the other stuff like related content, "what you'll learn" bullets, etc.

@Creeland I've been using resource with `landing-page` type for this kind of stuff.

jbranchaud:
> Schema looks good. It's missing some of the stuff on "resource" that gets displayed on the course landing page but that actually might beneficial.

I did a screen grab of a course page and marked up all the different pieces of content so that we can have an accounting of what those things are:

CleanShot 2022-04-14 at 12 40 30@2x

1. Pro vs Free course (access)
2. Title
3. Instructor (name, role, and avatar)
4. Course Stats (Duration, Rating out of 5 Stars, CC included, # of people who have completed the course — of these, probably only Duration and CC included are attributes that would be stored on the course in Sanity)
5. Published and Last Updated Dates
6. Course Cover Image
7. Attribution to cover image illustrator (name and role)
8. Description
9. The listing of lessons that comprise the course
10. Skills (or is this part of Description?)
11. Learner Reviews (this isn't strictly CMS content, so probably this is loaded in from somewhere else)
12. Related Resources

PR #1096, opened 2022-04-25
Reference Sanity project attributes from env vars

jbranchaud:
I was manually adjusting these values each time. And with there now being two different clients that get configured, it was now two places to check each time. This gives that job over to env vars.

environment


Creeland (approved):
approved

PR #1097, opened 2022-04-25
Minor refactorings to course builder

jbranchaud:
I noticed these two things while working from another feature branch. They are small, separate concerns so I decide to quickly peel off a separate PR for them.

- Add missing key prop for iterated elements
- Give appropriate name to anon function parameter

separate


Creeland (approved):
Good catches

PR #1098, opened 2022-04-26
Create sanity course from course builder

jbranchaud:
This PR has a _Course_ entity created in Sanity that gets associated with the _Lessons_ being created as well as the _Instructor_ selected from the drop down.

course

### DIFF
--- a/src/pages/api/sanity/lessons/create.ts
+++ b/src/pages/api/sanity/lessons/create.ts
@@ -2,6 +2,7 @@ import {NextApiRequest, NextApiResponse} from 'next'
 import {nanoid} from 'nanoid'
 import {ACCESS_TOKEN_KEY} from 'utils/auth'
 import {getAbilityFromToken} from 'server/ability'
+import {CourseData} from 'sanity-types'
 
 import client from '@sanity/client'
 
@@ -13,20 +14,39 @@ type LessonData = {
   }
 }
 
+type SanityReference<SanityType> = {
+  _type: SanityType
+  _ref: string
+}
+
+type SanityReferenceArray<SanityType> = Array<
+  {
+    _key: string
+  } & SanityReference<'reference'>
+>
+
 type SanityVideoResource = {
-  _type: string
+  _type: 'videoResource'
   _id: string
   filename: string
   originalVideoUrl: string
 }
 
 type SanityLesson = {
-  _type: string
+  _type: 'lesson'
+  _id: string
   title: string
   resource: {
     _type: string
     _ref: string
   }
+}
+
+type SanityCourse = {
+  _type: 'course'
+  title: string
+  collaborators: SanityReferenceArray<'collaborator'>
+  lessons: SanityReferenceArray<'lesson'>
 }
 
 const sanityClient = client({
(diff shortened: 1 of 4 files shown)


src/pages/api/sanity/lessons/create.ts
```diff
@@ -75,17 +141,21 @@ const createSanityLessons = async (
...
+
+      sanityResources.forEach((resource) => {
+        transaction.create(resource)
       })
 
       sanityLessons.forEach((lesson) => {
         transaction.create(lesson)
       })
 
+      transaction.create(sanityCourse)
```
jbranchaud, line 157: We are now creating the Sanity `course` along with the `lesson` and `videoResource` entities.

src/pages/upload/index.tsx
```diff
@@ -122,6 +131,14 @@ const Upload: React.FC<FormikProps<FormProps> & {instructors: Instructor[]}> = (
...
     viewer?.instructor_id || instructors[0]?.eggheadInstructorId,
   )
+  const sanityCollaboratorId: string | undefined = _find(instructors, [
+    'eggheadInstructorId',
+    courseInstructorId,
+  ])?._id
+
+  React.useEffect(() => {
+    setFieldValue('course.collaboratorId', sanityCollaboratorId)
+  }, [sanityCollaboratorId, setFieldValue])
```
jbranchaud, line 141: Match the instructor selected from the drop down to their `_id` and put that into Formik's values that will be submitted with the rest of the course builder metadata.

src/pages/upload/index.tsx
```diff
@@ -47,6 +51,7 @@ export const getServerSideProps: GetServerSideProps = async function ({req}) {
             'image': image.url,
           },
         eggheadInstructorId,
+        _id
```
jbranchaud, line 54: Adding `_id` to this GROQ query because that is needed for the collaborator `reference` on the `course`.

src/pages/api/sanity/lessons/create.ts
```diff
@@ -13,54 +14,119 @@ type LessonData = {
...
   token: process.env.SANITY_EDITOR_TOKEN,
 })
 
-async function formatSanityMutationForLessons(lessons: LessonData[]) {
+const sanityIdForDocumentType = async (
+  documentType: string,
+): Promise<string> => {
+  const id = await nanoid()
+  return `${documentType}-${id}`
+}
```
jbranchaud, line 64: Ensure nanoids that are generated are valid

Because `-` is in the base alphabet for nanoids, it sometimes generates
an id that starts with `-`. An id starting with `-` is not valid as a
Sanity `_id`. By prepending the document type to the generated id, we
ensure that doesn't happen and add a little helpful metadata to the
`_id` field.

src/pages/api/sanity/lessons/create.ts
```diff
@@ -13,54 +14,119 @@ type LessonData = {
...
-  })
+    ]
+  }
 
-  return {sanityLessons, sanityResources}
+  await Promise.all(
+    lessons.map(async (lesson) => {
+      const videoId = await sanityIdForDocumentType('videoResource')
+
+      sanityResources.push({
```
jbranchaud, line 102: I had to wrap this whole thing in an `await Promise.all(lessons.map(async () => { ... }))`. Because it is asynchronously adding each lesson to the `sanityLessons` array. Before it was a `forEach` which didn't work because the main thread was moving on with an empty `sanityLessons` array before the inline async functions could process.

Creeland (approved):
Looks good!

PR #1099, opened 2022-04-27
RFC: investigating how to handle a temporary duplicate instructor

jbranchaud:
Some (incomplete) notes from my investigation.

Most of the code isn't going to care about a duplicate. Where it gets complicated is the actual search page where Algolia gets involved. Still thinking through the implications and options..

think

### DIFF
--- a/src/components/search/index.tsx
+++ b/src/components/search/index.tsx
@@ -276,6 +276,7 @@ const Search: FunctionComponent<SearchProps> = ({
                       </>
                     )}
 
+                  {/* HIDE_DUPLICATES: this is where the instructor slug is used to look up and display specialized instructor landing pages, so it is important to make sure we are working with the primary instructor record and ignoring the duplicate one. */}
                   {!isEmpty(instructor) &&
                     shouldDisplayLandingPageForInstructor(instructor.slug) && (
                       <div className="pb-8 px-5">
@@ -284,6 +285,7 @@ const Search: FunctionComponent<SearchProps> = ({
                     )}
                   <ScrollElement name="hits" />
                   <Stats searchQuery={searchState.query} />
+                  {/* HIDE_DUPLICATES: the Hits component renders the Algolia Hits component which connects to an Algolia store that is tied in to any changes in the results that come from altering the search and/or facets of the search. So, the results that we show on the search page are deeply tied in to what Algolia returns. Because Instructors are one of the facets that Algolia searches against, we are going to have to let Algolia in on the action. */}
                   <Hits />
                   <div className="pb-16 pt-10 bg-gradient-to-t dark:from-gray-1000 dark:to-transparent from-gray-100 to-transparent">
                     <Pagination />
--- a/src/components/search/refinement-list.tsx
+++ b/src/components/search/refinement-list.tsx
@@ -268,6 +268,8 @@ const RefinementList: FunctionComponent<RefinementListProps> = ({
   )
 }
 
+// Docs on how `connectRefinementList` works
+// https://www.algolia.com/doc/api-reference/widgets/refinement-list/react/#connector
 const CustomRefinementList = connectRefinementList(RefinementList)
 
 export default CustomRefinementList
--- a/src/lib/instructors.ts
+++ b/src/lib/instructors.ts
@@ -25,6 +25,8 @@ export type Instructor = {
   avatar_64_url: string
 }
 
+// HIDE_DUPLICATES: there doesn't appear to be anywhere in the app that imports
+// this.
 export async function loadInstructors(page = 1) {
   const query = `query getInstructors($page: Int!){
     instructors(per_page: 24, page:$page){
--- a/src/pages/q/[[...all]].tsx
+++ b/src/pages/q/[[...all]].tsx
@@ -85,6 +85,11 @@ const SearchIndex: any = ({
     initialTopicData,
   )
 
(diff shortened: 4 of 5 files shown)


src/components/search/index.tsx
```diff
@@ -284,6 +285,7 @@ const Search: FunctionComponent<SearchProps> = ({
                     )}
                   <ScrollElement name="hits" />
                   <Stats searchQuery={searchState.query} />
+                  {/* HIDE_DUPLICATES: the Hits component renders the Algolia Hits component which connects to an Algolia store that is tied in to any changes in the results that come from altering the search and/or facets of the search. So, the results that we show on the search page are deeply tied in to what Algolia returns. Because Instructors are one of the facets that Algolia searches against, we are going to have to let Algolia in on the action. */}
```
garrettdimon, line 288: I assume the Algolia data updates are driven from Rails? If so, then we'd need to add that boolean to skip indexing the unlisted instructors? Or is it safe to simply let both instructors get indexed since it's the same content?

src/components/search/refinement-list.tsx
```diff
@@ -268,6 +268,8 @@ const RefinementList: FunctionComponent<RefinementListProps> = ({
   )
 }
 
+// Docs on how `connectRefinementList` works
+// https://www.algolia.com/doc/api-reference/widgets/refinement-list/react/#connector
```
garrettdimon, line 272: If I'm interpreting this correctly, if we prevent Algolia from indexing the unlisted instructor instance, we won't even have to make changes here, right?

src/lib/instructors.ts
```diff
@@ -25,6 +25,8 @@ export type Instructor = {
   avatar_64_url: string
 }
 
+// HIDE_DUPLICATES: there doesn't appear to be anywhere in the app that imports
```
garrettdimon, line 28: Based on the code, I'm assuming that if we did use this somewhere, it *would* need to be de-duped?

src/components/search/index.tsx
joelhooks: I'd be more inclined to tag the unlisted instructor over adding a new column to the database since this is a 1 off, but whatever makes sense.

Here's how the instructor relationship works. It's a **facet** of the content models:

https://www.loom.com/share/488eb44b08c74e4084cf3911274c6809

https://github.com/eggheadio/egghead-rails/blob/1d4e101d16c3062fd32025baf6dd8aa9b4fc30a1/app/models/playlist.rb#L222-L242

https://github.com/eggheadio/egghead-rails/blob/686ed03fcd70df56561da01a3ff99702d99c0114/app/models/lesson.rb#L10-L30

src/components/search/refinement-list.tsx
joelhooks: this is what we will want to do, but also make sure to associate the listed instructor in the index.

search index switch-a-roo should be our only backend change needed

src/lib/instructors.ts
joelhooks: I think that this was removed in favor of pointing at search page

src/pages/q/[[...all]].tsx
```diff
@@ -85,6 +85,11 @@ const SearchIndex: any = ({
     initialTopicData,
   )
 
+  // HIDE_DUPLICATES: the `loadInstructor` call here is only pulling in
+  // instructor metadata. So, if we have duplicate instructors living in
+  // egghead-rails and we are only pulling in one of them (based on slug), then
+  // as long as we are relying on the metadata for both duplicate instructors
+  // being the same, this won't be an issue.
```
Comment (joelhooks) [inline src/pages/q/[[...all]].tsx:92]: this isn't an exercise of deduplication IMO but more of a "if PSEUDO use REAL"

```
if (i.slug === 'fake-kcd') i.slug = 'real-kcd'
```

on the client, instructor meta is only for display and we don't want to have duplicate instructors we want to point the fake instructor to the real one's meta.

src/components/search/refinement-list.tsx
jbranchaud: > search index switch-a-roo

this sounds really promising to me. still need to look at exactly what is going into algolia, but I think this could work and make some of the other hackiness go away.

src/components/search/index.tsx
garrettdimon: There are no tags directly on the instructor are there? The lessons/playlists aren't what we need to hide. I'm all for tags on the instructor if we can do that. (Although, if the tags are displayed anywhere, a random `unlisted` tag would be awkward.)

src/components/search/refinement-list.tsx
garrettdimon: Because the Algolia stuff is declarative, it's not immediately clear to me how we can pull this off without letting Algolia index the unlisted instructor and including some form of flag for Algolia to ignore it.

If the connections to the Algolia index are name- or slug-based, then simply ensuring `published?` returns false for unlisted instructors should leave that instructor out of the index, and since it has the same name/slug as the real instructor, Algolia can be oblivious to the fact there's a different one, right?

jbranchaud: Algolia support conditional indexing https://www.algolia.com/doc/framework-integration/rails/indexing/indexing/?client=ruby#conditional-indexing

jbranchaud:
Resolved in https://github.com/eggheadio/egghead-rails/pull/4888

PR #1101, opened 2022-04-28
Use `&` to combine reference type into Array type

jbranchaud:
@Creeland we were wondering yesterday if there was a way to have `SanityReferenceArray` use the `SanityReference` type so that it didn't have to duplicate the individual type attributes. Looks like we can do it this way.

type

### DIFF
--- a/src/pages/api/sanity/lessons/create.ts
+++ b/src/pages/api/sanity/lessons/create.ts
@@ -21,11 +21,11 @@ type SanityReference = {
   _ref: string
 }
 
-type SanityReferenceArray = Array<{
-  _key: string
-  _type: 'reference'
-  _ref: string
-}>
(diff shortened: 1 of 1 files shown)


src/pages/api/sanity/lessons/create.ts
```diff
@@ -21,11 +21,11 @@ type SanityReference = {
...
 
-type SanityReferenceArray = Array<{
-  _key: string
-  _type: 'reference'
-  _ref: string
-}>
+type SanityReferenceArray = Array<
+  {
+    _key: string
+  } & SanityReference
```
Creeland, line 27: Nice! This will be a good one to remember 👍

Creeland (approved):

PR #1102, opened 2022-04-28
Add slug on sanity course create

jbranchaud:
- Include slugged-title when creating Sanity course
- Extract Sanity Slug type

slug


Creeland (approved):
slug

PR #1104, opened 2022-04-28
fix(pro): send in the proper loaded lesson to video machine

joelhooks:
mud

### DIFF
--- a/src/pages/lessons/[slug].tsx
+++ b/src/pages/lessons/[slug].tsx
@@ -390,6 +390,10 @@ const Lesson: React.FC<LessonProps> = ({
             send('SUBSCRIBE')
           }
         }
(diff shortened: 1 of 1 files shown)


src/pages/lessons/[slug].tsx
```diff
@@ -390,6 +390,10 @@ const Lesson: React.FC<LessonProps> = ({
             send('SUBSCRIBE')
           }
         }
+        videoService.send({
+          type: 'LOAD_RESOURCE',
+          resource: lessonState.context.lesson,
+        })
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:396]: after the actual video is loaded (not the basic SEO safe version) we need to pitch it into the video player

garrettdimon (approved):

PR #1107, opened 2022-05-05
Adds article for project based learning with egghead

zacjones93:
The goal of the article is to get learners excited about actually applying what they know in practical projects. It's a version of the pitch that we give learners in the learner interview before walking through the learning guide we put together for them.

i.e. You'll always be learning so leveling up how you learn is going to pay you back in large sums

We don't really have a distinction / strategy for when to put a resource like this in sanity vs somewhere in the site so would love any thoughts around that topic as well as a review on actual content!


laurosilvacom (commented):
content is ✨ 

IMO everything that's content should be published in Sanity 

see: https://github.com/eggheadio/egghead-next/pull/538

PR #1108, opened 2022-05-06
add flex to align elements when height is small

zacjones93:
I noticed when searching for content, if we don't have much to display the search result height makes the page look bad.

Adding flex aligns the pagination component to the bottom of the containing area.

## Before
element alignment is off

## After
bottom is aligned nicely

### DIFF
--- a/src/components/search/index.tsx
+++ b/src/components/search/index.tsx
@@ -285,7 +285,7 @@ const Search: FunctionComponent<SearchProps> = ({
                   <ScrollElement name="hits" />
                   <Stats searchQuery={searchState.query} />
                   <Hits />
-                  <div className="pb-16 pt-10 bg-gradient-to-t dark:from-gray-1000 dark:to-transparent from-gray-100 to-transparent">
+                  <div className="pb-16 pt-10 flex-grow bg-gradient-to-t dark:from-gray-1000 dark:to-transparent from-gray-100 to-transparent">
                     <Pagination />
                   </div>
                 </main>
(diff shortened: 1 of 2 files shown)


src/components/search/stats.tsx
```diff
@@ -13,7 +13,7 @@ const CustomStats: React.FunctionComponent<CustomStatsProps> = ({
   return !searchQuery || /^\s*$/.test(searchQuery) ? (
     <div />
   ) : (
-    <div className="sm:text-lg flex items-baseline flex-nowrap overflow-hidden max-w-full flex-grow pt-5 px-3">
+    <div className="sm:text-lg flex items-baseline flex-nowrap overflow-hidden max-w-full  pt-5 px-3">
```
zacjones93, line 16: Having flex-grow on the stats component created awkard space at the top of the search results. Moved this to the paginating component and it looks better to me.

reference: 
Resources for Web Developers in 2022  egghead io

src/components/search/index.tsx
```diff
@@ -227,7 +227,7 @@ const Search: FunctionComponent<SearchProps> = ({
               <div className="md:grid flex grid-cols-12 relative gap-3">
                 {RefinementsDesktop()}
                 {RefinementsMobile()}
-                <main className="col-span-10 w-full relative dark:bg-gray-900 bg-gray-50">
+                <main className="col-span-10 flex flex-col w-full relative dark:bg-gray-900 bg-gray-50">
```
zacjones93, line 230: flex to the rescue!

PR #1110, opened 2022-05-12
Remove unused <Stack /> component

jbranchaud:
I was looking at where `loadLesson` and `loadBasicLesson` were being imported and called. It is only a couple places. This one place is a component that hasn't been rendered anywhere in 18 months, so removing that component.

It looks like the original uses of Stack were removed in December 2020 (9d9e374663ee8423fffa5d057eb4dda8870e88f1) and replaced with the Resource component.

remove


jbranchaud:
Reduces the perceived footprint of the `loadLesson` utility. Once this is merged, it is only called in 3 places.

CleanShot 2022-05-12 at 13 14 14@2x

joelhooks (approved):

PR #1115, opened 2022-05-17
Spike out prisma talking directly to Postgres DB

jbranchaud:
We've been talking about moving various API endpoints and webhooks out of egghead-rails and into egghead-next. To the end of reducing (and eventually eliminating) egghead-next's dependence on egghead-rails. As I'm thinking about what it could look like to get off Rails, another strategy in our tool belt can be having egghead-next talk directly to the Postgres database.

As I see it, we can use this strategy from a couple angles:
1. use Prisma to talk to Postgres as part of transitional code that helps bridge the gap as we migrate some feature from egghead-rails to another system.
2. use Prisma to talk to Postgres as part of medium or long-term code that continues to interact with some version of the Postgres DB. For example, imagine transitioning egghead-specific coupon logic into a Next API endpoint which needs to talk to both Postgres and Stripe, while eliminating the Rails interaction.

As a demonstration, I replaced the _Check Pro User Status_ endpoint call with a Prisma->Postgres query. Note: it's not an exact replacement, but I think it gets the point across.

### Impressions

I was pleasantly surprised how quick this proof-of-concept was to throw together. The Prisma tooling is really slick. The only issues I ran into were a few shortcomings of Prisma not supporting type conversions (bigint -> int) and not supporting a table without a primary key (e.g. `users_roles`).

### Question

The main thing I'm not sure about is what to do in terms of publicly publishing the schema of the database. I don't know that it is particularly proprietary or sensitive, but I've never exposed an entire schema like that before, so wanted to discuss that first. **Thoughts?**

prism


prisma/schema.prisma.example
```diff
@@ -0,0 +1,17 @@
+generator client {
+  provider = "prisma-client-js"
+}
+
+datasource db {
+  provider = "postgresql"
+  url      = env("DATABASE_URL")
+}
```
jbranchaud, line 8: The `DATABASE_URL` will need to be specified in the `.env` file. For connecting to a local instance, I had something like:

```
DATABASE_URL="postgresql://jbranchaud:jbranchaud@localhost:5432/new_egghead_development?schema=public"
```

prisma/schema.prisma.example
```diff
@@ -0,0 +1,17 @@
...
+
+// The prisma models for the entire Postgres database schema are originally
+// generated by running `npx prisma db pull`. For the egghead database, this
+// required tweaking about a dozen errors in the generated models: type
+// mismatches (bigint -> int) and foreign key naming mismatches.
+//
+// Rather than sharing the schema publicly as part of this spike, I'll include
+// a private link to it separate from the PR. Just copy and paste that file
+// here to try it out.
```
jbranchaud, line 17: This is just a `.example` file, you'll want to create an actual `prisma/schema.prisma` file and paste the contents of the file I will share in Slack.

src/pages/api/users/check-pro-status.ts
```diff
@@ -13,19 +15,65 @@ const checkProStatus = async (req: NextApiRequest, res: NextApiResponse) => {
...
+        `,
+      )
+
+      console.log({result})
+
+      const hasProAccess = result.length >= 1
+
+      // const hasProAccess =
+      //   !isEmpty(eggheadUser) &&
+      //   (eggheadUser.is_pro || eggheadUser.is_instructor)
```
jbranchaud, line 76: As a proof-of-concept, I am replacing the API call to egghead-rails with a query directly against the database for whether or not the email is tied to a Pro status user.

prisma/schema.prisma.example
joelhooks: I think we should build this UP versus pulling in our entire database schema.

garrettdimon: @joelhooks I agree—assuming we find it practical to talk directly to the original DB as well as a newer one with our modern schema. (Whether hosted Postgres, Fly, or Planetscale.)

garrettdimon (commented):
This is encouraging for sure. A couple of questions/thoughts...

1. Does Prisma support and/or would it be practical to apply this in parallel to setting up the new modern database? I would guess that would work by having two separate Prisma clients instantiated with one connected to the original DB and the other (primary) connected to the new DB. 
2. I'm not sure how I feel about having `egghead-next` run multiple database connections from the standpoint of "junking" up the `egghead-next` repo, but if it buys us the flexibility we need to migrate safely, I'm certainly open to it. i.e. If it lets us migrate some API calls out of `egghead-rails` and into serverless functions by virtue of having Prisma to query the original database, then we could potentially migrate the logic first without touching the database, and once that's in place, we could create the new version of the database and update the logic to use the new-schema database at that point essentially separating the migration process into two phases.
3. Assuming this could be a piece to the puzzle, I think the next step would be determining if there are specific advantages/disadvantages for Postgres (via Render most likely), Fly.io, or Planetscale for the new database. I don't currently have any strong feelings, but I could definitely see Fly.io opening some doors for us performance-wise if we went that route. Similarly, the types of automation/preview things that Planetscale simplifies are really interesting too. (Think easy preview builds for testing or for sharing with instructors.)

Overall, I'd say this makes me that much more comfortable with our vision for getting `egghead-next` to be more self-contained.

joelhooks:
gonna close this for now

PR #1116, opened 2022-05-19
Separate Course and Overlay components from Lessons page

jbranchaud:
This PR does 3 things:
- Moves the `Course` component into its own file, the dependencies on that were small and clear cut.
- Moves the `specialLessons` data to its own file so that it can be managed apart from the component.
- Extracts an `Overlays` component which makes a clear delineation of what props are depended on for the various overlays. It handles rendering them, removes the `<OverlayWrapper>` duplication, and isolates several imports that `LessonPage` doesn't need.

Even though it is a pile of props being passed to `<Overlays ...>`, it is helpful to know exactly which ones are depended on.

lesson

### DIFF
--- a/src/components/pages/lessons/overlays.tsx
+++ b/src/components/pages/lessons/overlays.tsx
@@ -40,100 +40,99 @@ const Overlays: React.FC<OverlaysProps> = ({
   const {slug, tags = []} = lesson
   const primaryTag = get(first(tags), 'name', 'javascript')
 
+  let overlayToRender = null
+
+  if (lessonState.matches('joining')) {
+    overlayToRender = (
+      <EmailCaptureCtaOverlay lesson={lesson} technology={primaryTag} />
+    )
+  } else if (lessonState.matches('subscribing')) {
+    overlayToRender = (
+      <GoProCtaOverlay
+        lesson={lesson}
+        viewLesson={() => {
+          lessonSend({
+            type: 'LOAD',
+            lesson: lesson,
+            viewer,
+          })
+          // TODO: Make sure this is working as expected
+          videoService.send({
+            type: 'LOAD_RESOURCE',
+            resource: lesson,
+          })
+        }}
+      />
+    )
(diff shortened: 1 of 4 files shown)


src/components/pages/lessons/overlays.tsx
```diff
@@ -0,0 +1,141 @@
...
+  lessonSend: Function
+  lessonState: {matches: Function}
+  lesson: VideoResource
+  nextLesson: any
+  viewer: object
+  videoService: {send: Function}
+  lessonView: {collection_progress: {rate_url: string}}
+  subscriber?: CIOSubscriber
+  cioIdentify: Function
+}
```
jbranchaud, line 28: An exact list of what the overlays need to render

src/components/pages/lessons/overlays.tsx
```diff
@@ -0,0 +1,141 @@
...
+    overlayToRender = <RecommendNextStepOverlay lesson={lesson} />
+  }
+
+  return (
+    <>
+      {overlayToRender ? (
+        <OverlayWrapper>{overlayToRender}</OverlayWrapper>
+      ) : null}
+    </>
+  )
```
jbranchaud, line 138: This makes it clearer that exactly one overlay will ever be rendered, this was implied by the state machine matching, but this makes it more explicit.

src/pages/lessons/[slug].tsx
```diff
@@ -247,8 +195,6 @@ const Lesson: React.FC<LessonProps> = ({
   const transcriptAvailable = transcript || enhancedTranscript
   const {session_id} = router.query
 
-  const primary_tag = get(first(get(lesson, 'tags')), 'name', 'javascript')
-
```
Comment (jbranchaud) [inline src/pages/lessons/[slug].tsx:251]: This line sat in the middle of a bunch of data setup, but is only ever used by one overlay. It now lives much closer to where it is being used.

vojtaholik (approved):

PR #1120, opened 2022-06-10
Wire Stripe, CIO, and Mixpanel together for analytics

zacjones93:
@jbranchaud and I paired on this first commit to start egghead 2022 core metrics. This starts tracking when users sign up for a subscription by updating the user profile in mixpanel. 

One thing I'm unsure about this approach is if we also need to send a `track` call to build reports with it. That will be something to test out.

Specifically, this tracks the `customer.subscription.created` event in Stripe. Subsequent PR's for `customer.subscription.updated` and `customer.subscription.deleted` will be added to track all the data coming from stripe. `customer.subscription` looks like the best set of events to track that has the data that we need.

Josh and I raising our hands to the roof

### DIFF
--- a/package.json
+++ b/package.json
@@ -100,6 +100,7 @@
     "mdast-util-to-markdown": "^0.6.5",
     "mdx-bundler": "^5.2.1",
     "micro": "^9.3.4",
+    "mixpanel": "^0.16.0",
     "mixpanel-browser": "^2.41.0",
     "next": "^12.1.0",
     "next-compose-plugins": "^2.2.1",
--- a/package.json
+++ b/package.json
@@ -139,7 +139,7 @@
     "shortid": "^2.2.16",
     "simplebar-react": "^2.3.5",
     "slugify": "^1.6.0",
-    "stripe": "^8.170.0",
+    "stripe": "^9.8.0",
     "swr": "^0.5.5",
     "tailwindcss": "^3.0.13",
     "tailwindcss-autofill": "^0.1.0",
--- a/src/pages/api/stripe/webhook.ts
+++ b/src/pages/api/stripe/webhook.ts
@@ -1,6 +1,15 @@
+import axios from 'axios'
 import {buffer} from 'micro'
+import mixpanel from 'mixpanel-browser'
 import type {NextApiRequest, NextApiResponse} from 'next'
+import {track} from 'utils/analytics'
 import {stripe} from '../../../utils/stripe'
+
+const CIO_BASE_URL = `https://beta-api.customer.io/v1/api/`
+
+const cioAxios = axios.create({
+  baseURL: CIO_BASE_URL,
+})
 
 const webhookSecret = process.env.STRIPE_WEBHOOK_SECRET || ''
 
@@ -17,8 +26,35 @@ const stripeWebhookHandler = async (
     try {
       event = stripe.webhooks.constructEvent(buf, sig as string, webhookSecret)
 
-      if (event.type === 'checkout.session.completed') {
-        console.log('stripe webhook handler', {event})
+      let stripeCustomerId = event.data.object.customer
+
+      let stripeCustomer = await stripe.customers.retrieve(stripeCustomerId)
+
+      // if(typeof stripeCustomer === Customer)
+      let stripeEmail = stripeCustomer
+
+      await cioAxios
+        .get(`/customers/${stripeEmail}`, {
+          headers: {
+            Authorization: `Bearer ${process.env.CUSTOMER_IO_APPLICATION_API_KEY}`,
+          },
+        })
+        .then(({data}: {data: any}) => data.customer)
+        .catch((error: any) => {
+          console.error(error)
+        })
+
+      if (event.type === 'customer.subscription.created') {
+        // if running sale we would want to check for that but the regular case it's pro or PPP
+        let type = !event.data.object.discount ? 'pro' : 'PPP'
+
+        // how should we grab the current user? is there a pattern for this
+        let mixpanelCustomer = mixpanel.identify()
+
+        mixpanel.people.set({
+          type: type,
+          interval: event.data.object.plan.interval,
+        })
 
         res.status(200).send(`This works!`)
       } else {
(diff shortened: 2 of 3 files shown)


src/pages/api/stripe/webhook.ts
```diff
@@ -1,9 +1,21 @@
+import axios from 'axios'
 import {buffer} from 'micro'
+import Mixpanel from 'mixpanel'
```
zacjones93, line 3: Mixpanels node package

src/pages/api/stripe/webhook.ts
```diff
@@ -17,8 +29,43 @@ const stripeWebhookHandler = async (
...
+            params: {email: stripeEmail},
+            headers: {
+              Authorization: `Bearer ${process.env.CUSTOMER_IO_APPLICATION_API_KEY}`,
+            },
+          })
+          .then(({data}: {data: any}) => data.results[0])
+          .catch((error: any) => {
+            console.error(error)
+            return {}
+          })
```
zacjones93, line 57: We use users customerio id as the `unique_id` in mixpanel so when we need to look up the id from the email that comes back from stripe

src/pages/api/stripe/webhook.ts
```diff
@@ -17,8 +29,43 @@ const stripeWebhookHandler = async (
...
+
+        // if running sale we would want to check for that but the regular case it's pro or PPP
+        // TODO: need to differentiate between other types of discounts (e.g. sales)
+        let subscriptionType = !event.data.object.discount ? 'pro' : 'PPP'
```
zacjones93, line 61: as mentioned in comment, we'll need better 'type' checking here if we run a sale. When the full price is paid, the `discount` property is set to `null` on the event object that comes back from stripe. What that discount is could be handled better

src/pages/api/stripe/webhook.ts
```diff
@@ -17,8 +29,43 @@ const stripeWebhookHandler = async (
...
+        let interval = event.data.object.plan.interval
+
+        mixpanel.people.set(cioCustomer.id, {
+          pro: true,
+          type: subscriptionType,
+          interval,
+        })
```
zacjones93, line 68: Currently we are just setting profile property values. I'm fuzzy on how mixpanel will work for building reports if we will also have to send events to properly track the sign up rate

zacjones93:
We couldn't figure out why Stripe's typings were throwing an error here so the build is failing:

egghead-io-nextjs stripe error

package.json
```diff
@@ -138,7 +139,7 @@
     "shortid": "^2.2.16",
     "simplebar-react": "^2.3.5",
     "slugify": "^1.6.0",
-    "stripe": "^8.170.0",
+    "stripe": "^9.8.0",
```
joelhooks, line 142: https://github.com/stripe/stripe-node/wiki/Migration-Guide-for-v9

we don't use this API that was a breaking change so I moved us to latest here

src/pages/api/stripe/webhook.ts
```diff
@@ -17,8 +33,65 @@ const stripeWebhookHandler = async (
...
+          event.data.object.id,
+        )
+        const stripeCustomer = await stripe.customers.retrieve(
+          event.data.object.customer,
+        )
+
+        if (!('email' in stripeCustomer)) {
+          // solves a typescript error 🤡
+          throw new Error('No email found for customer')
+        }
```
joelhooks, line 51: this solved for the issue we were having with types

src/pages/api/stripe/webhook.ts
```diff
@@ -17,8 +33,65 @@ const stripeWebhookHandler = async (
...
+
+        const stripeEmail = stripeCustomer.email
+
+        // we want to get the `co_************` id from CustomerIO which we can
+        // cross-reference with MixPanel to uniquely identify the user.
+        // This correlates to the contact_id of the user in egghead.io
+        let cioCustomer = await cioAxios
```
joelhooks, line 58: I'm on the fence here and think of the egghead db as canonical but we don't currently fetch `Contact` by email anywhere. The `Contact#id` would be this same value.

src/pages/api/stripe/webhook.ts
```diff
@@ -17,8 +33,65 @@ const stripeWebhookHandler = async (
...
+          distinct_id: cioCustomer.id,
+          subscriptionType,
+          subscriptionInterval,
+          currentPeriodStart: stripeToMixpanelDataConverter(
+            event.data.object.current_period_start,
+          ),
+          currentPeriodEnd: stripeToMixpanelDataConverter(
+            event.data.object.current_period_end,
+          ),
+        })
```
joelhooks, line 90: Reviewing the plan, it really seems like we are interested in events and not customer properties for reporting so I moved most of this information here and associated it with the `distinct_id` 

regardless, I preferred longer property names here to avoid ambiguity (subscriptionType, subscriptionInterval, etc)

I didn't see a `Subscription Created` event in the plan doc

src/pages/api/stripe/webhook.ts
```diff
@@ -1,9 +1,49 @@
...
+      headers: {
+        Authorization: `Bearer ${process.env.CUSTOMER_IO_APPLICATION_API_KEY}`,
+      },
+    })
+    .then(({data}: {data: any}) => data.results[0])
+    .catch((error: any) => {
+      console.error(error)
+      return {}
+    })
+}
```
joelhooks, line 25: extracted because I used it again below

src/pages/api/stripe/webhook.ts
```diff
@@ -1,9 +1,49 @@
...
+
+function getCustomerEmail(stripeCustomer: any) {
+  if (!('email' in stripeCustomer) || !stripeCustomer.email) {
+    // solves a typescript error 🤡
+    throw new Error('No email found for customer')
+  }
+
+  return stripeCustomer.email
+}
```
joelhooks, line 34: feels like there has to be a better way but this works

src/pages/api/stripe/webhook.ts
```diff
@@ -17,8 +57,90 @@ const stripeWebhookHandler = async (
...
+        const stripeCustomer = await stripe.customers.retrieve(
+          event.data.object.customer,
+        )
+        const previousSubscription = event.data.previous_attributes
+        const newPlanAmount = stripeSubscription.items.data[0].plan.amount || 0
+
+        const state =
+          previousSubscription.plan.amount < newPlanAmount
+            ? 'upgrade'
+            : 'downgrade'
```
joelhooks, line 76: felt like this is what we want to know and this should answer it

I tested it a bit and it seems to do the right thing

joelhooks (approved):

PR #1121, opened 2022-06-15
test: add react-hooks testing library w/ example

jbranchaud:
This adds `testing-library/react-hooks` for being able to test custom react hooks.

I included a test for a fairly simple hook to demonstrate how to use it.

The motivation for adding this library is that I need it as part of another PR, but wanted to get it introduced in a less noisy context first.

hooks

### DIFF
--- a/src/hooks/__tests__/use-track-component.test.tsx
+++ b/src/hooks/__tests__/use-track-component.test.tsx
@@ -1,0 +1,23 @@
+import {renderHook, act} from '@testing-library/react-hooks'
+import {useTrackComponent} from '../use-track-component'
+import {track} from 'utils/analytics'
+
+jest.mock('utils/analytics')
+
+beforeEach(() => {
+  jest.clearAllMocks()
+})
+
+test('it calls track when it first renders', () => {
+  renderHook(useTrackComponent)
+
+  expect(track).toHaveBeenCalledTimes(1)
+})
+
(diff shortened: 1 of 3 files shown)


src/hooks/__tests__/use-track-component.test.tsx
```diff
@@ -0,0 +1,23 @@
+import {renderHook, act} from '@testing-library/react-hooks'
+import {useTrackComponent} from '../use-track-component'
+import {track} from 'utils/analytics'
+
+jest.mock('utils/analytics')
+
+beforeEach(() => {
+  jest.clearAllMocks()
+})
```
jbranchaud, line 9: This uses Jest's automatic mocking to enable the tests to see if `track` gets called.

https://jestjs.io/docs/es6-class-mocks#automatic-mock

src/hooks/__tests__/use-track-component.test.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+test('it calls track when it first renders', () => {
+  renderHook(useTrackComponent)
```
jbranchaud, line 12: The way this hook is supposed to work is that when the parent component first renders, it will trigger a `track` call. This `renderHook` call will trigger the initial `useEffect` call.

Creeland (approved):

PR #1122, opened 2022-06-15
Remove unused `useURL` and `useVideoJS` hooks

jbranchaud:
- Remove useURL hook, it is not being used.
- Remove useVideoJS hook, it isn't being used

The `useVideoJS` hook is the only place that the various `video.js` packages were being used, so I was able to remove these from our `package.json` as well.

cleanup


src/hooks/useVideo.tsx
```diff
@@ -1,34 +0,0 @@
-import {useRef, useState, useEffect, useCallback} from 'react'
-import videojs from 'video.js'
-import 'video.js/dist/video-js.css'
-import * as React from 'react'
-
-export const useVideoJS = (videoJsOptions: any) => {
```
jbranchaud, line 6: This is the only place where `video.js` packages were being used. So removing this hook means we can also remove those three packages from `package.json`.

joelhooks (approved):

PR #1123, opened 2022-06-15
feat(course-builder) disable submit button while uploading or incomplete

jbranchaud:
When the form isn't ready to be submitted, and especially if videos are
in the process of being uploaded, disable the button so that the form
cannot be prematurely submitted.

https://user-images.githubusercontent.com/694063/173902742-bc4cb406-408e-4766-b225-4ba1a0aa7859.mp4

submit


Creeland (approved):
Looks good. As a future TODO we should add error messages to the form

PR #1127, opened 2022-06-29
chore(deps): upgrade to next 12.2

joelhooks:
upgraded


src/lib/customer.ts
```diff
@@ -1,6 +1,6 @@
 const CIO_BASE_URL = `https://beta-api.customer.io/v1/api/`
 
-async function fetchCustomer(cioId: string) {
+async function fetchCustomer(cioId: string, timeout: number = 600) {
```
joelhooks, line 3: reduced this timeout since I added loading the auth'd user from egghead as well so the total time is under the 1.5s limit

src/middleware.ts
```diff
@@ -0,0 +1,162 @@
...
+ * cookies and such.
+ *
+ * This looks a lot like the i18n example:
+ * https://github.com/vercel/examples/tree/main/edge-functions/i18n
+ *
+ * an the split testing example that puts them in a bucket:
+ * https://github.com/vercel/examples/edge-functions/ab-testing-simple
+ *
+ */
+export async function middleware(req: NextRequest) {
```
joelhooks, line 32: this file was the biggest change since middleware isn't nested anymore with next 12.2 we needed to consolidate the two routes we use middleware for

https://nextjs.org/docs/messages/middleware-upgrade-guide#no-nested-middleware

src/middleware.ts
```diff
@@ -0,0 +1,162 @@
...
+  // if there's a cookie or a token they are logged in
+  let status = cioId || eggheadAccessToken ? 'identified' : 'anon'
+
+  // if there's a token, load the user (use cookie cached user if available)
+  const user =
+    eggheadAccessToken &&
+    (await loadUser(
+      eggheadAccessToken,
+      req.cookies.get(EGGHEAD_USER_COOKIE_KEY),
+    ))
```
joelhooks, line 54: previously we weren't loading the egghead user and I noticed that it had issues if for some reason the CIO user wasn't updated etc so this handles that. if the user is already cookied it will use that.

src/utils/auth.ts
```diff
@@ -199,6 +202,10 @@ export default class Auth {
           domain: process.env.NEXT_PUBLIC_AUTH_COOKIE_DOMAIN,
         })
 
+        cookie.remove(EGGHEAD_USER_COOKIE_KEY, {
+          domain: process.env.NEXT_PUBLIC_AUTH_COOKIE_DOMAIN,
+        })
```
joelhooks, line 207: if they logout, remove the user cookie entry

src/utils/auth.ts
```diff
@@ -243,6 +250,9 @@ export default class Auth {
             cookie.set(CIO_KEY, data.contact_id)
           }
           localStorage.setItem(USER_KEY, JSON.stringify(data))
+          cookie.set(EGGHEAD_USER_COOKIE_KEY, JSON.stringify(data), {
+            domain: process.env.NEXT_PUBLIC_AUTH_COOKIE_DOMAIN,
+          })
```
joelhooks, line 255: also setting the cookie here so it stays fresh

src/middleware.ts
```diff
@@ -0,0 +1,162 @@
+import {NextRequest, NextResponse} from 'next/server'
+import {ACCESS_TOKEN_KEY, EGGHEAD_USER_COOKIE_KEY} from './config'
+import {loadCio} from './lib/customer'
+import {loadUser} from './lib/current-user'
+
+const PUBLIC_FILE = /\.(.*)$/
+
+const CIO_COOKIE_KEY = 'cio_id'
+
+// Supports both a single string value or an array of matchers
```
jbranchaud, line 10:
```suggestion
// `matcher` determines what routes this middleware runs on.
// Supports both a single string value or an array of matchers
// https://nextjs.org/docs/advanced-features/middleware#matcher
```

src/middleware.ts
jbranchaud: This looks good. if it ends up with more route blocks, we could explore breaking the blocks out into separate functions/modules that get invoked here. That'd make it easier to glance at what is happening and also encapsulate the specifics of each (e.g. inputs and output).

jbranchaud (approved):
🚀

PR #1129, opened 2022-07-11
refactor: consolidate cio_id constants to config

jbranchaud:
We had `cio_id` floating around as a magic string in several modules. This consolidates it to a single value that can be imported from `config`.

cio


src/config.ts
```diff
@@ -1,2 +1,3 @@
 export const ACCESS_TOKEN_KEY = 'eh_token_2020_11_22'
 export const EGGHEAD_USER_COOKIE_KEY = 'eh_user'
+export const CIO_IDENTIFIER_KEY = 'cio_id'
```
jbranchaud, line 3: define it here, import it everywhere else

jbranchaud:
This PR is going to need a little updating in light of https://github.com/eggheadio/egghead-next/pull/1131

joelhooks (approved):

PR #1131, opened 2022-07-11
chore: refactor middleware routing and cookie handling

joelhooks:
routes

### DIFF
--- a/src/lib/customer.ts
+++ b/src/lib/customer.ts
@@ -34,8 +34,13 @@ async function fetchCustomer(
     })
       .then((response) => {
         try {
-          return response.json().then((customer: CIOCustomer) => {
-            if (!timedOut) resolve(customer)
+          return response.json().then((customer: any) => {
+            if (!timedOut)
+              resolve(
+                customer?.customer
+                  ? customer.customer
+                  : (customer as CIOCustomer),
+              )
           })
         } catch (error) {
           if (!timedOut) resolve(null)
--- a/src/lib/customer.ts
+++ b/src/lib/customer.ts
@@ -1,6 +1,15 @@
 const CIO_BASE_URL = `https://beta-api.customer.io/v1/api/`
 
-async function fetchCustomer(cioId: string, timeout: number = 400) {
+type CIOCustomer = {
+  attributes: {
+    react_score: number
+  }
+}
+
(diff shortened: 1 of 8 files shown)


src/lib/customer.ts
```diff
@@ -15,7 +24,7 @@ async function fetchCustomer(cioId: string, timeout: number = 400) {
 
     const id = setTimeout(() => {
       timedOut = true
-      reject(`timeout loading customer [${cioId}]`)
+      resolve(null)
```
joelhooks, line 27: just resolve it as nothing and move on

joelhooks (commented):
the `middleware.ts` file had a lot of duplicated logic across multiple routes so this is intended to break it apart into several files and reduce the duplicated functionality so it can be more testable and extensible in the future

src/server/customer-io-cookies.ts
```diff
@@ -0,0 +1,44 @@
...
+  })
+  res.cookies.delete(CIO_CUSTOMER_OBJECT_KEY, {
+    domain: process.env.NEXT_PUBLIC_AUTH_COOKIE_DOMAIN,
+  })
+}
+
+export function setCustomerCookie(res: NextResponse, customer: any) {
+  if (customer?.id) {
+    res.cookies.set(CIO_COOKIE_KEY, customer.id, {
+      maxAge: 1000 * 60 * 60 * 24 * 365,
```
jbranchaud, line 18:
```suggestion
      maxAge: 1000 * 60 * 60 * 24 * 365, // 1 year in milliseconds
```

src/server/egghead-user-cookies.ts
```diff
@@ -0,0 +1,34 @@
...
+      JSON.stringify({
+        contact_id,
+        email,
+        first_name,
+        last_name,
+        is_pro,
+        is_instructor,
+      }),
+      {
+        maxAge: 1000 * 60 * 60 * 24 * 2,
```
jbranchaud, line 19:
```suggestion
        maxAge: 1000 * 60 * 60 * 24 * 2, // 2 days in milliseconds
```

jbranchaud (approved):
solid refactor, this all feels a lot easier to wrap my head around this way 💯

PR #1137, opened 2022-07-14
fix: course redirects properly from playlist route

joelhooks:
traffic


src/pages/playlists/[slug].tsx
```diff
@@ -8,8 +8,10 @@ export default Playlist
...
   const slug = params && (params.slug as string)
-  res.setHeader('Location', `/courses/${slug}`)
-  res.statusCode = 302
-  res.end()
-  return {props: {}}
+  return {
+    redirect: {
+      destination: `/courses/${slug}`,
+      permanent: true,
+    },
```
Comment (joelhooks) [inline src/pages/playlists/[slug].tsx:15]: this was the issue. the header approach didn't work, sing the redirect syntax does.

Issue #1138, opened 2022-07-15
Differences in Lesson Loading GraphQL Requests


jbranchaud:
As part of trying to understand what attributes and associated data we need when loading a Lesson for a lesson page, I compared and contrasted the `loadLesson` and `loadBasicLesson` graphql queries that appear in [`src/lib/lessons.ts`](https://github.com/eggheadio/egghead-next/blob/main/src/lib/lessons.ts).

I'll summarize my findings and then reproduce the entire annotated queries.

![findings](https://media2.giphy.com/media/26n6WywJyh39n1pBu/giphy.gif?cid=d1fd59absc8u8s7jazc6x72c35ejzt7jax22aqt22rkfdkxu&rid=giphy.gif&ct=g)

## Headline

The headline for me is that I think we can make a couple (**4**) small adjustments to the `loadLesson` function for all our needs and then throw away the `loadBasicLesson` function. For starters, this will reduce confusion about which to use and when. There is no way to make that decision at a quick glance because both are loading most of the kitchen sink. From here, there is probably additional trimming down and clarifying that we can do.

## Summary

After doing a side-by-side comparison of the two queries, I was able to rearrange a few attributes and then add comments on both sides for the ones that were missing (_annotated by `""" some_attr """`_).

For `loadLesson`, the only attributes that `loadBasicLesson` requests that it doesn't already request are:

- `lesson.updated_at`
- `lesson.published_at`
- `collection.playlist.lesson[].media_url`
- `collection.course.lesson[].media_url`

For `loadBasicLesson`, they are still quite similar but it is missing a few more attributes:

- `lesson.duration`
- `lesson.transcript_url`
- `lesson.next_up_url`
- `lesson.dash_url`
- `lesson.http_url`
- `lesson.lesson_view_url`
- `lesson.download_url`
- `collection.playlist.lesson[].thumb_url`
- `collection.course.lesson[].thumb_url`
- `lesson.course.title`
- `lesson.course.square_cover_480_url`
- `lesson.course.slug`
- `lesson.repo_url`
- `lesson.code_url`

## Annotated Queries

```graphql
""" loadLesson """
const query = /* GraphQL */ `
  query getLesson($slug: String!) {
    lesson(slug: $slug) {
      slug
      title
      duration
      transcript
      transcript_url
      description
      free_forever
      subtitles_url
      media_url
      next_up_url
      hls_url
      dash_url
      http_url
      thumb_url
      lesson_view_url
      path
      icon_url
[...]

PR #1139, opened 2022-07-15
refactor: replace and remove loadBasicLesson

jbranchaud:
guiding issue: https://github.com/eggheadio/egghead-next/issues/1138

- refactor: add 4 missing attrs to loadLesson gql
- refactor: do logical rearrangement of lesson attrs
- refactor: replace loadBasicLesson w/ loadLesson

The headline for me is that I think we can make a couple (4) small adjustments to the `loadLesson` function for all our needs and then throw away the `loadBasicLesson` function. For starters, this will reduce confusion about which to use and when. There is no way to make that decision at a quick glance because both are loading most of the kitchen sink. From here, there is probably additional trimming down and clarifying that we can do.

Here is the commit (https://github.com/eggheadio/egghead-next/commit/7e4ce0300494f136cde46342752f3fc47b464cd5) that originally introduced the `loadBasicLesson` function. Since then a handful of attributes have been added to it. **@joelhooks, let me know if I'm missing some important context for why this separate query was originally added.**

basic


joelhooks (approved):
this approach is solid and makes sense to me

PR #1140, opened 2022-07-15
fix: update to new style of Next.js redirects

jbranchaud:
In this other PR
(https://github.com/eggheadio/egghead-next/pull/1137/files) we had to
update some redirects from an old style to one that is now supported by
Next.js 12.2
(https://nextjs.org/docs/api-reference/data-fetching/get-server-side-props#redirect).
This commit does the same for a couple remaining redirects in the
`lessons/[slug].tsx` page component.

redirect

### DIFF
--- a/src/pages/lessons/[slug].tsx
+++ b/src/pages/lessons/[slug].tsx
@@ -80,10 +80,12 @@ export const getServerSideProps: GetServerSideProps = async function ({
       params && (await loadBasicLesson(params.slug as string))
 
     if (initialLesson && initialLesson?.slug !== params?.slug) {
-      res.setHeader('Location', initialLesson.path)
-      res.statusCode = 302
-      res.end()
-      return {props: {}}
+      return {
+        redirect: {
+          destination: initialLesson.path,
+          permanent: true,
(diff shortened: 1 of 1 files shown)


src/pages/lessons/[slug].tsx
```diff
@@ -80,10 +80,12 @@ export const getServerSideProps: GetServerSideProps = async function ({
...
 
     if (initialLesson && initialLesson?.slug !== params?.slug) {
-      res.setHeader('Location', initialLesson.path)
-      res.statusCode = 302
-      res.end()
-      return {props: {}}
+      return {
+        redirect: {
+          destination: initialLesson.path,
+          permanent: false,
```
Comment (joelhooks) [inline src/pages/lessons/[slug].tsx:86]: this should be true!

Comment (jbranchaud) [reply src/pages/lessons/[slug].tsx:86]: I'll switch it to true.

I think it originally should have been 301 instead of 302 -- I did know that distinction off the top of my head, so looked it up https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/302

joelhooks (approved):

PR #1141, opened 2022-07-15
Add state field to course and resource in sanity

Creeland:
This adds a single select field to both of these document types in Sanity. 

Course isn't in use yet but since it will be soon we've added it to its schema. 

insane

### DIFF
--- a/studio/schemas/documents/course.js
+++ b/studio/schemas/documents/course.js
@@ -44,7 +44,7 @@ export default {
             value: 'contentReview',
           },
           {
-            title: 'pre-release',
+            title: 'pre release',
             value: 'preRelease',
           },
           {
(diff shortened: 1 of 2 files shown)


studio/schemas/documents/course.js
```diff
@@ -21,6 +21,39 @@ export default {
...
+          {
+            title: 'published',
+            value: 'published',
+          },
+          {
+            title: 'content review',
+            value: 'contentReview',
+          },
+          {
+            title: 'pre-release',
```
zacjones93, line 47:
```suggestion
            title: 'pre release',
```
we can keep this consistent with 'content review' above

zacjones93 (approved):

PR #1144, opened 2022-07-19
fix: add consistent style to course callout on blog pages

zacjones93:
I was looking at our SEO and noticed our top blog post had a course CTA component on it. It looked like it could use some clean up so I added some styles to match similar course CTAs.

Primarily for: https://egghead.io/blog/intercepting-network-requests-in-cypress

painting

## Before
course call out before

## After
course call out darkmode
course call out light mode
course call out mobile


Creeland (approved):
👍

PR #1145, opened 2022-07-21
feat: add Sanity Studio environment switcher

jbranchaud:
_Update: this now uses Node and handlebars instead of Ruby and ERB._

The way that we separate both data and schema between staging and
production Sanity is to use separate projects. This is a solution that
we sort of hacked together because there isn't an official way to get
this kind of separation. Because of that, it is tedious to swap out the
staging and production config values when wanting to switch between
those two projects. This script introduces an ERB template that makes it
is to toggle between the environments by running a single command.

The way it works is, based on the environment you provide (either `staging` or `production`), it will generate and overwrite the `sanity.json` file using the ERB template and the environment-specific values.

This is a spike. Ideally we'd find a way to do this in JS/TS. I'm not
sure what the ERB equivalent is in the Node ecosystem.

This is something I had written earlier in the week before the internal discussion on preferring Node tooling for this kinds of scripts. I'd love to discuss how something equivalent to this could be implemented with TS and some templating library in that ecosystem.

template

### DIFF
--- a/package.json
+++ b/package.json
@@ -214,6 +214,7 @@
     "eslint-plugin-react-hooks": "^4.0.8",
     "eslint-plugin-testing-library": "^4.11.0",
     "globby": "^12.0.1",
+    "handlebars": "^4.7.7",
     "husky": "^7.0.1",
     "istanbul-lib-coverage": "^3.0.0",
     "jest": "^27.0.6",
--- a/studio/sanity.json.erb
+++ b/studio/sanity.json.erb
@@ -1,0 +1,46 @@
+{
+  "root": true,
+  "project": {
+    "name": "<%= @project_name %>"
+  },
+  "api": {
+    "projectId": "<%= @project_id %>",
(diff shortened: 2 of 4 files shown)


jbranchaud:
Try rewriting this using handlebars.js: https://handlebarsjs.com/

Tutorial: https://vahid.blog/post/2021-03-19-how-to-use-handlebars.js-for-templating/

- Read in the template file
- Compile the template into a function `let templateFunc = Handlebars.compile(template);`
- Generate env-specific file using template function `let templateObject = templateFunc(envSpecificData);`
- Write out the result to `sanity.json`

Creeland:
I haven't read this article yet but it seems relevant to us https://www.sanity.io/guides/multi-environment-deployments

jbranchaud:
Latest update to this PR: the script now uses Node and handlebars instead of Ruby and ERB.

joelhooks (approved):

PR #1146, opened 2022-07-21
Remove unused load lessons functions

jbranchaud:
- cleanup: remove unused loadLessons function

The component that used this function was removed back in August 2020.
It hasn't been used since.
https://github.com/eggheadio/egghead-next/commit/ff184c09e66a66f4b7a9ed0f7aae649962408dc5

Plus we are trying to remove data loading dependence on egghead-rails,
especially for lesson metadata.

- cleanup: remove unused loadLessonForUser function

This function was only used in one place and that was removed back in
February 2021 and replaced with a call to `loadLesson`.
https://github.com/eggheadio/egghead-next/commit/11767d06554c85de203631c30ad0917cd4d11508

load lesson

### DIFF
--- a/src/lib/lessons.ts
+++ b/src/lib/lessons.ts
@@ -1,29 +1,6 @@
 import {LessonResource} from 'types'
 import {getGraphQLClient} from '../utils/configured-graphql-client'
-import config from './config'
 import getAccessTokenFromCookie from '../utils/get-access-token-from-cookie'
-
-export async function loadLessons(): Promise<LessonResource[]> {
-  const query = /* GraphQL */ `
-    query getLessons {
-      lessons(per_page: 25) {
-        title
-        slug
-        icon_url
-        instructor {
-          full_name
-          avatar_64_url
-        }
-      }
-    }
-  `
-
-  const graphQLClient = getGraphQLClient()
-
-  const {lessons} = await graphQLClient.request(config.graphQLEndpoint, query)
-
-  return lessons
-}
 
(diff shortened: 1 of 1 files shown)


src/lib/lessons.ts
```diff
@@ -137,31 +114,3 @@ export async function loadLesson(slug: string, token?: string) {
 
   return lesson as LessonResource
 }
-
-export async function loadLessonForUser(slug: string) {
```
jbranchaud, line 141: This function was only used in one place and that was removed back in
February 2021 and replaced with a call to `loadLesson`.
https://github.com/eggheadio/egghead-next/commit/11767d06554c85de203631c30ad0917cd4d11508

src/lib/lessons.ts
```diff
@@ -1,30 +1,7 @@
 import {LessonResource} from 'types'
 import {getGraphQLClient} from '../utils/configured-graphql-client'
-import config from './config'
 import getAccessTokenFromCookie from '../utils/get-access-token-from-cookie'
 
-export async function loadLessons(): Promise<LessonResource[]> {
```
jbranchaud, line 6: The component that used this function was removed back in August 2020.
It hasn't been used since.
https://github.com/eggheadio/egghead-next/commit/ff184c09e66a66f4b7a9ed0f7aae649962408dc5

Plus we are trying to remove data loading dependence on egghead-rails,
especially for lesson metadata.

Creeland (approved):
I double checked to make sure that there weren't any missed uses. Looks good!

PR #1148, opened 2022-07-26
feat: load Lesson metadata from Sanity

jbranchaud:
This PR puts up the scaffolding for loading Lesson metadata from Sanity alongside egghead-rails (where it is already coming from).

This merges the lesson metadata from both egghead-rails and Sanity, so
if one or the other is either missing or incomplete, the other will make
up for it. This helps as we transition from egghead-rails to Sanity as
the source of truth for lesson metadata.

Individual Commits:

- refactor: extract graphql query to const
- refactor: be explicit about loadLesson return type
- cleanup: remove unused `course` from Lesson query
- refactor: load lesson comments separately
- refactor: move load fn to lesson-comments module
- feat: load some lesson metadata from Sanity

metadata

### DIFF
--- a/src/lib/lesson-comments.ts
+++ b/src/lib/lesson-comments.ts
@@ -1,4 +1,6 @@
 import axios from 'utils/configured-axios'
+import {getGraphQLClient} from '../utils/configured-graphql-client'
+import getAccessTokenFromCookie from '../utils/get-access-token-from-cookie'
 
 export type Comment = {
   title?: string
@@ -13,4 +15,62 @@ export async function saveCommentForLesson(slug: string, comment: Comment) {
       {comment},
     )
     .then(({data}) => data)
+}
+
+type GraphQLComment = {
+  comment: string
+  commentable_id: number
+  commentable_type: string
+  created_at: string
+  id: number
+  is_commentable_owner: boolean
+  state: string
+  user: {
+    avatar_url: string
+    full_name: string
+    instructor: {
+      first_name: string
+    }
+  }
+}
+
+export async function loadLessonComments(
+  slug: string,
+  token?: string,
+): Promise<GraphQLComment[]> {
+  token = token || getAccessTokenFromCookie()
+  const graphQLClient = getGraphQLClient(token)
+
+  const variables = {
+    slug: slug,
+  }
+
+  const query = /* GraphQL */ `
+    query getLesson($slug: String!) {
+      lesson(slug: $slug) {
+        comments {
+          comment
+          commentable_id
+          commentable_type
+          created_at
+          id
+          is_commentable_owner
+          state
+          user {
+            avatar_url
+            full_name
+            instructor {
+              first_name
+            }
+          }
+        }
+      }
+    }
+  `
+
+  const {
+    lesson: {comments},
+  } = await graphQLClient.request(query, variables)
+
+  return comments
 }
--- a/src/lib/lessons.ts
+++ b/src/lib/lessons.ts
@@ -84,11 +84,6 @@ const loadLessonGraphQLQuery = /* GraphQL */ `
           }
         }
       }
-      course {
-        title
-        square_cover_480_url
-        slug
-      }
       tags {
         name
         label
--- a/src/lib/lessons.ts
+++ b/src/lib/lessons.ts
@@ -18,7 +18,11 @@ export async function loadLesson(
     variables,
   )
 
-  return lesson as LessonResource
+  // comments are user-generated content that must come from the egghead-rails
+  // backend, so load those separately from the rest of lesson metadata.
+  const comments = await loadLessonComments(slug, token)
+
+  return {...lesson, comments} as LessonResource
 }
 
 const loadLessonGraphQLQuery = /* GraphQL */ `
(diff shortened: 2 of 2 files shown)


src/lib/lesson-comments.ts
```diff
@@ -14,3 +16,61 @@ export async function saveCommentForLesson(slug: string, comment: Comment) {
...
+): Promise<GraphQLComment[]> {
+  token = token || getAccessTokenFromCookie()
+  const graphQLClient = getGraphQLClient(token)
+
+  const variables = {
+    slug: slug,
+  }
+
+  const query = /* GraphQL */ `
+    query getLesson($slug: String!) {
```
jbranchaud, line 49: Load lesson comments separately since those are user-generated content. That will make it easier to swap out the other graphql query once it loads exclusively metadata.

src/lib/lesson-comments.ts
```diff
@@ -14,3 +16,61 @@ export async function saveCommentForLesson(slug: string, comment: Comment) {
...
+          state
+          user {
+            avatar_url
+            full_name
+            instructor {
+              first_name
+            }
+          }
+        }
+      }
```
jbranchaud, line 67: pulled this directly out of the `loadLesson` graphql query

src/lib/lessons.ts
```diff
@@ -1,116 +1,149 @@
...
 import {getGraphQLClient} from '../utils/configured-graphql-client'
 import getAccessTokenFromCookie from '../utils/get-access-token-from-cookie'
+import {loadLessonComments} from './lesson-comments'
+import {sanityClient} from 'utils/sanity-client'
+import groq from 'groq'
 
-export async function loadLesson(slug: string, token?: string) {
-  const query = /* GraphQL */ `
-    query getLesson($slug: String!) {
-      lesson(slug: $slug) {
```
jbranchaud, line 8: Moved the query (the string itself) out of the function and down below. This makes it so that when you open this file you can read the whole `loadLesson` function at a glance to see what it is doing.

src/lib/lessons.ts
```diff
@@ -1,116 +1,149 @@
...
-        published_at
-        collection {
-          ... on Playlist {
-            title
+const lessonQuery = groq`
+*[_type == 'lesson' && slug == $slug][0]{
+  title,
+  slug,
+  description
+}`
```
jbranchaud, line 13: absolute base groq query for loading lesson metadata from Sanity, more to come here in followup PRs once we expand the Lesson schema.

src/lib/lessons.ts
```diff
@@ -1,116 +1,149 @@
...
+ */
+async function loadLessonMetadataFromSanity(slug: string) {
+  const params = {
+    slug,
+  }
+
+  const lesson = await sanityClient.fetch(lessonQuery, params)
+
+  return lesson
+}
```
jbranchaud, line 27: use groq and SanityClient to load lesson metadata from Sanity

src/lib/lessons.ts
```diff
@@ -1,116 +1,149 @@
...
+
+  const lessonMetadataFromSanity = await loadLessonMetadataFromSanity(slug)
+
+  const lessonMetadata = {...lesson, ...lessonMetadataFromSanity}
+
+  // comments are user-generated content that must come from the egghead-rails
+  // backend, so load those separately from the rest of lesson metadata.
+  const comments = await loadLessonComments(slug, token)
+
+  return {...lessonMetadata, comments} as LessonResource
```
jbranchaud, line 53: merge in the comments as well

src/lib/lessons.ts
```diff
@@ -1,116 +1,149 @@
...
-              thumb_url
-              media_url
-            }
+            completed
+            duration
+            thumb_url
+            media_url
           }
         }
-        course {
```
jbranchaud, line 69: remove `course` because it wasn't being used anywhere downstream once it was loaded in.

CleanShot 2022-07-26 at 15 02 10@2x


Commit: https://github.com/eggheadio/egghead-next/pull/1148/commits/6f3bdd19213606cd3e312cf2662fd4cbadd491d1

src/lib/lesson-comments.ts
joelhooks: suggestion(non-essential): we might consider flipping this and have some sort of comments endpoint instead on the egghead side. Kind of the same thing really, but would be clearer

src/lib/lessons.ts
joelhooks: suggestion: we might consider making the comments request in a totally separate call from "load lesson" for performance reasons since this adds an additional network call? dunno if it matters but "filling in" comments on the client is probably fine. They might have some SEO advantage.

jbranchaud: I agree with loading them separately. In this first pass, I was trying to keep the change-surface-area low.

src/lib/lesson-comments.ts
jbranchaud: That would be clearer. That will require a small PR to egghead-rails, so I'll do that separately from this PR.

Creeland (approved):

PR #1150, opened 2022-07-27
feat: groq for a Lesson's collection and siblings

jbranchaud:
This is the result of some pairing between @Creeland and I.

This adds the collection portion of the graphql Lesson emulation. It
finds the first containing collection (a course) as well as all the
lessons tied to that course (this lesson's siblings).

tl;dr: this is some cool GROQ, but it is also slow GROQ.

To query for this data from the context of a lesson requires some
"sub-queries" (not sure if GROQ calls them that). This appears to be
leading to a big hit in performance. In the Studio Vision tab, running
this query, we are seeing execution time jump to a little above 1000ms.
We are going to need to either adjust this query to get better
performance or break it up into separate queries that each have a better
performance profile.

We are working with a small amount of data in a small graph (in our staging dataset), so I'm hoping we are just missing a better way to do this with GROQ.

Some notes/questions on the performance of this query: https://roamresearch.com/#/app/egghead/page/QQjJQ1QE6

performance

### DIFF
--- a/src/lib/lessons.ts
+++ b/src/lib/lessons.ts
@@ -21,6 +21,22 @@ const lessonQuery = groq`
     'label': library->slug.current,
     'http_url': library->url,
     'image_url': library->image.url
+  },
+  'collection': *[_type=='course' && references(^._id)][0]{
+    title,
+    'slug': slug.current,
+    'type': 'playlist',
+    'square_cover_480_url': image,
+    'path': '/courses/' + slug.current,
+    'lessons': lessons[]-> {
+      'slug': slug.current,
+      'type': 'lesson',
+      'path': '/lessons/' + slug.current,
+      'title': title,
+      'duration': 0,
+      'thumb_url': null,
+      'media_url': *[_type=='videoResource' && _id == ^.resource->_id][0].hslUrl
+    }
   }
 }`
 


jbranchaud:
I opened this as a separate PR from https://github.com/eggheadio/egghead-next/pull/1149 because there are some questions and concerns around performance that we need to work out.

joelhooks (approved):

joelhooks:
perf in what regard? sanity is usually very fast and well cached.

jbranchaud:
> perf in what regard? sanity is usually very fast and well cached.

From Sanity Studio Viewer we were seeing this query take anywhere from 1200ms to 2400ms. @Creeland did some investigation and benchmarking on it yesterday and noticed that the query was only slow on the first run. I assumed due to caching, but even changing the lesson slug (to a lesson that couldn't have been cached yet) would be quick. So seems like it is possibly an initial perf penalty for setting up the query (a la a stored procedure) and then it is quick to process after that. This is speculation. Regardless, we are less concerned about perf at this point.

Anything to add to that, @Creeland ?

src/lib/lessons.ts
```diff
@@ -21,6 +21,22 @@ const lessonQuery = groq`
...
+    'square_cover_480_url': image,
+    'path': '/courses/' + slug.current,
+    'lessons': lessons[]-> {
+      'slug': slug.current,
+      'type': 'lesson',
+      'path': '/lessons/' + slug.current,
+      'title': title,
+      'duration': 0,
+      'thumb_url': null,
+      'media_url': *[_type=='videoResource' && _id == ^.resource->_id][0].hslUrl
```
jbranchaud, line 38: The `media_url` can be rewritten like this which is a little faster (because it doesn't have to do a global filter) and more direct (because it is just working with the ref'd resource).

```suggestion
      resource->_type == 'videoResource' => {
        'media_url': resource->hslUrl
      },
```

PR #1151, opened 2022-08-01
feat: use getLessonComments graphql field

jbranchaud:
This updates the fetching of lesson comments to use a field about
comments, rather than indirectly getting them from the lesson field.

Originally suggested here: https://github.com/eggheadio/egghead-next/pull/1148/files#r930385696

Related PR: https://github.com/eggheadio/egghead-rails/pull/5021

comments

### DIFF
--- a/src/lib/lesson-comments.ts
+++ b/src/lib/lesson-comments.ts
@@ -46,31 +46,28 @@ export async function loadLessonComments(
   }
 
   const query = /* GraphQL */ `
-    query getLesson($slug: String!) {
-      lesson(slug: $slug) {
-        comments {
-          comment
-          commentable_id
-          commentable_type
-          created_at
-          id
-          is_commentable_owner
-          state
-          user {
-            avatar_url
-            full_name
-            instructor {
-              first_name
-            }
+    query getLessonComments($slug: String!) {
+      lesson_comments(slug: $slug) {
+        title
+        comment
+        commentable_id
+        commentable_type
+        created_at
+        id
+        is_commentable_owner
+        state
+        user {
+          avatar_url
+          full_name
+          instructor {
+            first_name
           }
         }
       }
     }
   `
 
-  const {
-    lesson: {comments},
-  } = await graphQLClient.request(query, variables)
+  const {lesson_comments} = await graphQLClient.request(query, variables)
 
-  return comments
+  return lesson_comments
 }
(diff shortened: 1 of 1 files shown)


src/lib/lesson-comments.ts
```diff
@@ -46,31 +46,28 @@ export async function loadLessonComments(
...
-          state
-          user {
-            avatar_url
-            full_name
-            instructor {
-              first_name
-            }
+    query getLessonComments($slug: String!) {
+      lesson_comments(slug: $slug) {
+        title
```
jbranchaud, line 51: I think `title` can be removed.

jbranchaud, line 51:
```suggestion
```

Creeland (commented):
Will this break how the data returned from that query is consumed by the frontend? The objects returned are each shaped differently

The returned data from the original query:
```
"data": {
    "lesson": {
      "comments": [
        {
          "comment": "I don't see where \"model Post\" from  schema.prisma is used. \nI would have thought in the seed.ts file it would specify each post in the posts array is of the model \"Post\" in the schema file.",
          "commentable_id": "9368",
          "commentable_type": "Lesson",
          "created_at": "2022-05-26T14:47:33-05:00",
          "id": "33314",
          "is_commentable_owner": false,
          "state": "published",
          "user": {
            "avatar_url": "//gravatar.com/avatar/2d98eefa37de68a73e026e6ca5d94deb.png?s=128&d=mp",
            "full_name": "ed leach",
            "instructor": null
          }
        },
```

And the returned data from the new query:
```
"data": {
    "lesson_comments": [
      {
        "comment": "I don't see where \"model Post\" from  schema.prisma is used. \nI would have thought in the seed.ts file it would specify each post in the posts array is of the model \"Post\" in the schema file.",
        "commentable_id": "9368",
        "commentable_type": "Lesson",
        "created_at": "2022-05-26T14:47:33-05:00",
        "id": "33314",
        "is_commentable_owner": false,
        "state": "published",
        "user": {
          "avatar_url": "//gravatar.com/avatar/2d98eefa37de68a73e026e6ca5d94deb.png?s=128&d=mp",
          "full_name": "ed leach",
          "instructor": null
        }
      },
```

jbranchaud:
> Will this break how the data returned from that query is consumed by the frontend? The objects returned are each shaped differently

The query shape is slightly different, but I account for it in how I extract and return the data at the bottom of the function.

CleanShot 2022-08-10 at 10 05 04@2x

I also smoke tested it to be sure comment still load on lesson pages.

\cc @Creeland

Creeland (approved):

PR #1153, opened 2022-08-02
cleanup: re-remove comments from graphql

jbranchaud:
There must have been a bad merge that re-introduced the comments block
into this graphql query. It was originally removed last week in
https://github.com/eggheadio/egghead-next/pull/1148/files.

Comment loading is now handled elsewhere, so this isn't needed here.

no comment

### DIFF
--- a/src/lib/lessons.ts
+++ b/src/lib/lessons.ts
@@ -156,22 +156,6 @@ const loadLessonGraphQLQuery = /* GraphQL */ `
         slug
         twitter
       }
-      comments {
-        comment
-        commentable_id
-        commentable_type
-        created_at
-        id
-        is_commentable_owner
-        state
-        user {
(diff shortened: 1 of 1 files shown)


src/lib/lessons.ts
```diff
@@ -156,22 +156,6 @@ const loadLessonGraphQLQuery = /* GraphQL */ `
         slug
         twitter
       }
-      comments {
```
jbranchaud, line 159: Comment loading happens on these lines and is then merged in with the results of this graphql request.

https://github.com/eggheadio/egghead-next/blob/main/src/lib/lessons.ts#L77-L81

CleanShot 2022-08-02 at 09 54 25@2x

Creeland (approved):

PR #1154, opened 2022-08-02
feat: request full lesson metadata from Sanity with GROQ

jbranchaud:
_Collaboration: @Creeland and I worked closely on this PR._

The idea behind this PR is to get us into a place where we have near-parity with our data fetching of Lesson metadata from Sanity.

This involved:
- adding some fields to the Lesson document in the Sanity schema
- adding some video resource specific fields to the VideoResource document in the Sanity Schema
- updating the GROQ Lesson Loading query to fetch all the things that are fetched by GraphQL
- derive some data values that can be computed instead of stored in Sanity
- merge the GraphQL and Sanity query results together, with a preference to Sanity, to account for interim data discrepancies
- handle 404 Not Found scenarios, while also throwing an error if a lesson isn't found from either source

What's next:

- [ ] First, apply the schema changes in production.
- [ ] Second, work with @zacjones93 to import a few lessons one-by-one into Sanity to smoke test this more thoroughly.

grock

### DIFF
--- a/src/lib/lesson-comments.ts
+++ b/src/lib/lesson-comments.ts
@@ -66,7 +66,13 @@ export async function loadLessonComments(
     }
   `
 
-  const {lesson_comments} = await graphQLClient.request(query, variables)
+  try {
+    const {lesson_comments} = await graphQLClient.request(query, variables)
 
-  return lesson_comments
+    return lesson_comments
+  } catch (e) {
+    console.log('Error fetching lesson comments: ', e)
+
+    return []
+  }
 }
--- a/src/lib/lessons.ts
+++ b/src/lib/lessons.ts
@@ -11,7 +11,7 @@ import groq from 'groq'
 const lessonQuery = groq`
 *[_type == 'lesson' && slug.current == $slug][0]{
   title,
-  slug,
+  'slug': slug.current,
   description,
   resource->_type == 'videoResource' => {
     ...(resource[]-> {
--- a/src/lib/lessons.ts
+++ b/src/lib/lessons.ts
@@ -11,16 +11,20 @@ const lessonQuery = groq`
   slug,
   description,
   'instructor': collaborators[0]-> {
-    'full_name': person->name,
-    'slug': person->slug.current,
-    'avatar_64_url': person->image.url,
-    'twitter_url': person->twitter
+    ...(person[]-> {
+      'full_name': name,
+      'slug': slug.current,
+      'avatar_64_url': image.url,
+      'twitter_url': twitter
+    }),
   },
   'tags': softwareLibraries[] {
-    'name': library->name,
-    'label': library->slug.current,
-    'http_url': library->url,
-    'image_url': library->image.url
+    ...(library[]-> {
+       name,
+      'label': slug.current,
+      'http_url': url,
+      'image_url': image.url
+    }),
   },
   'collection': *[_type=='course' && references(^._id)][0]{
     title,
@@ -35,7 +39,9 @@ const lessonQuery = groq`
       'title': title,
       'duration': 0,
       'thumb_url': null,
-      'media_url': *[_type=='videoResource' && _id == ^.resource->_id][0].hlsUrl
+      resource->_type == 'videoResource' => {
+        'media_url': resource->hslUrl
+      }
     }
   }
 }`
--- a/src/lib/lessons.ts
+++ b/src/lib/lessons.ts
@@ -5,6 +5,7 @@ import {loadLessonComments} from './lesson-comments'
 import {sanityClient} from 'utils/sanity-client'
 import groq from 'groq'
 import some from 'lodash/some'
+import isEmpty from 'lodash/isEmpty'
 
 // next_up_url can be derived on the front-end from collection and
 // staff_notes_url seems to be null for all lessons
@@ -152,6 +153,12 @@ export async function loadLesson(
     lessonMetadataFromGraphQL,
     lessonMetadataFromSanity,
   )
+
+  // if we aren't able to find Lesson metadata at either source, throw an
+  // error.
+  if (isEmpty(lessonMetadata.slug)) {
+    throw new Error(`Unable to lookup lesson metadata (slug: ${slug})`)
+  }
 
   return {...lessonMetadata, comments} as LessonResource
 }
--- a/src/lib/lessons.ts
+++ b/src/lib/lessons.ts
@@ -165,7 +165,7 @@ const mergeLessonMetadata = (
   lessonMetadataFromSanity: LessonResource,
 ): LessonResource => {
   // we can merge most of it together as is, but there are a few nested pieces
-  // that need to be handled either manually.
+  // that need to be handled manually.
   //
   // e.g. if tags haven't been set yet on Sanity, they will appear as an empty
   // array. With a standard spread, they empty tags from Sanity would override
--- a/src/lib/lessons.ts
+++ b/src/lib/lessons.ts
@@ -76,11 +76,18 @@ async function loadLessonMetadataFromSanity(slug: string) {
     slug,
   }
 
-  const baseValues = await sanityClient.fetch(lessonQuery, params)
+  try {
+    const baseValues = await sanityClient.fetch(lessonQuery, params)
 
-  const derivedValues = deriveDataFromBaseValues(baseValues)
+    const derivedValues = deriveDataFromBaseValues(baseValues)
 
-  return {...baseValues, ...derivedValues}
+    return {...baseValues, ...derivedValues}
+  } catch (e) {
+    // Likely a 404 Not Found error
+    console.log('Error fetching from Sanity: ', e)
+
+    return {}
+  }
 }
 
 // TODO: Derive the next_up_url from the collection and lesson slug
@@ -92,19 +99,36 @@ const deriveDataFromBaseValues = (result: any) => {
   return {http_url, lesson_view_url, download_url}
 }
 
+async function loadLessonMetadataFromGraphQL(slug: string, token?: string) {
+  const graphQLClient = getGraphQLClient(token)
+
+  try {
+    const {lesson: lessonMetadataFromGraphQL} = await graphQLClient.request(
+      loadLessonGraphQLQuery,
+      {slug},
+    )
+
+    return lessonMetadataFromGraphQL
+  } catch (e) {
+    // Likely a 404 Not Found error
+    console.log('Error fetching from GraphQL: ', e)
+
+    return {}
+  }
+}
+
 export async function loadLesson(
   slug: string,
   token?: string,
 ): Promise<LessonResource> {
   token = token || getAccessTokenFromCookie()
-  const graphQLClient = getGraphQLClient(token)
 
   /******************************************
    * Primary Lesson Metadata GraphQL Request
    * ****************************************/
-  const {lesson: lessonMetadataFromGraphQL} = await graphQLClient.request(
-    loadLessonGraphQLQuery,
-    {slug},
+  const lessonMetadataFromGraphQL = await loadLessonMetadataFromGraphQL(
+    slug,
+    token,
   )
 
   /**********************************************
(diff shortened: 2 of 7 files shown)


src/lib/lessons.ts
```diff
@@ -4,23 +4,49 @@ import getAccessTokenFromCookie from '../utils/get-access-token-from-cookie'
...
+    })
+  },
+  'free_forever': isCommunityResource,
+  'path': '/lessons/' + slug.current,
+  'thumb_url': thumbnailUrl,
+  'repo_url': repoUrl,
+  'code_url': codeUrl,
+  'createdAt': eggheadRailsCreatedAt,
+  'updatedAt': eggheadRailsUpdatedAt,
+  'publishedAt': eggheadRailsPublishedAt,
```
Creeland, line 34: In the lesson schema we made explicitly prefixed these field names with `egghead` rails to make it clear that these are values from rails we are tracking. 

We had the thought of using Sanity's field group's feature to separate these fields into their own tab. https://www.sanity.io/docs/field-groups

src/lib/lessons.ts
```diff
@@ -4,23 +4,49 @@ import getAccessTokenFromCookie from '../utils/get-access-token-from-cookie'
...
-    'full_name': person->name,
-    'slug': person->slug.current,
-    'avatar_64_url': person->image.url,
-    'twitter_url': person->twitter
+    ...(person-> {
+      'full_name': name,
+      'slug': slug.current,
+      'avatar_64_url': image.url,
+      'twitter_url': twitter
+    }),
```
Creeland, line 41: This merge reference syntax is more performant than using an `->` for every reference we need off of person. The arrow in syntactic sugar for `*[_id == ^.thing._ref][0]`

src/lib/lessons.ts
```diff
@@ -32,10 +58,12 @@ const lessonQuery = groq`
...
-      'title': title,
-      'duration': 0,
-      'thumb_url': null,
-      'media_url': *[_type=='videoResource' && _id == ^.resource->_id][0].hslUrl
+      title,
+      'thumb_url': thumbnailUrl,
+      resource->_type == 'videoResource' => {
+        'media_url': resource->hslUrl,
+        duration,
+      },
```
Creeland, line 66: this is sanity's conditional syntax. We put this in place to handle future resource types.

src/lib/lessons.ts
```diff
@@ -49,38 +77,155 @@ async function loadLessonMetadataFromSanity(slug: string) {
...
+  lessonMetadataFromSanity: LessonResource,
+): LessonResource => {
+  // we can merge most of it together as is, but there are a few nested pieces
+  // that need to be handled either manually.
+  //
+  // e.g. if tags haven't been set yet on Sanity, they will appear as an empty
+  // array. With a standard spread, they empty tags from Sanity would override
+  // an actual list of tags from graphql. We can instead handle this manually
+  // by checking for `_.some()` and falling back to graphql if there aren't
+  // any.
```
Creeland, line 177: @jbranchaud Have you seen lodash's _.mergeWith method? It may be what we are looking for. It does a deep merge but also lets you compare values using a _customizer_ function, which could be used here to check for empty arrays

src/lib/lessons.ts
```diff
@@ -49,38 +77,155 @@ async function loadLessonMetadataFromSanity(slug: string) {
...
+
   return {...lessonMetadata, comments} as LessonResource
 }
 
+const mergeLessonMetadata = (
+  lessonMetadataFromGraphQL: LessonResource,
+  lessonMetadataFromSanity: LessonResource,
+): LessonResource => {
+  // we can merge most of it together as is, but there are a few nested pieces
+  // that need to be handled either manually.
```
jbranchaud, line 171:
```suggestion
  // that need to be handled manually.
```

src/lib/lessons.ts
jbranchaud: I should write some tests for this since it involves a bit of logic.

studio/schemas/documents/videoResource.js
```diff
@@ -37,6 +37,28 @@ export default {
           return true
         }),
     },
+    {
```
jbranchaud, line 40: The fields we've added to the `videoResource` document used to be attributes associated with a lesson. This was a chance to improve the data model by moving things like duration, transcript, and subtitles to the video resource. This makes the lesson document more flexible, it could be associated with any type of primitive resource, not just a video.

studio/schemas/documents/lesson.js
```diff
@@ -103,6 +103,49 @@ export default {
...
+      title: 'egghead Rails Updated At',
+      description: 'Date this lesson resource last updated on egghead.io',
+      type: 'datetime',
+    },
+    {
+      name: 'eggheadRailsPublishedAt',
+      title: 'egghead Rails Published At',
+      description: 'Date this lesson resource was published on egghead.io',
+      type: 'datetime',
+    },
```
jbranchaud, line 148: The reason we added fields for these timestamp values is that they represent useful metadata that we wouldn't otherwise have in Sanity.

For example, we display when a lesson was last updated on the lesson page. That is based on the `updated_at` field. If a lesson was updated 2 months ago, but we import it from egghead-rails to Sanity today, we don't want to go off Sanity's internal updated attribute, we want to know when it was last actually updated in a meaningful way by an editor/publisher.

Perhaps it would be better to instead add a field called something like _Displayed Updated At_ (or _Last Updated_). That'd make it a little clearer what it is used for and keep it separate from the _internal_ timestamp values. It doesn't need to be named after `eggheadRails...`, we'd just use the value from egghead-rails when first importing it. 🤔

studio/schemas/documents/lesson.js
```diff
@@ -103,6 +103,49 @@ export default {
...
+    {
+      name: 'repoUrl',
+      title: 'Repo URL',
+      type: 'url',
+    },
+    {
+      name: 'codeUrl',
+      title: 'Code URL',
+      type: 'url',
+    },
```
jbranchaud, line 130: These values are still up for discussion: the source code only ever uses one or the other.

The `repoUrl` always ends up being a github URL built by a service. Seems unnecessary for this system to do the same, just past in whatever the URL is.

If we just had `codeUrl`, then it could point to github, codesandbox, jsbin, codepen, or wherever.

I guess the two values are handled slightly differently and/or potentially both displayed. So, something to think about how to best represent in the schema. https://github.com/eggheadio/egghead-next/blob/main/src/pages/lessons/%5Bslug%5D.tsx#L685-L714

src/types.ts
```diff
@@ -31,7 +31,7 @@ export type LessonResource = Resource & {
   completed: boolean
   duration: number
   instructor: any
-  collection: Resource
+  collection: Resource & {lessons: any[]}
```
jbranchaud, line 34: A collection in the context of a LessonResource is a `Resource` with `lessons`.

src/lib/lessons.ts
```diff
@@ -49,38 +77,155 @@ async function loadLessonMetadataFromSanity(slug: string) {
...
+    return {...baseValues, ...derivedValues}
+  } catch (e) {
+    // Likely a 404 Not Found error
+    console.log('Error fetching from Sanity: ', e)
+
+    return {}
+  }
+}
+
+// TODO: Derive the next_up_url from the collection and lesson slug
```
jbranchaud, line 94: We aren't trying to handle `next_up_url` anymore. The lesson page already knows how to do that.

```suggestion
```

src/lib/lessons.ts
```diff
@@ -4,23 +4,49 @@ import getAccessTokenFromCookie from '../utils/get-access-token-from-cookie'
 import {loadLessonComments} from './lesson-comments'
 import {sanityClient} from 'utils/sanity-client'
 import groq from 'groq'
+import some from 'lodash/some'
+import isEmpty from 'lodash/isEmpty'
 
+// next_up_url can be derived on the front-end from collection and
```
jbranchaud, line 10: We aren't trying to handle `next_up_url` anymore. The lesson page already knows how to do that.

```suggestion
```

src/lib/lessons.ts
```diff
@@ -4,23 +4,49 @@ import getAccessTokenFromCookie from '../utils/get-access-token-from-cookie'
...
+// staff_notes_url seems to be null for all lessons
```
jbranchaud, line 11: Not true, I was looking at inaccurate data when I made this claim.

```suggestion
```

src/lib/lessons.ts
jbranchaud: Added tests in this commit https://github.com/eggheadio/egghead-next/pull/1154/commits/7da0426b31d2f042fda1d7422b5fd3c42a7a29b1

src/utils/lesson-metadata.ts
```diff
@@ -0,0 +1,65 @@
+import {LessonResource} from 'types'
+import some from 'lodash/some'
+
+export const mergeLessonMetadata = (
```
jbranchaud, line 4: Move `mergeLessonMetadata` into its own module. That makes it easier to test and declutters the `lib/lesson.ts` module.

src/utils/__tests__/lesson-metadata.test.ts
```diff
@@ -0,0 +1,135 @@
+import {LessonResource} from 'types'
+import {mergeLessonMetadata} from '../lesson-metadata'
```
jbranchaud, line 2: test coverage for the different ways that data is merged in `mergeLessonMetadata`

src/lib/lessons.ts
```diff
@@ -4,23 +4,47 @@ import getAccessTokenFromCookie from '../utils/get-access-token-from-cookie'
...
+    })
+  },
+  'free_forever': isCommunityResource,
+  'path': '/lessons/' + slug.current,
+  'thumb_url': thumbnailUrl,
+  'repo_url': repoUrl,
+  'code_url': codeUrl,
+  'createdAt': eggheadRailsCreatedAt,
+  'updatedAt': eggheadRailsUpdatedAt,
+  'publishedAt': eggheadRailsPublishedAt,
```
jbranchaud, line 32: TODO: we accidentally reverted the camel case on these three, so they aren't going to override the values in the graphql query.

src/lib/lesson-comments.ts
```diff
@@ -66,7 +66,13 @@ export async function loadLessonComments(
...
+  try {
+    const {lesson_comments} = await graphQLClient.request(query, variables)
 
-  return lesson_comments
+    return lesson_comments
+  } catch (e) {
+    console.log('Error fetching lesson comments: ', e)
+
+    return []
+  }
```
jbranchaud, line 77: better "Lesson Comments Not Found" handling

src/lib/lessons.ts
```diff
@@ -49,38 +75,97 @@ async function loadLessonMetadataFromSanity(slug: string) {
...
+  }
+
   return {...lessonMetadata, comments} as LessonResource
 }
 
+// values in the graphql that are being skipped/ignored by Sanity
+//
+// - dash_url - not used
+// - staff_notes_url - not used
+// - icon_url - this is being used by the lesson page
```
jbranchaud, line 167: The `icon_url`, when coming from egghead-rails, is an image attached to the lesson. If one of the lesson tags has a thumb-size icon, we use that. Otherwise, there is a fallback egghead svg.
https://github.com/eggheadio/egghead-rails/blob/d8a9480088e3196c9844250218d4fc146c7b5fcd/app/models/concerns/content_model.rb#L89-L100

We could approximate the same thing client-side by grabbing the tag images and/or using the fallback. Then Sanity doesn't need to know about it. Unless, we want to start doing custom `icon_url` values for lessons.

src/lib/lessons.ts
jbranchaud: I had to snake_case the left-hand-side names, but otherwise, cree's comments still stand.

src/lib/lessons.ts
jbranchaud: fixed here: https://github.com/eggheadio/egghead-next/pull/1154/commits/847f1dc537ac08adda6c81f7a0f5a0fa28dd61a6

studio/schemas/documents/videoResource.js
joelhooks: lesson has been so overloaded forever, this is great

joelhooks (approved):

PR #1157, opened 2022-08-04
prototype custom course page layout for PHP 101

Creeland:
Here we are experimenting with a modules layout for Mark's upcoming PHP course. As a first step towards building page layouts with Sanity, this is also an experiment for handling collections incoming from Sanity with multiple types of potential content. 

Lessons are grouped into individual modules, and their grid layout is dynamically determined based on the number of lessons in a module.

image

image

https://user-images.githubusercontent.com/518406/183513605-d8f53d83-149d-4a26-94d0-ffb6d00b9671.mov

module


src/components/layouts/php-collection-page-layout.tsx
```diff
@@ -0,0 +1,1161 @@
...
+}
+
+const ModulesList = ({modules, render}: {modules: any; render: any}) => {
+  return modules.map((module: any) => {
+    return (
+      <div key={module.id}>
+        <h2 className="text-xl font-bold mt-4 mb-2">{module.title}</h2>
+        {(() => {
+          switch (true) {
+            case module.lessons.length < 5:
```
zacjones93, line 115: Switching on module lesson length to handle where description is put and columns for the lesson list

src/components/layouts/php-collection-page-layout.tsx
```diff
@@ -0,0 +1,1161 @@
...
+    // Manually slicing lessons into modules
+    const lessonModules = [
+      {title: 'Project Setup', lessons: lessons.slice(0, 3)},
+      {title: 'Tags', lessons: lessons.slice(3, 7)},
+      {title: 'Variables and Constants', lessons: lessons.slice(7, 15)},
+      {title: 'Conditionals', lessons: lessons.slice(15, 21)},
+      {title: 'Arrays and Loops', lessons: lessons.slice(21, 29)},
+      {title: 'Functions', lessons: lessons.slice(29, 37)},
+      {title: 'Classes', lessons: lessons.slice(37, 46)},
+    ]
```
zacjones93, line 287: hard coded modules list, we will expect this type of data to come through sanity. We could hard code this in the groq query for this experiment but doing it in the component felt better

src/components/layouts/php-collection-page-layout.tsx
```diff
@@ -0,0 +1,1161 @@
...
+                                                </Markdown>
+                                              </a>
+                                            </Link>
+                                          )}
+                                        </div>
+                                      </li>
+                                    )
+                                  },
+                                )}
+                              </ul>
```
zacjones93, line 1028: This should probably be removed?
```suggestion

```

src/components/layouts/php-collection-page-layout.tsx
```diff
@@ -0,0 +1,1161 @@
...
+              <div>
+                <ModulesList
+                  modules={lessonModules}
+                  render={(lessonList) => (
+                    <LessonsList
+                      lessons={lessonList}
+                      completedLessonSlugs={completedLessonSlugs}
+                    />
+                  )}
+                />
```
zacjones93, line 1047: render props felt appropriate here as the module list shouldn't care how lessons are being rendered, just where in the component they are.

This also avoids having to pass `completedLessonSlugs` to `ModulesList` which isn't a concern of the component

src/components/layouts/php-collection-page-layout.tsx
```diff
@@ -0,0 +1,1161 @@
...
+    return (
+      <div key={module.id}>
+        <h2 className="text-xl font-bold mt-4 mb-2">{module.title}</h2>
+        {(() => {
+          switch (true) {
+            case module.lessons.length < 5:
+              return (
+                <div className={'grid grid-cols-2'}>
+                  <ul className="grid gap-4">{render(module.lessons)}</ul>
+                  <p className="mb-6 prose text-gray-900 dark:prose-dark md:prose-lg md:dark:prose-lg-dark dark:text-gray-100 dark:prose-a:text-blue-300 dark:hover:prose-a:text-blue-200 prose-a:text-blue-500 hover:prose-a-:text-blue-600 ">
```
Creeland, line 119: Description paragraph is currently hardcoded but will later be coming from an external source

src/components/layouts/php-collection-page-layout.tsx
```diff
@@ -0,0 +1,1161 @@
...
+  })
+}
+
+const ModulesList = ({modules, render}: {modules: any; render: any}) => {
+  return modules.map((module: any) => {
+    return (
+      <div key={module.id}>
+        <h2 className="text-xl font-bold mt-4 mb-2">{module.title}</h2>
+        {(() => {
+          switch (true) {
```
Creeland, line 114: Using the switch true pattern to compare the lesson array length in each case

src/components/layouts/php-collection-page-layout.tsx
```diff
@@ -0,0 +1,1161 @@
...
+    )
+  })
+}
+
+const ModulesList = ({modules, render}: {modules: any; render: any}) => {
+  return modules.map((module: any) => {
+    return (
+      <div key={module.id}>
+        <h2 className="text-xl font-bold mt-4 mb-2">{module.title}</h2>
+        {(() => {
```
Creeland, line 113: Wrapped the switch in an IIFE since a statement is expected. We wanted the module title on the outside of the list so we thought this would be a better option. 

Though it might be better to have that h2 repeated in each case

src/components/layouts/php-collection-page-layout.tsx
Creeland: We are unsure of what the best course of action is to make module descriptions look decent if there is nothing in the empty space. Should the text be wider? I've heard that it's best not to exceed a 75 character width

image

src/components/layouts/php-collection-page-layout.tsx
```diff
@@ -0,0 +1,1189 @@
...
+          default:
+            return <ul className="grid grid-cols-1">{children}</ul>
+        }
+      })()}
+    </div>
+  )
+}
+
+const CollectionContent = (props: any) => {
+  return props.contentList.map((content: any, index: number) => {
```
Creeland, line 157: Incoming collections from sanity should have a "content list" array where other collections and resources may live

src/components/layouts/php-collection-page-layout.tsx
```diff
@@ -0,0 +1,1189 @@
...
+    switch (content.type) {
```
Creeland, line 158: Switches on the `type` of content. This is `type` instead of `_type` since we are currently supporting data from graphql

src/components/layouts/php-collection-page-layout.tsx
```diff
@@ -0,0 +1,1189 @@
...
+      case 'module':
+        return (
+          <ModuleCollection module={content}>
+            <CollectionContent {...props} contentList={content.contentList} />
+          </ModuleCollection>
+        )
```
Creeland, line 164: In this case of type `module` we are returning a ModuleCollection component. We are recursing on `CollectionContent` as the child. 

We spread the props and then override the parent contentList with the mapped content's contentList

src/components/layouts/php-collection-page-layout.tsx
```diff
@@ -0,0 +1,1189 @@
...
+      case 'lesson':
+        return (
+          <LessonLinkResource
+            lesson={content}
+            index={index}
+            completedLessonSlugs={props.completedLessonSlugs}
+          />
+        )
```
Creeland, line 172: In the case of the content type being lesson, we return the `LessonLinkResource`. There is no need to recurse here since this isn't a collection and won't have children.

src/components/layouts/php-collection-page-layout.tsx
```diff
@@ -0,0 +1,1189 @@
...
+const ModuleCollection = ({module, children}: {module: any; children: any}) => {
+  return (
+    <div key={module.id}>
+      <h2 className="text-xl font-bold mt-4 mb-2">{module.title}</h2>
+      {(() => {
+        switch (true) {
+          case module.contentList.length < 5:
+            return (
+              <div className={'grid grid-cols-2'}>
+                <ul className="grid gap-4">{children}</ul>
```
Creeland, line 117: Here we render the children. This is so the module can have whatever type of content we support in `CollectionContent`

src/components/layouts/php-collection-page-layout.tsx
Creeland: The idea here is that we can support whatever type of content we have within a collection using recursion and switching on the content type. Child "collections" will recurse and then the types of their content will be handled

src/components/layouts/php-collection-page-layout.tsx
```diff
@@ -0,0 +1,1189 @@
...
+                  Ut enim ad minim veniam, quis nostrud exercitation ullamco
+                  laboris nisi ut aliquip ex ea commodo consequat. Duis aute
+                  irure dolor in reprehenderit in voluptate velit esse cillum
+                  dolore eu fugiat nulla pariatur. Excepteur sint occaecat
+                  cupidatat non proident, sunt in culpa qui officia deserunt
+                  mollit anim id est laborum.
+                </p>
+              </div>
+            )
+          case module.contentList.length > 5 && module.contentList.length <= 10:
```
Creeland, line 130: Here we are switching on the length of the contentList. Items will be displayed differently depending on the number of them in the array. 

In the future we may also need to use conditionals for different content types being present in the contentList

src/components/layouts/php-collection-page-layout.tsx
```diff
@@ -0,0 +1,816 @@
...
+          Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
+          eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad
+          minim veniam, quis nostrud exercitation ullamco laboris nisi ut
+          aliquip ex ea commodo consequat. Duis aute irure dolor in
+          reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla
+          pariatur. Excepteur sint occaecat cupidatat non proident, sunt in
+          culpa qui officia deserunt mollit anim id est laborum.
+        </p>
+        <ul
+          className={`grid sm:grid-rows-${Math.ceil(
```
Creeland, line 125: currently this is just using some simple math to determine number of rows needed. We are using this instead of setting columns so that the lessons are ordered top to bottom in each column instead of alternating between them.

src/components/layouts/php-collection-page-layout.tsx
```diff
@@ -0,0 +1,816 @@
...
+                    </div>
+                    <div className="flex lg:scale-100 md:scale-75 justify-center items-center hidden md:block xl:pr-12">
+                      <Image
+                        className="rounded-full ml-4 justify-center"
+                        src={avatar_url}
+                        layout="fixed"
+                        width="300"
+                        height="300"
+                        alt={`${name}'s avatar`}
+                      />
```
Creeland, line 797: @zacjones93 mostly wrote the instructor introduction. I just adjusted the image's responsiveness

zacjones93: I pulled the introduction from Marks site bio

src/components/layouts/php-collection-page-layout.tsx
joelhooks: traditionally "items" has been the name I've given or seen used for the actual list. 

my problem with contents list is that the items might not be "content"

another option might be "resources" which would be aligned with our collection/resources language

https://www.dropbox.com/s/qg4ppej4az9sulg/egghead%20Templates%20v4.pdf?dl=0

reading this and the description of the "standard list" in the context of a collection, i'd probably lean "resources"

I don't feel super strongly about this and consistency is probably the most important going forawrd!

src/components/layouts/php-collection-page-layout.tsx
joelhooks: "resourceType" perhaps?

src/components/layouts/php-collection-page-layout.tsx
```diff
@@ -0,0 +1,689 @@
...
+import BookmarkIcon from '../icons/bookmark'
+import axios from 'utils/configured-axios'
+import friendlyTime from 'friendly-time'
+import LearnerRatings from '../pages/courses/learner-ratings'
+import CommunityResource from 'components/community-resource'
+import TagList from './tag-list'
+import DialogButton from '../pages/courses/dialog-button'
+import MembershipDialogButton from '../pages/courses/membership-dialog-button'
+
+// imports from collection-page-layout.tsx
```
jbranchaud, line 27: This doesn't need to be said.

```suggestion
```

src/components/layouts/php-collection-page-layout.tsx
```diff
@@ -0,0 +1,689 @@
...
+        resourceList: lessons.slice(15, 21),
+      },
+      {
+        type: 'module',
+        title: 'Arrays and Loops',
+        resourceList: lessons.slice(21, 29),
+      },
+      {type: 'module', title: 'Functions', resourceList: lessons.slice(29, 37)},
+      {type: 'module', title: 'Classes', resourceList: lessons.slice(37, 46)},
+    ]
```
jbranchaud, line 184: nice! Is the idea that if we like this pattern we could eventually encode this shape of data in Sanity?

src/components/layouts/php-collection-page-layout.tsx
```diff
@@ -0,0 +1,689 @@
...
+        >
+          {children}
+        </ul>
+      </div>
+    </div>
+  )
+}
+
+const ResourceCollection = (props: any) => {
+  return props.resourceList.map((resource: any, index: number) => {
```
jbranchaud, line 128: since this is a map, do both of the rendered children need `key={index}` props?

src/components/layouts/php-collection-page-layout.tsx
```diff
@@ -0,0 +1,689 @@
...
+    switch (resource.type) {
+      case 'module':
+        return (
+          <ModuleCollection module={resource}>
+            <ResourceCollection
+              {...props}
+              resourceList={resource.resourceList}
```
jbranchaud, line 135: What would you think of pulling `resource.resourceList` out into a const right above the return on 131 and naming it something like `nestedResourceList` or `childResourceList`? I think that would help drive home the recursive rendering scheme. It would also make it less confusing (for me) because I can clearly see that `props.resourceList` and `resource.resourceList` are distinct things.

src/components/layouts/php-collection-page-layout.tsx
Creeland: Yeah this is the idea! All collections should have a "resourceList" array field

src/components/layouts/php-collection-page-layout.tsx
```diff
@@ -0,0 +1,713 @@
...
+        <p className="mb-6 prose text-gray-900 dark:prose-dark md:prose-lg md:dark:prose-lg-dark dark:text-gray-100 dark:prose-a:text-blue-300 dark:hover:prose-a:text-blue-200 prose-a:text-blue-500 hover:prose-a-:text-blue-600 ">
+          {module.description}
+        </p>
+        <ul
+          className="grid sm:grid-flow-col gap-4"
+          style={{
+            gridTemplateRows: `repeat(${Math.ceil(
+              module.resourceList.length / 2,
+            )}, minmax(0, 1fr))`,
+          }}
```
DrShpongle, line 114: The only solution I found is to use inline styles here.
Arbitrary values for custom tailwindcss classnames should be predefined and explicitly written, so we can't use something like 
```
grid-rows-[repeat(${Math.ceil(module.resourceList.length / 2)},_minmax(0,_1fr))]
```

jbranchaud (approved):
nice work 🌮 !

PR #1161, opened 2022-08-11
chore: Improve the type safety of php-collection-page-layout.tsx

Creeland:
security monitor


src/components/layouts/php-collection-page-layout.tsx
```diff
@@ -41,13 +41,24 @@ type CoursePageLayoutProps = {
...
 }
 
+type ModuleResource = {
+  type: 'module'
+  resourceList: ResourceList
+  title: string
+  description: string
+}
+
+type NestedResource = ModuleResource | LessonResource
```
Creeland, line 51: Made a union type here so that the individually mapped resources can be typed properly

src/components/layouts/php-collection-page-layout.tsx
```diff
@@ -137,7 +146,7 @@ const ResourceCollection = ({
 }) => {
   return (
     <>
-      {resourceList.map((resource: any, index: number) => {
+      {resourceList.map((resource: NestedResource, index: number) => {
```
joelhooks, line 149: this could be as simple as `{type: string, resourceList: Resource[]}` since those are the properties you need below, then it'd be only restrictive in the sense that "whatever I pass in here needs to look like this" versus a larger more complete structure that won't even be accessed.

src/components/layouts/php-collection-page-layout.tsx
```diff
@@ -120,35 +135,36 @@ const ModuleCollection = ({module, children}: {module: any; children: any}) => {
...
 }: {
-  resourceList: ResourceList
+  resourceList: Resource[]
   completedLessonSlugs: string[]
 }) => {
   return (
     <>
-      {resourceList.map((resource: any, index: number) => {
+      {resourceList.map((resource: Resource, index: number) => {
+        if ('resourceList' in resource) {
```
Creeland, line 148: I added a type guard here since some types that `resource` could be won't contain a `resourceList` since they're either an existing type or in the future an "atomic" type like `videoResource`

joelhooks: suggestion: use a type that would show you an error on the calling side:

`(resourceWithList: {type: 'module' | 'lesson', resourceList: Resource[]}, index: number)`

otherwise you might as well switch back to `any` imo because this isn't providing any type benefit and adds conditional complexity

jbranchaud:
It looks like you’re trying to achieve a few different things in this PR:
1. Upgrade `ModuleCollection` and `LessonLinkResource` with better types so that we aren’t just using `any`s.
2. Rename `NestedResource` to `ModuleResource` (good catch, that’s an improvement!)

After that, I see some reorganization of the structure of the types which seems to be leading to a refactoring of the `ResourceCollection` JSX. I’m less clear about the “why” on those changes. Am I reading this PR right? Could you say more about what else you’re trying to do?

joelhooks:
Thanks for this line of discussion Josh. 


Stepping back from my fine-grained nits to the broader context of "why?" 💯

jbranchaud (approved):

PR #1163, opened 2022-08-12
update: php course page layout

vojtaholik:
screenshot

gif


joelhooks:
I like the lists over the grids, but i feel like "what's going on here" is probably lost to most learners and the instructor block inline with the sections adds to my confusion

where do I start? 

a common affordance is to label page sections, and we are boxing up the sections of the module but not doing much to section out the course presentation and give learners a clear starting point/next step for interacting with the learning material

joelhooks:
I wouldn't hold this PR over the above just a general line of consideration. Might be worth doing some user testing and see how actual learners interact with the updated structure.

joelhooks (approved):

PR #1164, opened 2022-08-12
fix: omit nil values when merging lesson metadata

jbranchaud:
@vojtaholik reported that the Lesson Viewed checkmark wasn't working and tracked it down the the `deriveDataFromBaseValues` function in `src/lib/lessons.ts`.

There were a couple issues which this PR fixes:
1. We weren't checking whether the `path` was an actual value before computing fully-qualified URLs from it. This meant that we were often passing back `https://egghead.io/api/v1undefined/views` as the URL.
2. It was my understanding that ES6 destructured merges (`{ ...obj1, ...obj2 }`) would not allow `null` and `undefined` values in `obj2` to override actual values in `obj1`. This is not the case. So, we had `null` and `undefined` values from Sanity overriding values from graphql that it should have instead been deferring to.

To address **1**, I check the path before computing and passing back those derived values. For **2**, I added a `compactedMerge` function that winnows the `null` and `undefined` values before the lesson metadata objects are merged.

winnow


joelhooks:
`null` in particular is a problem in JS. we utilize `pickBy` in lodash in other places to strip null/undefined

src/utils/compacted-merge.ts
```diff
@@ -0,0 +1,16 @@
...
+// of `null` and `undefined` to override actual values. This compactedMerge
+// function first removes the `null` and `undefined` key-value pairs from each
+// object before merging them.
+//
+// This ensures that if the first object has a key-value pair like
+// `lesson_view_url: '/some/url'` that it doesn't get overridden by the second
+// object having `lesson_view_url: undefined`.
+const compactedMerge = (obj1: object, obj2: object): object => {
+  return {...omitBy(obj1, isNil), ...omitBy(obj2, isNil)}
+}
```
joelhooks, line 14: `pickBy` is my goto for this. Under the hood it uses `identity` and is very effective for removing `null` and `undefined` 
Screen Shot 2022-08-12 at 09 26 39

jbranchaud: oh nice, didn't know about `pickBy`. That's a more direct way to do what I'm doing here. I'll update it.

src/lib/lessons.ts
```diff
@@ -80,7 +82,7 @@ async function loadLessonMetadataFromSanity(slug: string) {
 
     const derivedValues = deriveDataFromBaseValues(baseValues)
 
-    return {...baseValues, ...derivedValues}
+    return compactedMerge(baseValues, derivedValues)
```
jbranchaud, line 85: more reliable merging of base values with derived values

src/lib/lessons.ts
```diff
@@ -89,12 +91,22 @@ async function loadLessonMetadataFromSanity(slug: string) {
   }
 }
 
-const deriveDataFromBaseValues = (result: any) => {
-  const http_url = `${process.env.NEXT_PUBLIC_DEPLOY_URL}${result.path}`
-  const lesson_view_url = `${process.env.NEXT_PUBLIC_AUTH_DOMAIN}/api/v1${result.path}/views`
-  const download_url = `${process.env.NEXT_PUBLIC_AUTH_DOMAIN}/api/v1${result.path}/signed_download`
+// TODO: Move this into `src/utils/lesson-metadata.ts` and add tests.
+const deriveDataFromBaseValues = ({path}: {path: string}) => {
+  if (!isEmpty(path)) {
```
jbranchaud, line 96: Only compute this derived URLs if the incoming `path` value is not empty.

src/lib/lessons.ts
```diff
@@ -89,12 +91,22 @@ async function loadLessonMetadataFromSanity(slug: string) {
...
+    invariant(
+      path.startsWith('/'),
+      'Path value must begin with a forward slash (`/`).',
+    )
```
jbranchaud, line 100: invariant check that the path begins with a forward slash, otherwise we'll end up with malformed URLs

Going to add this as a validation check on the Sanity schema `path` attribute as well.

src/utils/__tests__/compacted-merge.test.ts
```diff
@@ -0,0 +1,93 @@
+import compactedMerge from '../compacted-merge'
+
+test('it can merge two objects together', () => {
```
jbranchaud, line 3: tests for this new `compactedMerge` function

src/utils/compacted-merge.ts
```diff
@@ -0,0 +1,15 @@
...
+// of `null` and `undefined` to override actual values. This compactedMerge
+// function first removes the `null` and `undefined` key-value pairs from each
+// object before merging them.
+//
+// This ensures that if the first object has a key-value pair like
+// `lesson_view_url: '/some/url'` that it doesn't get overridden by the second
+// object having `lesson_view_url: undefined`.
+const compactedMerge = (obj1: any, obj2: any): any => {
+  return {...pickBy(obj1), ...pickBy(obj2)}
+}
```
jbranchaud, line 13: general-purpose utility function for combining metadata objects coming from two places

src/utils/lesson-metadata.ts
```diff
@@ -51,7 +52,7 @@ export const mergeLessonMetadata = (
     ? primaryInstructor
     : secondaryInstructor
 
-  const rest = {...secondaryRest, ...primaryRest}
+  const rest = compactedMerge(secondaryRest, primaryRest)
```
jbranchaud, line 55: compacted merge so that we don't have something like `duration: null` overriding an actual `duration: 123` value.

Creeland (approved):

src/utils/compacted-merge.ts
joelhooks: this is a problem with graphql as well, where `null` the devil shows up

`null` is an object and super weird

image

PR #1166, opened 2022-08-13
build: add build script for dev w/o sitemaps

jbranchaud:
Whenever I run the `yarn build` command locally, I takes longer then it
needs to because it is generating a bunch of sitemap files and then
afterward I have to manually cleanup those sitemap files. This commit
adds a `dev-build` script that does just the testing and `next build`
without all the sitemap noise.

Meanwhile, this leaves the `build` script alone to continue doing what is expected in CI / Vercel.

noise

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,8 +7,9 @@
   "private": true,
   "scripts": {
     "dev": "next dev",
+    "dev-build": "yarn test:ci && next build",
     "prebuild": "yarn make-tags-sitemap",
-    "build": "yarn test:ci && next build",
+    "build": "yarn dev-build",
     "postbuild": "NODE_ENV=production next-sitemap",
     "build:style": "tailwind build src/styles/index.css -o src/styles/output.css",
     "start": "next start",


package.json
```diff
@@ -7,8 +7,9 @@
   "private": true,
   "scripts": {
     "dev": "next dev",
+    "dev-build": "yarn test:ci && next build",
     "prebuild": "yarn make-tags-sitemap",
-    "build": "yarn test:ci && next build",
+    "build": "yarn dev-build",
```
jbranchaud, line 12: Here I add `dev-build` to do what `build` was doing. Then `build` invokes `yarn dev-build` to get those same steps without duplicating the command. The `pre` and `post` steps, which do the sitemap generation, will still run for `build`, but not for `dev-build`.

joelhooks: i think build:dev would be more typical (or dev:build")

joelhooks (approved):

PR #1172, opened 2022-08-16
feat: add icon_url to the GROQ lesson query

jbranchaud:
The coalesce function (https://www.sanity.io/docs/groq-functions#80d6f2c07a20) supported by GROQ allows us to try to use the first software library icon, or if that is null, then fallback to the eggo default icon.

coa

### DIFF
--- a/src/lib/lessons.ts
+++ b/src/lib/lessons.ts
@@ -27,6 +27,7 @@ const lessonQuery = groq`
   'free_forever': isCommunityResource,
   'path': '/lessons/' + slug.current,
   'thumb_url': thumbnailUrl,
+  'icon_url': 'https://res.cloudinary.com/dg3gyk0gu/image/upload/v1567198446/og-image-assets/eggo.svg',
   'repo_url': repoUrl,
   'code_url': codeUrl,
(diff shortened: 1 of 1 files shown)


src/lib/lessons.ts
```diff
@@ -27,6 +27,7 @@ const lessonQuery = groq`
   'free_forever': isCommunityResource,
   'path': '/lessons/' + slug.current,
   'thumb_url': thumbnailUrl,
+  'icon_url': coalesce(softwareLibraries[0].library->image.url, 'https://res.cloudinary.com/dg3gyk0gu/image/upload/v1567198446/og-image-assets/eggo.svg'),
```
joelhooks, line 30: 🔥

one problem here is that we aren't ranking the libraries which we do on the egghead-rails side so if multiple libs are associated we need to make sure they are "manually ranked" so that a "react" lesson doesn't present as some side topic

joelhooks (approved):

src/lib/lessons.ts
jbranchaud: Interesting! I didn't realize ranking was factoring into this on the egghead-rails side. I'll give that some thought.

PR #1177, opened 2022-08-17
fix: grab duration off resource ref

jbranchaud:
The `duration` is an attribute on the `videoResource`, so it needs to be
accessed off that ref. As is, it was trying to access it off the collection
and inevitably just getting `null`.

reference

### DIFF
--- a/src/lib/lessons.ts
+++ b/src/lib/lessons.ts
@@ -65,7 +65,7 @@ const lessonQuery = groq`
       'thumb_url': thumbnailUrl,
       resource->_type == 'videoResource' => {
         'media_url': resource->hlsUrl,
-        duration,
+        'duration': resource->duration,
       }
     }
   }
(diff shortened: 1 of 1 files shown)


src/lib/lessons.ts
```diff
@@ -64,8 +64,10 @@ const lessonQuery = groq`
       title,
       'thumb_url': thumbnailUrl,
       resource->_type == 'videoResource' => {
-        'media_url': resource->hlsUrl,
-        duration,
+        ...(resource-> {
+          'media_url': hlsUrl,
+          duration
+        })
```
jbranchaud, line 70: What's up with `...(resource-> { })`?

This GROQ trick allows you to grab the `resource` once and access
multiple values off of that. The way it was before, I believe under the
hood it does the equivalent of a subquery to get each ref attribute.

Creeland (approved):

PR #1181, opened 2022-08-19
feat: add essential fields to Sanity course schema

jbranchaud:
After @creeland and I walked through the Course page and what the other
course metadata queries are requesting, we updated the Course schema
with some additional essential fields.

- Illustration / Image (same thing)
- We combined the concept of _Community Resource_ into access-level
- Search Indexing State (hidden/indexed)

Beyond that we had already done a comprehensive job with the schema.

There are several fields that the `Resource` supports and which the GROQ query is currently fetching that we decided to disregard for now. Those include:
- essentialQuestions
- pairWithResources
- customOgImage
- features (first)
- prerequisites
- projects

Our rationale is that these can be easily added later on if there is demand for those. We want to err on the side of adding less rather than more to this `Course` schema.

More details behind all of this can be found here: https://roamresearch.com/#/app/egghead/page/rS6yS6uBM

course

### DIFF
--- a/studio/schemas/documents/course.js
+++ b/studio/schemas/documents/course.js
@@ -69,8 +69,8 @@ export default {
     },
     {
       name: 'image',
-      description: 'Links to a full-sized primary image',
-      title: 'Image Url',
+      description: 'Links to a full-sized primary image/illustration',
+      title: 'Image/Illustration Url',
       type: 'url',
     },
     {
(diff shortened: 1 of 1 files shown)


studio/schemas/documents/course.js
```diff
@@ -126,9 +126,31 @@ export default {
         list: [
           {title: 'Free', value: 'free'},
           {title: 'Pro', value: 'pro'},
+          {
+            title: 'Community Resource',
+            value: 'community-resource',
+            description: 'Free Forever',
+          },
```
Creeland, line 133: Moved Community Resource into the access level field. The reasoning behind that is that community resources are a special case of free where they are considered free forever. It's also impossible for a resource to both be PRO and a community resource at the same time

Creeland (approved):

PR #1184, opened 2022-08-23
feat: use zod to validate and type Course/Lesson data from Course Builder

jbranchaud:
This is an experiment to introduce zod as a schema validator for data coming into the API layer supporting Course Builder.

This replaces hand-rolled types with types that are generated from zod schemas. These zod schemas (one for _course_ data and one for _lesson_ data) are used to validate the data in the API request. They will throw right away if anything is off (`courseSchema.parse(...)`), otherwise everything downstream knows the shape of the data to expect.

This is more of an RFC than anything else. This is my first time using Zod, so definitely open to better ways of doing things.

zod


src/pages/api/sanity/lessons/create.ts
```diff
@@ -75,6 +67,22 @@ const sanityIdForDocumentType = async (
   return `${documentType}-${id}`
 }
 
+const courseSchema = z.object({
+  title: z.string(),
+  collaboratorId: z.string().optional(),
+  topicIds: z.string().array(),
+})
+export type CourseData = z.infer<typeof courseSchema>
```
jbranchaud, line 75: First, define a `courseSchema` that we can use to parse incoming request data in the main function.

Second, `infer` a type from that schema that can replace the `CourseData` type we are importing. Very little code needs to change since this type is one-for-one with the other type.

src/pages/api/sanity/lessons/create.ts
```diff
@@ -180,7 +188,8 @@ const createSanityLessons = async (
     } else {
       let transaction = sanityClient.transaction()
 
-      const {course, lessons} = req.body
+      const course = courseSchema.parse(req.body.course)
+      const lessons = z.array(lessonSchema).parse(req.body.lessons)
 
       const {sanityCourse, sanityLessons, sanityResources} =
         await formatSanityMutationForLessons(course, lessons)
```
jbranchaud, line 195: Validate incoming request data before passing it along to the mutation formatter. If anything is off, it will throw. If extra, unexpected data is coming in, that will be stripped out by the `parse`. Clean!

src/types.ts
```diff
@@ -149,9 +149,3 @@ export type ParityCouponMessageProps = {
   isPPP?: boolean
   isLoading?: boolean
 }
-
-export type CourseData = {
-  title: string
-  collaboratorId?: string
-  topicIds: string[]
-}
```
jbranchaud, line 157: Don't need this type now, just gonna use the zod inferred one

src/pages/api/sanity/lessons/create.ts
Creeland: I haven't seen Zod before. It looks a lot like Yup to me https://github.com/jquense/yup

PR #1185, opened 2022-08-23
feat: improve form UX with better messaging

jbranchaud:
- cleanup: remove unused FileUpload type
- feat: add alert about how to use uploader for now
- feat: explicit fail response when Sanity POST fails
- feat: display success/fail message after submit

image

ux

### DIFF
--- a/src/pages/upload/index.tsx
+++ b/src/pages/upload/index.tsx
@@ -13,13 +13,6 @@ import _find from 'lodash/find'
 import {CourseData} from 'types'
 import useFileUploadReducer from 'hooks/use-file-upload-reducer'
 
-type FileUpload = {
-  file: File
-  percent: number
-  message: string
-  signedUrl: string | undefined
-}
-
 type Instructor = {
   _id: string
   eggheadInstructorId: string
--- a/src/pages/upload/index.tsx
+++ b/src/pages/upload/index.tsx
@@ -288,6 +288,7 @@ const Upload: React.FC<
             <label className="block text-sm font-medium text-gray-700">
               Video Files
             </label>
+            <VideoUploadAlert />
             {/* Drop zone UI adapted from https://larainfo.com/blogs/tailwind-css-drag-and-drop-file-upload-ui */}
             <div className="max-w-xl">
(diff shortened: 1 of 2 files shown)


src/pages/upload/index.tsx
```diff
@@ -116,20 +113,31 @@ const UploadWrapper = ({
           lessons: values.lessons,
         })
 
+        setSubmitResponse(response)
```
jbranchaud, line 116: capture the response from the sanity POST so that we can display a Success/Fail message from it.

src/pages/upload/index.tsx
```diff
@@ -191,7 +199,9 @@ const Upload: React.FC<
   const isIncomplete =
     noAttachedFiles || anyFilesCurrentlyUploading || values.course.title === ''
 
-  const submitDisabled = isIncomplete || isSubmitting
+  const submitWasSuccessful = submitResponse?.status === 200
+
+  const submitDisabled = isIncomplete || isSubmitting || submitWasSuccessful
```
jbranchaud, line 204: Disable the submit button if the form submitted successfully, don't want to allow a double-submit after the fact.

src/pages/upload/index.tsx
```diff
@@ -295,6 +305,7 @@ const Upload: React.FC<
             <label className="block text-sm font-medium text-gray-700">
               Video Files
             </label>
+            <VideoUploadAlert />
```
jbranchaud, line 308: CleanShot 2022-08-23 at 16 42 37@2x

Display some information to avoid failed uploads (for now, still root-causing this).

src/pages/upload/index.tsx
```diff
@@ -371,4 +383,134 @@ const Upload: React.FC<
...
+            </div>
+            <div className="mt-4">
+              <div className="-mx-2 -my-1.5 flex">
+                <a
+                  href={studioUrl}
+                  className="bg-green-50 px-2 py-1.5 rounded-md text-sm font-medium text-green-800 hover:bg-green-100 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-offset-green-50 focus:ring-green-600"
+                >
+                  View Studio
+                </a>
+              </div>
```
jbranchaud, line 431: Link to the environment-appropriate Sanity studio after a successful upload so that an editor can easily navigate there to work on the rest of the metadata.

Creeland (approved):
The alerts are a good addition while we are figuring out what's causing the uploads to fail

PR #1186, opened 2022-08-24
feat: add Sanity webhook projects to git

jbranchaud:
These, unfortunately, have to be managed in the Sanity Studio web UI. I
want to make sure we have a copy of them available in the codebase
though. This adds a directory for storing that kind of file. We'll have
to be sure to manually keep these up-to-date, but it is better than
nothing.

I pinged Sanity about this [on twitter](https://twitter.com/jbrancha/status/1562452507141361672?s=20&t=wgPedSv53RSRqMfhDKaSzg) and they said they're looking into it. So maybe this will be a future feature 🤞 

store


Creeland (approved):
It may be confusing to determine which one is the source of truth if they get out of sync. But I do agree it's better than nothing.

PR #1187, opened 2022-08-24
feat: set prodProcessState on course through API

jbranchaud:
According to the Sanity docs, `initialValue` settings in an
object/document schema are only used when a record is created through
the Studio. These are not applied when creating an object through the
API.

So, to get our Course Builder courses to start in the `new` state, we
have to set that manually.

Source: https://www.sanity.io/docs/initial-value-templates

def


Creeland (approved):
Interesting. I'm not sure if I like that behavior, but it makes sense

PR #1189, opened 2022-08-25
Repo URL and Description metadata fields in /upload form

Creeland:
During a test run of the course builder, editors noted that they wouldn't be able to add descriptions and repo urls to lessons on creation.

This PR adds these fields to the lessons in the form, saving editors time since they won't have to immediately navigate to the Sanity studio and add this data to each lesson document in the course.

image

metaverse

### DIFF
--- a/src/pages/upload/index.tsx
+++ b/src/pages/upload/index.tsx
@@ -30,8 +30,6 @@ type UploadedFile = {
 
 type LessonMetadata = {
   title: string
-  description?: string
-  repoUrl?: string
   fileMetadata: UploadedFile
 }
 
@@ -98,6 +96,8 @@ const UploadWrapper = ({
   const initialValues: FormProps = {
     course: {
       title: '',
+      description: '',
+      repoUrl: '',
       collaboratorId: undefined,
       topicIds: [],
     },
@@ -175,8 +175,6 @@ const Upload: React.FC<
       if (existingLesson) {
         return {
           title: existingLesson.title,
-          description: existingLesson.description,
-          repoUrl: existingLesson.repoUrl,
           fileMetadata: {
             ...existingLesson.fileMetadata,
             signedUrl: existingLesson.fileMetadata.signedUrl || file.signedUrl,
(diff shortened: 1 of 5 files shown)


jbranchaud:
@Creeland Both description and repo URL should be Lesson-level attributes. We might need to add a Course-level description at some point, but my understanding of when that gets written suggests that it is best to happen in Sanity for now.

I'm not sure if we'll eventually want repo URL on the Course-level or not. Neither Sanity Resources nor egghead-rails Playlists have a repo URL attribute. There is a code_url on playlists, but that is only used across 8 records.

src/pages/upload/index.tsx
```diff
@@ -229,7 +235,7 @@ const Upload: React.FC<
               htmlFor="course-name"
               className="block text-sm font-medium text-gray-700"
             >
-              Course Name
+              Course Name*
```
jbranchaud, line 238: Is this a "required field" marker? Or just a stray character?

studio/schemas/documents/lesson.js
```diff
@@ -54,6 +55,13 @@ export default {
       description:
         'This can be used to provide a short description of the lesson.',
     },
+    {
+      name: 'repoUrl',
+      type: 'url',
+      title: 'Github Repository Url',
+      description:
+        "A link to the Github repository where the lesson's code is hosted",
+    },
```
jbranchaud, line 64: we already have a `repoUrl` field in here, but it could benefit from the better title and description fields

jbranchaud (changes requested):
One change requested, otherwise looks good to me 🎸 

Also, would you mind adding a small "what" and "why" description to the PR and linking to any relevant external source, like the roam notes on this? It's super valuable to leave some context and breadcrumbs on PRs in the event that we come back to them later wondering about some changes. I find myself doing that a lot.

src/pages/upload/index.tsx
Creeland: I had that there to denote a required field

jbranchaud (approved):
🏗

PR #1190, opened 2022-08-25
feat: try metadata query against Sanity Course doc

jbranchaud:
The goal is to get egghead-next set up to read from both Course and
Resource in tendem when requesting data for the Course page. If it is a
course that lives in the 'Course' document, then use that. Otherwise,
fallback to 'Resource', where most course metadata is living.

This is a first step. The load order and logic will probably shift around at some point, but I wanted to get this in place as a starting point so that we can test out a 'Course' course.

cours


src/lib/courses.ts
```diff
@@ -69,22 +69,83 @@ const courseQuery = groq`
 }
 `
 
+// TODO: go through the minimum Course page requirements, find the overlap with
+// what this Course document supports, and then lineup the naming.
+const courseQuery = groq`
+*[_type == 'course' && slug.current == $slug][0]{
```
jbranchaud, line 75: Still working on the shape of this groq query, some details here: https://roamresearch.com/#/app/egghead/page/R-ebGysU7

PR #1191, opened 2022-08-29
feat: combine community resource into access state

jbranchaud:
- feat: Community Resource is an Access Level option

This way we can mark a Lesson (as well as a Course) as either `Pro`,
`Free`, or `Community Resource` ("Free Forever"). Essentially use an
enum rather than several booleans.

- feat: check free & community resource access state

Forward-compatible way of checking if the course is a free Community
Resource and if so, that we should display Community Resource messaging
on the page.

Forward-compatible because the Sanity Studio options for access-level on
a course is Pro / Free / Community Resource. If it is one of the
latter two, then we want to display it as a Community Resource.
Eventually, we can update the display logic to differentiate between
(temporarily) Free content and Community Resource (free-forever)
content.

community

### DIFF
--- a/src/components/layouts/collection-page-layout.tsx
+++ b/src/components/layouts/collection-page-layout.tsx
@@ -222,6 +222,10 @@ const CollectionPageLayout: React.FunctionComponent<CoursePageLayoutProps> = ({
     tags: railsTags = [],
   } = course
 
+  const communityResource = ['free', 'community-resource'].includes(
+    access_state,
+  )
+
   const sanityTagsPresent = () => {
     return sanityTags.length > 0
   }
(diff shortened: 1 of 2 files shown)


src/components/layouts/collection-page-layout.tsx
```diff
@@ -222,6 +222,10 @@ const CollectionPageLayout: React.FunctionComponent<CoursePageLayoutProps> = ({
     tags: railsTags = [],
   } = course
 
+  const communityResource = ['free', 'community-resource'].includes(
```
joelhooks, line 225: there is no "community resource" state. it's either free or pro, that's a label for free.

joelhooks: "community resource" is marketing not database

jbranchaud:
Replacing this PR with https://github.com/eggheadio/egghead-next/pull/1195

PR #1192, opened 2022-08-29
feat: add list of Related Content to course schema

jbranchaud:
This allows us to tag a Course with _Related Content_ which for now is `Course` and `Resource` documents but could eventually include content like Articles, Podcasts, and even individual Lessons.

Roam: https://roamresearch.com/#/app/egghead/page/mruiUHoQP

CleanShot 2022-08-31 at 11 54 20@2x

related


jbranchaud:
This PR has been updated to name the field 'Related Content' which is more expansive, allowing eventually for associating things like Articles, Podcasts, etc.

Creeland (approved):

PR #1193, opened 2022-08-29
feat: adds first course software library to lesson

Creeland:
Small PR that adds a course's primary softwareLibrary to each lesson. 

> "80% of the time" the tag for the Course is what all the Lessons will end up being tagged with. So this tag is safe to generally apply to each of the lessons. The publisher can change this for individual lessons on an as-needed basis in Sanity
- [Test-Run Roam Notes](https://roamresearch.com/#/app/egghead/page/wanWMtJ5k)

library

### DIFF
--- a/src/pages/api/sanity/lessons/create.ts
+++ b/src/pages/api/sanity/lessons/create.ts
@@ -36,6 +36,7 @@ type SanityLesson = {
   title: string
   description?: string
   repoUrl?: string
+  softwareLibraries: SanitySoftwareLibrary[]
   slug: SanitySlug
   resource: SanityReference
 }
@@ -163,6 +164,8 @@ async function formatSanityMutationForLessons(
       )
       const {description = '', repoUrl = ''} = lesson
 
+      const topicKey = await nanoid()
+
       sanityLessons.push({
         _id: lessonId,
         _type: 'lesson',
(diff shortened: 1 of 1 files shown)


src/pages/api/sanity/lessons/create.ts
```diff
@@ -36,6 +36,7 @@ type SanityLesson = {
   title: string
   description?: string
   repoUrl?: string
+  softwareLibraries: SanitySoftwareLibrary[]
```
Creeland, line 39: Chose to have this as an array since that reflects the data type in Sanity. We may also want to expand on lesson tagging in the upload UI in the future, potentially giving lessons multiple tags

src/pages/api/sanity/lessons/create.ts
```diff
@@ -174,6 +177,16 @@ async function formatSanityMutationForLessons(
           _type: 'reference',
           _ref: videoId,
         },
+        softwareLibraries: [
+          {
+            _type: 'versioned-software-library',
+            _key: topicKey,
+            library: {
+              _type: 'reference',
+              _ref: topicIds[0],
```
Creeland, line 186: Grabs the first topicId in the array since we are considering this the "primary tag"

src/pages/api/sanity/lessons/create.ts
jbranchaud: I expect we will eventually add support for multiple tags in the Course Builder UI. That would involve pulling in, configuring, and styling something like `react-select`. Seems best to punt on that until we are in more of a v2 frame of mind.

PR #1196, opened 2022-09-01
Switch mixpanel tracking logic to use interval instead of amount

zacjones93:
@nicollguarnizo and I were digging around in our analytics and ended up going through the subscription purchase to upgrade flow.

We noticed that the object that stripe sends back doesn't include any amounts and has those set to null or they don't exist. 

For example, the `previous_attributes.plan.amount` doesn't exist.
```json
"previous_attributes": {
    ...
    "plan": {
      "id": "price_1IJSQ12nImeJXwdJ9WyUxdBZ",
      "created": 1613000905,
      "interval": "month",
      ...
    }
  }
```
As well as the current plans `items.data[0].plan?.amount` is set to `null`.
```json
    "items": {
      "object": "list",
      "data": [
        {
          "id": "si_MLzwOpt9msxUAA",
          "object": "subscription_item",
          "billing_thresholds": null,
          "created": 1662056237,
          "metadata": {
          },
          "plan": {
            "id": "price_1IJSO22nImeJXwdJHKrb5j2I",
            "object": "plan",
            "active": true,
            "aggregate_usage": null,
            "amount": null,     <-----
            ...
          },
      }],
      ...
    },
```
This forces the webhook to send `Subscription Downgrade` events because `0 < 0` is returning false.

This PR switches the logic to check the interval that is set. The options we have are `month`, `quarter`, and `year`. If `month` is the previous attribute we know we have an upgrade. If `year` is the previous attribute we know we have a downgrade. If `quarter` is the previous attribute we check current attribute for `month` or `year` and downgrade/upgrade accordingly.

The logic is not pretty.

it works

### DIFF
--- a/src/pages/api/stripe/webhook.ts
+++ b/src/pages/api/stripe/webhook.ts
@@ -11,19 +11,29 @@ function stripeToMixpanelDataConverter(stripeDate: number) {
   return date.toISOString()
 }
 
-type PlanInterval = 'month' | 'year' | 'quarter'
+const UPGRADE = 'Upgrade'
+const DOWNGRADE = 'Downgrade'
+
+const intervalOptions = z.union([
+  z.literal('month'),
+  z.literal('quarter'),
+  z.literal('year'),
+])
+
+type Interval = z.infer<typeof intervalOptions>
+type IntervalChange = typeof UPGRADE | typeof DOWNGRADE
 
 function checkForUpgrade(
-  previousPlanInterval: PlanInterval,
-  newPlanInterval: PlanInterval,
-) {
-  const previousPlanIntervalUpgrades = {
+  previousPlanInterval: Interval,
+  newPlanInterval: Interval,
+): IntervalChange {
+  const isAnUpgrade = {
     month: true,
     year: false,
     quarter: newPlanInterval === 'year',
-  }
+  }[previousPlanInterval]
 
-  return previousPlanIntervalUpgrades[previousPlanInterval] ?? false
+  return isAnUpgrade ? UPGRADE : DOWNGRADE
 }
 
 async function getCIO(email: string) {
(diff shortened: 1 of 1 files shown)


src/pages/api/stripe/webhook.ts
```diff
@@ -11,6 +11,21 @@ function stripeToMixpanelDataConverter(stripeDate: number) {
...
+  previousPlanInterval: PlanInterval,
+  newPlanInterval: PlanInterval,
+) {
+  const previousPlanIntervalUpgrades = {
+    month: true,
+    year: false,
+    quarter: newPlanInterval === 'year',
+  }
+
+  return previousPlanIntervalUpgrades[previousPlanInterval] ?? false
```
zacjones93, line 26: Multiple if checks can be turned into a data structure for cleaner code. Thanks @theianjones

src/pages/api/stripe/webhook.ts
```diff
@@ -76,12 +91,13 @@ const stripeWebhookHandler = async (
...
-          const previousPlanAmount = previousSubscription.plan?.amount || 0
+          const previousPlanInterval = previousSubscription.plan?.interval || ''
+          const newPlanInterval =
+            stripeSubscription.items.data[0].plan?.interval || ''
 
-          const state =
-            previousPlanAmount < newPlanAmount ? 'upgrade' : 'downgrade'
+          const state = checkForUpgrade(previousPlanInterval, newPlanInterval)
+            ? 'upgrade'
+            : 'downgrade'
```
zacjones93, line 100: pulls interval data from stripe and calls `checkForUpgrade`

theianjones: you could return `upgrade` or `downgrade` instead of `true` or `false` from the `checkForUpgrade`?

theianjones: looks like `subscriptionInterval` is being created down below too on line 104

jbranchaud: > looks like `subscriptionInterval` is being created down below too on line 104

good catch, `newPlanInterval` (on 95) and `subscriptionInterval` (on 104) are basically the same

src/pages/api/stripe/webhook.ts
jbranchaud: This is a bit more verbose, but I think it gives more information and ratchets up the type safety. I'm on a bit of Zod kick and I think it is well suited for this scenario, where we have unknown/uncontrolled outside input coming in from a 3rd party webhook.

This is what the `checkForUpgrade` function could look like along with types.

```typescript
import {z} from 'zod';

const UPGRADE = "Upgrade";
const DOWNGRADE = "Downgrade";

const intervalOptions = z.union([
  z.literal("month"),
  z.literal("quarter"),
  z.literal("year")
]);

type Interval = z.infer<typeof intervalOptions>;
type IntervalChange = typeof UPGRADE | typeof DOWNGRADE;

function checkForUpgrade(
  previousPlanInterval: Interval,
  newPlanInterval: Interval
): IntervalChange {
  const isAnUpgrade = {
    month: true,
    year: false,
    quarter: newPlanInterval === "year"
  }[previousPlanInterval];

  return isAnUpgrade ? UPGRADE : DOWNGRADE;
}
```

And then the webhook code that checks the upgrade type would look something like:

```typescript
const previousPlanInterval = previousSubscription.plan?.interval
const newPlanInterval = stripeSubscription.items.data[0].plan?.interval

const intervalTuple = z.tuple([intervalOptions, intervalOptions]);
const result = intervalTuple.safeParse([
  previousPlanInterval,
  newPlanInterval
]);

if(!result.success) {
  const errorMessage = `
  Invalid interval types ahead of checkForUpgrade
  - previousPlanInterval: ${previousPlanInterval}
  - newPlanInterval: ${newPlanInterval}

  Detailed ZodError: ${result.error}
  `

  // OR, if we don't want to throw an error, we could send the message to something
  // like Honeybadger and then early return from the webhook handler.
  throw new Error(errorMessage)
}

// because of the `intervalTupe.safeParse(...)` above, we know we are passing in
// valid inputs to the checkForUpgrade
const state = checkForUpgrade(...result.data);
```

That's it for the code suggestion.

As a side note, I ran through the different possibilities to double check with you that this function produces the results you want.

```typescript
const options = intervalOptions.options;
const values = options.map((option) => option.value);

for (const previousInterval of values) {
  for (const newInterval of values) {
    const result = checkForUpgrade(
      previousInterval as Interval,
      newInterval as Interval
    );
    console.log(`- ${previousInterval} | ${newInterval} --> ${result}`);
  }
}
```

which spits out:

```
- month | month --> Upgrade 
- month | quarter --> Upgrade 
- month | year --> Upgrade 
- quarter | month --> Downgrade 
- quarter | quarter --> Downgrade 
- quarter | year --> Upgrade 
- year | month --> Downgrade 
- year | quarter --> Downgrade 
- year | year --> Downgrade 
```

jbranchaud: also, these are just suggestions, no pressure if you're wanting to get out the fix and not get caught up in this kind of refactor

zacjones93: Love this, I'm going to reach for Zod when I have heavier-duty typing to do

joelhooks (approved):

PR #1206, opened 2022-09-15
Add new tracking events for mixpanel

zacjones93:
This PR starts the implementation of the new tracking plan @nicollguarnizo and I worked on. 

The goal here is to abstract the low level `track` call so that throughout the code base you will start to use the pre-defined events that have been created. Currently we have +800 events logged in mixpanel which a large majority are variations of the same type of event but with a slightly different event name. (e.g. `clicked Ari Picker instructor profile` and `clicked Andy Van Slaars instructor profile`). The variations and specificity and be put into a variable we track on a more generic event.

To do this, we are exporting an object of events that contains the predetermined events that we want to track. These events are typed and variable names based on the event group (e.g. `events.engagementCompletedCourse(...)` and  `events.engagementCompletedLesson(...)`.

When building out a page you'll have a nice list to choose from and also when adding events you will only need to look in the `analytics` folder (or tracking plan spreadsheet) to get a grasp of what is implemented and where to add an event moving forward.

One possibly weird decision we made was in order to include the tracking events that we use server side in the same analytics folder, we instantiate mixpanel in the webhook and pass the object to the event function to be called in `server-side-events.ts`.

Currently we still export `track` from the index of `utils/analytics` but ideally when we clean up the old tracking events we can move these out and not expose `track` to the rest of the application.

tracking


src/pages/api/stripe/webhook.ts
```diff
@@ -165,13 +173,15 @@ const stripeWebhookHandler = async (
...
-        })
+        analytics.serverSideEvents.purchaseSubscriptionCanceled(
+          mixpanel,
+          cioCustomer.id,
+        )
+        analytics.serverSideEvents.purchaseSetSubscriptionStatus(
+          mixpanel,
+          cioCustomer.id,
+          'canceled',
+        )
```
zacjones93, line 184: mixpanel-node client gets passed in with inputs to tracking function

src/utils/analytics/events.ts
```diff
@@ -0,0 +1,169 @@
+import {track} from 'utils/analytics'
+
+//! ENGAGEMENT EVENT GROUP
```
zacjones93, line 3: We've named each event group in the comments

src/utils/analytics/events.ts
```diff
@@ -0,0 +1,169 @@
...
+// start course
+/* When a course get started by a user  */
+/* This will show courses that get started but the don't get finished  */
+
+export const engagementStartCourse = (slug: string) =>
+  track('start course', {
+    slug,
+  })
```
zacjones93, line 11: included the trigger and event description for each event, this would be nice to add to the typescript overlay but haven't looked into that.

src/utils/analytics/index.ts
```diff
@@ -1,11 +1,92 @@
...
+  engagementSearchedWithQuery,
+  engagementWatchedTalk,
+  engagementStartCourse,
+  engagementCourseProgress,
+  engagementCompletedCourse,
+  engagementCompletedLesson,
+  engagementClickedWatchedLessonAgain,
+  engagementListenPodcast,
+  engagementReadArticle,
+}
```
zacjones93, line 49: build list of analytic events to export and use throughout the site

src/utils/analytics/index.ts
```diff
@@ -75,45 +156,11 @@ export const track = (
...
-
-  return Promise.resolve(data)
-}
-
 const analytics = {
   track,
   identify,
+  events,
+  serverSideEvents,
 }
```
zacjones93, line 164: `track` and `identify` are still exported with `events` and `serverSideEvents`. Ideally we wouldn't export `track` here so that you have to use a predefined event or put the thought into adding an event

src/utils/analytics/server-side-events.ts
```diff
@@ -0,0 +1,102 @@
...
+  currentPeriodStart,
+  currentPeriodEnd,
+}: MixpanelData) =>
+  mixpanel.track('Subscription Created', {
+    distinct_id,
+    subscriptionType,
+    subscriptionInterval,
+    currentPeriodStart,
+    currentPeriodEnd,
+  })
```
zacjones93, line 29: invoke the mixpanel instance passed in with the rest of the variables

src/utils/analytics/index.ts
```diff
@@ -0,0 +1,58 @@
...
+  engagementCourseProgress,
+  engagementCompletedCourse,
+  engagementCompletedLesson,
+  engagementClickedWatchedLessonAgain,
+  engagementListenPodcast,
+  engagementReadArticle,
+}
+
+export const track = (event: string, paramsOrCallback?: any, callback?: any) =>
+  noop
```
zacjones93, line 49: the old track calls throughout the sight are now a `noop` so we should be starting fresh after this PR

nicollguarnizo (approved):

PR #1213, opened 2022-09-21
feat: handle coupons that have a minimum

jbranchaud:
- Logic was recently added to egghead-rails to allow for a coupon to have a minimum. If a particular plan's price is at or above the minimum, then it qualifies for that coupon discount. If not, then it should ignore that coupon.
- In the case of the Annual-only Fall Sale, we have a minimum set that includes the annual plan and excludes the quarterly and monthly plans.
- The `commerce-machine` in egghead-next assumes that coupons are available across all plans.
- Because of this assumption, we have the annual fall sale coupon code getting included in the Stripe Checkout Session for monthly plans which then fails on the Stripe-side because that is not a valid coupon for that plan.

Note: all of this logic is either for display purposes or communicating to Stripe what coupon to apply. Final say-so about whether a coupon is valid, whether it applies to a particular plan, and how much it affects the price happens server-side/Stripe-side.

More details in Roam: https://roamresearch.com/#/app/egghead/page/alSfRvrtT

coupon minimum

### DIFF
--- a/src/machines/commerce-machine.ts
+++ b/src/machines/commerce-machine.ts
@@ -1,7 +1,8 @@
 import {createMachine, assign} from 'xstate'
 import {loadPricingData} from 'lib/prices'
 import isEmpty from 'lodash/isEmpty'
-import {PricingData} from 'types'
+import {pricingPlanSchema, PricingData} from 'types'
+import {z} from 'zod'
 
 type CouponToApply = {
   couponCode: string
@@ -13,7 +14,7 @@ const findAvailablePPPCoupon = (pricingData: PricingData) => {
 }
 
 export interface CommerceMachineContext {
-  pricingData: PricingData
+  pricingData: PricingData | {}
   priceId: string | undefined
   quantity: number
   couponToApply: CouponToApply | undefined
@@ -37,7 +38,7 @@ export const commerceMachine = createMachine<
   {
     id: 'commerce',
     context: {
-      pricingData: {} as PricingData,
+      pricingData: {},
       priceId: undefined as string | undefined,
       quantity: 1,
       couponToApply: undefined as CouponToApply | undefined,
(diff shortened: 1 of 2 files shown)


src/machines/commerce-machine.ts
```diff
@@ -182,9 +183,27 @@ export const commerceMachine = createMachine<
             return {}
           }
 
+          const result = z
+            .object({plans: z.array(pricingPlanSchema)})
+            .safeParse(context.pricingData)
+          if (!result.success) return {pricingData: event.data}
```
jbranchaud, line 189: When the machine first boots, it is possible for `context.pricingData` to be `{}`. This bit of Zod `safeParse` gives us confidence that if we don't early return, then we have access to a `plans` array of `PricingPlan` objects.

src/machines/commerce-machine.ts
```diff
@@ -182,9 +183,27 @@ export const commerceMachine = createMachine<
...
+          // (`priceId`), then we should NOT include the `couponToApply`.
+          const currentPlan = result.data.plans.find(
+            (plan) => plan.stripe_price_id === context.priceId,
+          )
+          const noCouponToApply = isEmpty(currentPlan?.price_discounted)
+
+          const couponData: {} | {couponToApply: undefined | object} =
+            noCouponToApply
+              ? {couponToApply: undefined}
+              : extractAppliedDefaultCoupon(event.data)
```
jbranchaud, line 202: Without this ternary, we are overeagerly applying the default/site-wide coupon to plans for which it doesn't apply. If we detect that the price of the current plan hasn't been discounted, then we know the default coupon is _not_ applied. So we can ensure it isn't set (hence, `{couponToApply: undefined}`).

jbranchaud:
It looks like I have some CI failures to cleanup. I'd still be interested to hear any thoughts on the approach in the meantime. More rationale on this approach can be found in Roam: https://roamresearch.com/#/app/egghead/page/alSfRvrtT

jbranchaud:
Closing this, https://github.com/eggheadio/egghead-next/pull/1214 is a better candidate for fixing this issue.

PR #1214, opened 2022-09-22
feat: handle site-wide coupon with a minimum

jbranchaud:
- Logic was recently added to egghead-rails to allow for a coupon to
  have a minimum. If a particular plan's price is at or above the
  minimum, then it qualifies for that coupon discount. If not, then it
  should ignore that coupon.
- In the case of the Annual-only Fall Sale, we have a minimum set that
  includes the annual plan and excludes the quarterly and monthly plans.
- The `commerce-machine` in egghead-next assumes that coupons are
  available across all plans.
- Because of this assumption, we have the annual fall sale coupon code
  getting included in the Stripe Checkout Session for monthly plans
  which then fails on the Stripe-side because that is not a valid coupon
  for that plan.

Roam: https://roamresearch.com/#/app/egghead/page/alSfRvrtT

coupons

### DIFF
--- a/src/machines/commerce-machine.ts
+++ b/src/machines/commerce-machine.ts
@@ -12,6 +12,36 @@ const findAvailablePPPCoupon = (pricingData: PricingData) => {
   return pricingData?.available_coupons?.ppp
 }
 
+const extractAppliedDefaultCoupon = (
+  pricingData: PricingData,
+): {couponToApply: CouponToApply} | {} => {
+  // check if there is a site-wide/'default' coupon available
+  const availableDefaultCoupon = pricingData?.available_coupons?.default
+
+  // check if there is a site-wide/'default' coupon being applied
+  const appliedCoupon = pricingData?.applied_coupon
+
+  // no applied default coupon found
+  if (isEmpty(availableDefaultCoupon) || isEmpty(appliedCoupon)) {
+    return {}
+  }
+
+  // if a site-wide/'default' coupon is being applied, set that as the
+  // `couponToApply` in the machine context. The CouponToApply is
+  // needed when transitioning to the Stripe Checkout Session.
+  if (availableDefaultCoupon?.coupon_code === appliedCoupon?.coupon_code) {
+    return {
+      couponToApply: {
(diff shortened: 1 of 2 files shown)


src/machines/commerce-machine.ts
```diff
@@ -12,6 +12,36 @@ const findAvailablePPPCoupon = (pricingData: PricingData) => {
   return pricingData?.available_coupons?.ppp
 }
 
+const extractAppliedDefaultCoupon = (
```
jbranchaud, line 15: Nothing new about this function, just moved it up here so that it can be called from multiple actions.

src/machines/commerce-machine.ts
```diff
@@ -81,9 +111,7 @@ export const commerceMachine = createMachine<
         initial: 'checkingCouponStatus',
         on: {
           SWITCH_PRICE: {
-            actions: assign({
-              priceId: (_context, event) => event.priceId,
-            }),
+            actions: ['switchPriceIdAndAssociatedData'],
```
jbranchaud, line 114: Adding a _switch price id_ action with a bit more logic so that we can check and adjust the applied coupon whenever the `priceId` changes.

src/machines/commerce-machine.ts
```diff
@@ -145,43 +173,26 @@ export const commerceMachine = createMachine<
...
+        )
+        const noCouponToApply = !currentPlan?.price_discounted
+        const couponData = noCouponToApply
+          ? {couponToApply: undefined}
+          : extractAppliedDefaultCoupon(context.pricingData)
+
+        return {
+          priceId: event.priceId,
+          ...couponData,
+        }
```
jbranchaud, line 190: Here is the bulk of the fix: we check if a couple should be applied (based on the server-computed `price_discounted` value) and update the machine context accordingly. For site-wide coupons that have a minimum, we'll detect whether it applies to the specific plan and either apply it or wipe it out with `undefined`.

garrettdimon (approved):
Feels like the right medium-term solution to me.

PR #1221, opened 2022-10-04
Improve search page heading structure

zacjones93:
@laurosilvacom @nicollguarnizo and I worked through ahref reports and found that we had thousands of pages with missing `h1`s. 

Digging into the issue, it mainly stemmed from our generated search pages not having an `h1` set. I noticed our heading structure was all over the place for the different landing pages types that we have. There are quite a few bespoke landing pages that were using components that assumed a heading level that didn't fit the page. 

I found [a good solution](https://dev.to/s_aitchison/psa-stop-hard-coding-heading-levels-in-your-react-components-2ekp) on passing in heading levels to components that I implemented.

Some examples of issues:

## Before
No topic selected:
before - search page structure with no topic selected

Topic page mis-structured:
before - search page structure of topic

## After

after - search page structure with no topic selected

after - search page structure of topic

cat


zacjones93:
Need to figure out a correct typing for the dynamic heading:
`Type error: Type '{ children: string; }' has no properties in common with type 'IntrinsicAttributes'.`

src/components/card/safe-title.tsx
```diff
@@ -0,0 +1,19 @@
...
+  className = '',
+}) => {
+  const Heading = setSafeHeadingLevel(headingLevel) as React.FC<{
+    children: React.ReactNode
+    className?: string
+  }>
+  return <Heading className={className}>{children}</Heading>
+}
+
+export default SafeTitle
```
zacjones93, line 19: The best solution I found for dynamically setting a heading. TypeScript does not like the conversion from string to ReactNode.

zacjones93: I also restricted heading options to 4 levels because that seems deep enough in 99% of cases. Want to add some friction to more heading levels

joelhooks (approved):

src/components/card/safe-title.tsx
theianjones, line 19: This is how I would build the component:

```jsx
import cx from 'classnames'

type Heading = React.ComponentPropsWithoutRef<any> & {
  as?: 'h1' | 'h2' | 'h3' | 'h4' | 'p'
}

const SafeTitle = ({
  children,
  as,
  className,
}: Heading) => {
  const Component = as ?? 'p'
  return <Component className={cx(className)}>{children}</Component>
}
```

theianjones: `classnames` will apply the class if its passed in, if not it will return an empty string I believe

theianjones: Yeah that works too. I just like classnames because if you ever need to add default styles and combine the classes coming in from the props, it will do that for you. But it's definitely not needed here

PR #1230, opened 2022-10-12
Turn subscription info back

DrShpongle:
Image 2022-10-12 at 15 16 33

### DIFF
--- a/src/components/pages/user/components/subscription-details.tsx
+++ b/src/components/pages/user/components/subscription-details.tsx
@@ -17,7 +17,7 @@ const SubscriptionDetails: React.FunctionComponent<SubscriptionDetailsProps> =
       stripeCustomerId,
     })
 
-    const subscriptionName = subscriptionData && subscriptionData.product?.name
+    const subscriptionName = subscriptionData?.product?.name
     const subscriptionUnitAmount = get(
       subscriptionData,
       'latestInvoice.amount_due',
--- a/src/components/pages/user/components/subscription-details.tsx
+++ b/src/components/pages/user/components/subscription-details.tsx
@@ -1,0 +1,133 @@
+import React from 'react'
+import Link from 'next/link'
+import {track} from 'utils/analytics'
+import {useViewer} from 'context/viewer-context'
+import {get} from 'lodash'
+import useSubscriptionDetails, {recur} from 'hooks/use-subscription-data'
+
+type SubscriptionDetailsProps = {
+  stripeCustomerId: string
+  slug: string
+}
+
+const SubscriptionDetails: React.FunctionComponent<SubscriptionDetailsProps> =
+  ({stripeCustomerId, slug}) => {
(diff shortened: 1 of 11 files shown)


src/components/pages/user/tabs-content/account-info-tab-content.tsx
```diff
@@ -60,3 +99,29 @@ const GithubConnectButton: React.FunctionComponent<{
...
+
+// React.useEffect(() => {
+//   const loadAccountForSlug = async (slug: string) => {
+//     if (slug) {
+//       const account: any = await loadAccount(slug, authToken)
+//       setAccount(account)
+//     }
+//   }
+//   loadAccountForSlug(slug)
+// }, [slug, authToken])
```
jbranchaud, line 127: It looks like the parts of this that were needed are used above. This can be deleted.

```suggestion
```

src/components/pages/user/components/subscription-details.tsx
```diff
@@ -0,0 +1,133 @@
...
+}
+
+const SubscriptionDetails: React.FunctionComponent<SubscriptionDetailsProps> =
+  ({stripeCustomerId, slug}) => {
+    const {viewer} = useViewer()
+    const {subscriptionData, loading} = useSubscriptionDetails({
+      stripeCustomerId,
+    })
+
+    const subscriptionName = subscriptionData && subscriptionData.product?.name
```
jbranchaud, line 20:
```suggestion
    const subscriptionName = subscriptionData?.product?.name
```

src/components/pages/user/components/subscription-details.tsx
```diff
@@ -0,0 +1,133 @@
...
+        ) : (
+          <>
+            <h3 className="mb-2 text-lg font-medium">
+              No paid subscription found.
+            </h3>
+            {(viewer.is_pro || viewer.is_instructor) && (
+              <p>
+                You still have access to a Pro Membership. If you feel this is
+                in error please email{' '}
+                <a
```
jbranchaud, line 90: I know there are some weird scenarios that lead to a person having access. But I wonder if one of the main scenarios in which this message displays is because a person has access through team membership. In which case, the messaging might be confusing.

jbranchaud: On second thought, I think this component is only going to render if the user has (owned) accounts tied to them. Which is to say, that only account owners are going to see this. My previous comment, in that case, is not a concern.

jbranchaud (changes requested):
This looks great. The main thing I'd request is some restructuring of the SubscriptionDetails rendering logic. I get the impression that some unintended states could occur where inconsistent messages display at the same time. I suggested some changes that may flush these out.

DrShpongle:
I did rough mockup of each states content, does it look correct?
https://excalidraw.com/#room=ab824380d1e835dc9951,cDM_1X3AziG71pYgV7FZdg

PR #1233, opened 2022-10-14
deleting login tracking events

nicollguarnizo:
Some users are presenting login problems with email and github, this problem seems to come from the tracking events implemented coming from this [PR](https://github.com/eggheadio/egghead-next/pull/1222)

image

Deleting the events solve the console problems. Not sure what's the root of the problem, but the theory is that since we `noop` the old events, adding the `onClick` event in `ExternalTrackedLink` messed up the redirect using github. 

Hopefully, this PR allows users to login, and we will further investigate how to track the events properly. 

oops


garrettdimon (approved):
🙌🏻

joelhooks (approved):

PR #1234, opened 2022-10-24
seo: update canonical url and description for SEO

zacjones93:
What is DOM? and what is the dom? are high volume searches (33k/m) which we have very little clicks for. This updates the page with a canonical url and better description which will hopefully increase SEO and click through.

Our page description is lacking for this: 
image

image

canon

### DIFF
--- a/src/layouts/fancy-guide.tsx
+++ b/src/layouts/fancy-guide.tsx
@@ -1,6 +1,7 @@
 import React, {FunctionComponent} from 'react'
 import {NextSeo} from 'next-seo'
 import Contributors from '../components/contributors'
+import {useRouter} from 'next/router'
 
 type LayoutProps = {
   meta?: {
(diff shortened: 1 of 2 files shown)


src/layouts/fancy-guide.tsx
```diff
@@ -17,14 +18,20 @@ const FancyGuideLayout: FunctionComponent<LayoutProps> = ({
...
     title,
     description,
     titleAppendSiteName = false,
-    url,
+    url: urlFromProps,
     ogImage,
     contributors,
   } = meta
+
+  const canonicalUrl = urlFromProps ? urlFromProps : url
```
zacjones93, line 33: updates our <FancyGuide> component used on a few MDX pages to set the canonicalUrl from the router if none was passed. Currently I couldn't find any of these pages that use the component where we pass a url.

nicollguarnizo (approved):

Issue #1240, opened 2022-11-28
Not all completed courses are shown in "Completed courses" on user page


DrShpongle:
Some users complaint that not all completed courses are shown in "Completed courses" on user page.
https://eggheadio.slack.com/archives/G01JXF8N3LK/p1669483237431139

DrShpongle:
fixed with https://github.com/skillrecordings/egghead-next/pull/1284

PR #1241, opened 2022-11-28
"Completed" label on course page

DrShpongle:
This adds the "Completed label" on the Course page in case user have already completed it.

Screen Recording 2022-11-28 at 14 26 15

There are few bugs to fix:
1) some users complaint that not all completed courses are marked as "completed", I have created related issue: https://github.com/eggheadio/egghead-next/issues/1240
2) Play button should display "Start watching" or "Continue watching" depends on progress, but it seems it doesn't work ("Start watching" is shown always) - will tackle it in separate PR.

shia-labeouf-even-stevens (1)

P.S. Also fixed "unique key attribute" warning on the Home page, just it was too minor for separate PR.


zacjones93 (approved):
LGTM

PR #1246, opened 2022-12-03
Prisma Starter

garrettdimon:
Adds the prisma schema (and a fully-generated version for reference).


prisma/schema.prisma
```diff
@@ -0,0 +1,84 @@
...
+generator client {
+  provider = "prisma-client-js"
+}
+
+datasource db {
+  provider = "postgresql"
+  url      = env("DATABASE_URL")
+}
+
+model OathAccessToken {
```
jbranchaud, line 13: should this be `OauthAccessToken`? Notice the missing `u`?

jbranchaud:
This looks good to me. I'm not as familiar with the details on setting up tRPC, but I looked over those various files and it makes sense.

I liked that you only pulled in the bits of the schema that you need at the moment.

The tRPC layer seems like a natural place to (eventually) layer in CASL for mirroring (or reimplementing) the Ability checks that are going to be needed for other features that we want to transition from `egghead-rails` to `egghead-next`.

src/pages/user/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+  )
+}
+
+const tabs = [
+  {label: 'Activity', component: <ActivityTabContent />},
+  {label: 'Bookmarks', component: <BookmarksTabContent />},
+  {label: 'Account Info', component: <AccountInfoTabContent />},
+]
+
+export default User
```
DrShpongle, line 98: I think this page should be removed

src/components/pages/user/tabs-content/account-info-tab-content.tsx
```diff
@@ -1,17 +1,44 @@
...
 import {ItemWrapper} from 'components/pages/user/components/widget-wrapper'
+import SubscriptionDetails from 'components/pages/user/components/subscription-details'
+
 import {useViewer} from 'context/viewer-context'
 import {
   AvatarForm,
   RequestEmailChangeForm,
   // RequestNameChangeForm,
 } from '../components'
+import {trpc} from '../../../../utils/trpc'
```
DrShpongle, line 14: this could be reduced to just `import {trpc} from 'utils/trpc'` I think

src/components/pages/user/tabs-content/account-info-tab-content.tsx
```diff
@@ -1,17 +1,44 @@
...
-  const {email: currentEmail, providers} = viewer || {}
+  const {viewer, authToken, loading} = useViewer()
+  const {email: currentEmail, accounts, providers} = viewer || {}
   const isConnectedToGithub = providers?.includes('github')
+  const {slug} = getAccountWithSubscription(accounts)
+
+  const {data: account, status: accountLoadingStatus} =
+    trpc.user.current.useQuery()
+
+  console.log({account, accountLoadingStatus})
```
DrShpongle, line 41: this could be removed

PR #1254, opened 2022-12-10
Fix completed courses on User page

DrShpongle:
Completed courses amount was equal zero.

Image 2022-12-09 at 16 34 04

Episode 2 GIF by The Simpsons

### DIFF
--- a/src/components/pages/user/components/completed-courses.tsx
+++ b/src/components/pages/user/components/completed-courses.tsx
@@ -34,7 +34,7 @@ const CompletedCourses: React.FC<{
         </div>
       ) : completedCourseStatus === 'error' ? (
         <span>There was an error fetching stats</span>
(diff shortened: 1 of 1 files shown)


src/components/pages/user/components/completed-courses.tsx
```diff
@@ -33,7 +33,7 @@ const CompletedCourses: React.FC<{
         </div>
       ) : completedCourseStatus === 'error' ? (
         <span>There was an error fetching stats</span>
-      ) : !null || completeCourseData?.length === 0 ? (
```
DrShpongle, line 36: I think this condition is not correct - `!null` is also `true` so we will always get "You haven't finished any courses yet". Also I think we don't need optional chaining here because there is default `[]` value for `completeCourseData ` set up above.

zacjones93:
fixed with #1253

PR #1258, opened 2022-12-14
Fixes EGG-50: fetch resource state from rails and add retired banner

zacjones93:
Update course and lesson page to actually show retired banner if resource is in fact retired.

course banner desktop
course banner mobile

lesson banner desktop
lesson banner mobile

retired

### DIFF
--- a/src/pages/lessons/[slug].tsx
+++ b/src/pages/lessons/[slug].tsx
@@ -568,7 +568,7 @@ const Lesson: React.FC<LessonProps> = ({
             />
           </div>
           {withSidePanel && (
-            <div className="col-span-3 flex flex-col dark:bg-gray-800 bg-gray-50">
+            <div className="flex flex-col col-span-3 dark:bg-gray-800 bg-gray-50">
               <PlayerSidebar
                 relatedResources={specialLessons[lesson.slug]}
                 videoResource={lesson}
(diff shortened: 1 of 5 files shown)


joelhooks:
@zacjones93 can we also remove the download button since it doesn't work (or fix it)

DrShpongle (approved):

src/pages/lessons/[slug].tsx
```diff
@@ -581,6 +582,12 @@ const Lesson: React.FC<LessonProps> = ({
       <div className="container max-w-screen-lg py-8 md:py-12 lg:py-16">
         <div className="grid grid-cols-1 gap-8 divide-y lg:grid-cols-1 lg:gap-12 md:divide-transparent divide-gray-50">
           <div className="row-start-1 space-y-6 md:col-span-8 md:row-start-1 md:space-y-8 lg:space-y-10">
+            {lesson.state === 'RETIRED' && (
```
Comment (jbranchaud) [inline src/pages/lessons/[slug].tsx:585]: Is it supposed to be all-caps here and lowercase up above?

CleanShot 2022-12-15 at 10 23 59@2x

joelhooks (approved):

src/pages/lessons/[slug].tsx
Comment (zacjones93) [reply src/pages/lessons/[slug].tsx:585]: lesson state was coming in all-caps for whatever reason

Use Chrome Developer Tools to Emulate Different Mobile Devices egghead io 2022-12-15 at 4 20 53 PM

PR #1265, opened 2023-01-03
adds ability to send a coupon code via url for pricing page

joelhooks:
the price is wrong


src/machines/commerce-machine.ts
```diff
@@ -166,9 +166,12 @@ export const commerceMachine = createMachine<
   {
     services: {
       fetchPricingData: async (context): Promise<PricingData | undefined> => {
+        let params = new URLSearchParams(window.location.search)
```
joelhooks, line 169: we weren't able to send a code via the url previously

PR #1269, opened 2023-01-09
Fixes EGG-101: provide a direct link to invoice after purchase

DrShpongle:
Actually, problem is not solved - after purchase transactions are not immediately available, so call `axios.get(`${process.env.NEXT_PUBLIC_AUTH_DOMAIN}/api/v1/transactions`)` returns the data after about ~1 min (our db  processing time?). It means that on the MembershipConfirmation screen that user gets in browser just after purchase completed (and it redirects from stripe portal) `transactions` data is not loaded.
Is there a way to get `stripe_transaction_id` immediately? Can it be passed along with with redirect from stripe portal somehow?

No funny GIF for now 🥸

### DIFF
--- a/src/components/invoices/index.tsx
+++ b/src/components/invoices/index.tsx
@@ -2,6 +2,8 @@ import * as React from 'react'
 import axios from 'utils/configured-axios'
 import {isEmpty} from 'lodash'
 import {format, parseISO} from 'date-fns'
+import Spinner from 'components/spinner'
+import Link from 'next/link'
 
 type HeadingProps = React.ComponentPropsWithoutRef<any> & {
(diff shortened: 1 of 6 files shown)


joelhooks:
we do this on partner products by having the charge page (so /invoices/transaction_id in this case) open, doesn't even require a login. 

that doesn't work here because we also bolted the "here's your team link" on the invoice page (it looks awful) but it makes it SUPER convenient to provide quick links to invoices and allow for sharing of invoice link to their internal stakeholders

@zacjones93 @garrettdimon

joelhooks (approved):

DrShpongle:
fixed with https://github.com/skillrecordings/egghead-next/pull/1290

PR #1270, opened 2023-01-11
Fixes EGG-86: continue enhancing the "what I've viewed" functionality

DrShpongle:
this adds ✅ label to completed courses on the Search page:

- [x] put label on curated part of the search page
- [x] put label on completed course cards in search results

<details>
  <summary>Curated part of the Search Page:</summary>
  
</details>
<details>
  <summary>Search results:</summary>
  
</details>

Technically, it works.
What I'm not happy with is that I use the same `useQuery` call twice on the same page (`src/components/search/index.tsx`) in two child components: `<NewCuratedTopicPage />` and `<Hits />`. Moreover, I was forced to wrap `<Hits />` with additional `<QueryClientProvider>`, otherwise I get the error described below. The `useQuery` call in `<NewCuratedTopicPage />` works smooth though. I've tried to move `useQuery` call on parent level (`src/components/search/index.tsx`, `<Search />`) and just pass the call result as props to child components, but I've got the error described here: https://eggheadio.slack.com/archives/G01CAT0P9TL/p1673398982014119

Would be happy if someone could help me to understand what I'm missing here 🥸

### DIFF
--- a/src/components/card/search-hit-resource-card.tsx
+++ b/src/components/card/search-hit-resource-card.tsx
@@ -16,6 +16,7 @@ import {CardResource} from 'types'
 import {Textfit} from 'react-textfit'
 import ReactMarkdown from 'react-markdown'
 import useFitText from 'use-fit-text'
+import CheckIcon from 'components/icons/check'
 
 const SearchHitResourceCard: React.FC<{
   resource: CardResource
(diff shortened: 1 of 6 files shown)


joelhooks (approved):
`useQuery` has efficient cache so calling it multiple times is not an issue and is encouraged. another approach is the wrap useQuery in custom hook like you've done. it looks good. could extract to shared hook

PR #1273, opened 2023-01-13
fix: correctly display gift membership message

zacjones93:
This fixes [EGG-134](https://linear.app/skillrecordings/issue/EGG-134/re-my-early-subscription-is-gone) where gift memberships are showing 'No Subscription Found' because we are only checking if `stripe_subscription_id` exists. Gift memberships have a null `stripe_subscrition_id`

Currently very simple messaging, any suggestions there? 
image

gift

### DIFF
--- a/src/pages/user/subscription.tsx
+++ b/src/pages/user/subscription.tsx
@@ -10,6 +10,7 @@ import UserLayout from './components/user-layout'
 import PricingWidget from 'components/pricing/pricing-widget'
 import Invoices from 'components/invoices'
 import Spinner from 'components/spinner'
+import {format} from 'date-fns'
 
 type ViewerAccount = {
   stripe_customer_id: string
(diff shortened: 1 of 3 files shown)


src/pages/user/subscription.tsx
```diff
@@ -52,6 +56,16 @@ const Account = () => {
         <div className="relative flex justify-center">
           <Spinner className="w-6 h-6 text-gray-600" />
         </div>
+      ) : isGiftMembership ? (
+        <div className="h-40 sm:h-60 flex flex-col justify-center">
+          <h2 className="pb-3 md:pb-4 text-lg font-medium md:font-normal md:text-xl leading-none w-fit mx-auto">
+            You have claimed a gift membership.
```
joelhooks, line 62:
```suggestion
            You have an egghead membership.
```

joelhooks (approved):
a follow up here might be to allow them to extend their membership if they have an active gift, but that's more complicated

PR #1278, opened 2023-01-20
add message to email support for help if account is cancelled

zacjones93:
cry for help

show cancel message


src/components/pages/user/components/subscription-details.tsx
```diff
@@ -61,8 +61,18 @@ const SubscriptionDetails: React.FunctionComponent<SubscriptionDetailsProps> =
                   </strong>
                   ). You can renew at any time.
                 </p>
-                <p></p>
-                <p></p>
+                <p>
+                  You can also get help by emailing{' '}
+                  <strong>
+                    <a
+                      href="mailto:support@egghead.io"
```
joelhooks, line 68: mailto links can have a subsject AND body added which should almost always be the case for this type of usage https://mailtolinkgenerator.com/

DrShpongle: something like:
```
<a
  href={`mailto:support@egghead.io?subject=${encodeURIComponent(
    `Subscription ${subscriptionData.subscription.id} is cancelled`,
  )}&body=${encodeURIComponent(
    `Subscription ${subscriptionData.subscription.id} is cancelled`,
  )}`}
  style={{textDecoration: 'underline'}}
>
  support@egghead.io
</a>
```

src/components/users/subscription-details.tsx
```diff
@@ -97,10 +97,7 @@ const SubscriptionDetails: React.FunctionComponent<SubscriptionDetailsProps> =
                         <p>
                           You still have access to a Pro Membership. If you feel
                           this is in error please email{' '}
-                          <a
-                            className="text-blue-600 underline hover:text-blue-700"
-                            href="mailto:support@egghead.io"
-                          >
+                          <a href="mailto:support@egghead.io?subject=support%20needed%20for%20egghead%20membership">
```
DrShpongle, line 100: it seems that this component is not used anywhere

src/pages/user/membership.tsx
```diff
@@ -68,8 +68,12 @@ const Account = () => {
           </h2>
           <p className="w-fit mx-auto mb-12">
             If this is incorrect, please reach out to{' '}
-            <strong>support@egghead.io</strong> or your team owner{' '}
-            {accountOwner.email}.
+            <strong>
+              <a href="mailto:support@egghead.io?subject=Support needed for egghead team membership">
```
DrShpongle, line 72: I think we want to use `encodeURIComponent` for the strings here

src/components/pages/user/components/subscription-details.tsx
```diff
@@ -61,8 +61,18 @@ const SubscriptionDetails: React.FunctionComponent<SubscriptionDetailsProps> =
...
+                      style={{textDecoration: 'underline'}}
```
DrShpongle, line 69: can be just `underline` class

src/pages/user/membership.tsx
DrShpongle: probably not 🤔

src/components/users/subscription-details.tsx
zacjones93: ah, wasn't sure

DrShpongle (approved):
👍

PR #1280, opened 2023-01-24
Fixes EGG-155: subscription expiration date is in the past

DrShpongle:
Should work as expected now and show Pricing widget for the `gift` subscription with `status !== "active"`.

giphy

### DIFF
--- a/src/hooks/use-account.ts
+++ b/src/hooks/use-account.ts
@@ -31,7 +31,9 @@ export const useAccount = () => {
       return account.owner?.id === viewer.id
     })
 
-  const isGiftMembership = account?.subscriptions?.[0]?.type === 'gift'
+  const isGiftMembership =
+    account?.subscriptions?.[0]?.type === 'gift' &&
+    account?.subscriptions?.[0]?.status === 'active'
   const giftExpiration = account?.subscriptions?.[0]?.current_period_end
 
   const isTeamMember = isActiveAccountMember && !isAccountOwner
--- a/src/components/pages/user/components/user-layout.tsx
+++ b/src/components/pages/user/components/user-layout.tsx
@@ -90,8 +90,10 @@ export default function UserLayout({children}: any) {
                   })}
                 </nav>
               </aside>
-              <section className="divide-y divide-gray-200 lg:col-span-9">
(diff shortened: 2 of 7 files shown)


src/hooks/use-account.ts
```diff
@@ -31,7 +31,9 @@ export const useAccount = () => {
       return account.owner?.id === viewer.id
     })
 
-  const isGiftMembership = account?.subscriptions?.[0]?.type === 'gift'
+  const isGiftMembership =
+    account?.subscriptions?.[0]?.type === 'gift' &&
+    account?.subscriptions?.[0]?.status === 'active'
```
DrShpongle, line 36: made `isGiftMembership` dependent on `status` here

joelhooks (approved):

src/hooks/use-account.ts
garrettdimon: It should check for `past_due` and `trialing` as valid active options too. (Although, we shouldn't have any accounts in the trialing state.)

Alternatively, we could just have `egghead-rails` return `enabled` or `disabled` so we're not duplicating the logic? See statuses from `egghead-rails`:
https://github.com/skillrecordings/egghead-rails/blob/main/app/models/account_subscription.rb#L2

garrettdimon:
https://github.com/skillrecordings/egghead-next/pull/1280#discussion_r1084776658

Technically, it's an array, but `egghead-rails` is ~only~ usually returning the One Subscription to Rule them All. 

If GraphQL:
https://github.com/skillrecordings/egghead-rails/blob/main/app/graphql/types/account_type.rb#LL35C8-L35C8

If `api/v1/users`:
- https://github.com/skillrecordings/egghead-rails/blob/main/app/views/api/v1/users/_user.json.jbuilder#L96
- https://github.com/skillrecordings/egghead-rails/blob/main/app/views/api/v1/users/_user.json.jbuilder#L105

Although, the `minimal` user profile view does include a full array, but I would think we probably don't want to do that?
https://github.com/skillrecordings/egghead-rails/blob/main/app/views/api/v1/users/_minimal_user.json.jbuilder#L52

I would think if we're standardizing on GraphQL/tPRC, then these user controller endpoints should probably get phased out of use?

PR #1282, opened 2023-01-24
Probably fixes subscription expiration date is in the past

DrShpongle:
This adds isGiftExpired flag for the isGiftMembership condition.


joelhooks (approved):
in this circumstance they are also continuing to receive free access which would be interesting to remove as well, but know that's out of scope for the UI problem.

PR #1284, opened 2023-01-25
fixes egg-159: add course type to collection query

zacjones93:
We weren't pulling in all the data that we should with `courseCompletions` query. 

This adds the course type which affects any course that was migrated over from course to playlist. This makes me wonder if there are other instances where this migration has affected courses. For instance, the Web Security Essentials course last review was 2 years ago even though in slack we have a review logged at Nov 2nd, 2022.

activity shows correctly

something is missing


DrShpongle:
Cool! 👍

joelhooks (approved):

PR #1285, opened 2023-01-25
Fixes EGG-161: current_period_end date formatting is incorrect

DrShpongle:
before:
Image 2023-01-25 at 12 39 16

after:
Image 2023-01-25 at 12 39 00

Tv Land Dating GIF by YoungerTV


zacjones93 (approved):

garrettdimon:
Separately, but maybe tangentially helpful related to this, I found about 15,000 `AccountSubcription`'s for which `current_period_end` was `nil`. I scripted a fix that looked at the billing interval to generate the `current_period_end` from the `current_period_start`, so hopefully there won't be any nil values there any longer.

PR #1288, opened 2023-01-27
demonstrate the use of trpc to load subscription data for user page

joelhooks:
shell game

PR #1289, opened 2023-01-27
Refactor SubscriptionDetails view logic into switch statement

zacjones93:
The `SubscriptionDetails` component was getting a bit hard to understand with the 5 different states that users could be in and conditionals scattered around the JSX. 

This refractors the this component to follow the same patter as the `/user/membership` with a series of switch cases to handle the different states. 

The states that are handled: 
- single paid membership (default case)
- single membership pending cancelation
- team owner
- team owner pending cancelation
- disabled account (previous membership that is now delinquent - handled on `/user/membership` page)

kitten

## single membership subscribed  
image
## single membership pending cancelation
image
## team membership owner
image

## team membership owner pending cancellation
image

## disabled account
image


zacjones93:
Merged with changes from #1288

joelhooks (approved):

PR #1290, opened 2023-01-27
demonstrate the use of trpc to load transactions for invoice link

joelhooks:
https://www.loom.com/share/022d1f4b1f274132bfcefa437312f4a8

transaction

### DIFF
--- a/src/components/pages/confirm/membership/index.tsx
+++ b/src/components/pages/confirm/membership/index.tsx
@@ -20,6 +20,51 @@ type ConfirmMembershipProps = {
   session_id: string
 }
 
+const ExistingMemberConfirmation: React.FC<{session_id: string}> = ({
+  session_id,
+}) => {
+  const {data} = trpc.stripe.checkoutSessionById.useQuery({
+    checkoutSessionId: session_id as string,
+  })
+
+  return data ? (
+    <>
+      <Header
+        heading={<>Thank you so much for joining egghead!</>}
+        primaryMessage={
+          <>
+            <p className="text-lg text-center">
+              We've charged your credit card{' '}
+              <strong>
+                ${(data?.session?.amount_total || 0) / 100} for your egghead
+                membership
+              </strong>{' '}
+              and sent a receipt to <strong>{data.customer.email}</strong>.
+            </p>
+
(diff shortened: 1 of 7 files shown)


DrShpongle:
Actually the same problem I described here: https://github.com/skillrecordings/egghead-next/pull/1269 :

"...after purchase transactions are not immediately available, so call axios.get(${process.env.NEXT_PUBLIC_AUTH_DOMAIN}/api/v1/transactions) returns the data after about ~1 min (our db processing time?). It means that on the MembershipConfirmation screen that user gets in browser just after purchase completed (and it redirects from stripe portal) transactions data is not loaded.
Is there a way to get stripe_transaction_id immediately?"

<details>
  <summary>Immediately after redirection from stripe portal</summary>
  
</details>
<details>
  <summary>After ~1 min</summary>
  
</details>

DrShpongle:
from the related discussion with Garrett:
Image 2023-01-27 at 15 24 15

src/components/pages/confirm/membership/index.tsx
```diff
@@ -7,16 +7,50 @@ import Spinner from 'components/spinner'
...
+const LinkToLatestInvoice = () => {
+  const {account} = useAccount()
+  const {data: subscriptionData, status: subscriptionLoadingStatus} =
+    trpc.subscriptionDetails.forStripeCustomerId.useQuery({
+      stripeCustomerId: account?.stripe_customer_id,
+    })
+  const {data: charge, status: chargeLoadingStatus} =
+    trpc.stripe.balanceTransactionById.useQuery({
+      chargeId: subscriptionData?.latestInvoice?.charge as string,
+    })
```
DrShpongle, line 32: I did few calls here: `account` (to get `stripe_customer_id `) --> `subscriptionData` (to get `latestInvoice. charge ` id) --> `charge` (to get actually `balance_transaction`).
There are couple of errors during fetching process (because I don't get data for the next call immediately):
Image 2023-01-27 at 17 58 04

Not sure how to handle it. Wrap in promises?

DrShpongle: technically it works:
charge

PR #1291, opened 2023-02-01
honor coupon codes

joelhooks:
if a coupon was passed in as a query param it was being cast aside and not applied to the checkout session only the price check so a displayed price didn't match the stripe checkout price

coupon

### DIFF
--- a/src/machines/commerce-machine.ts
+++ b/src/machines/commerce-machine.ts
@@ -5,7 +5,7 @@ import {PricingData} from 'types'
 
 type CouponToApply = {
   couponCode: string
-  couponType: 'ppp' | 'default'
+  couponType: 'ppp' | 'special' | 'default'
 }
 
 const findAvailablePPPCoupon = (pricingData: PricingData) => {
@@ -22,7 +22,7 @@ const extractAppliedDefaultCoupon = (
   const appliedCoupon = pricingData?.applied_coupon
 
   // no applied default coupon found
-  if (isEmpty(availableDefaultCoupon) || isEmpty(appliedCoupon)) {
+  if (isEmpty(availableDefaultCoupon) && isEmpty(appliedCoupon)) {
     return {}
   }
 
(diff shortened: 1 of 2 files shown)


src/machines/commerce-machine.ts
```diff
@@ -22,7 +22,7 @@ const extractAppliedDefaultCoupon = (
   const appliedCoupon = pricingData?.applied_coupon
 
   // no applied default coupon found
-  if (isEmpty(availableDefaultCoupon) || isEmpty(appliedCoupon)) {
+  if (isEmpty(availableDefaultCoupon) && isEmpty(appliedCoupon)) {
```
joelhooks, line 25: we still want to check below iff we have an applied coupon that isn't region restricted.

src/machines/commerce-machine.ts
```diff
@@ -38,6 +38,17 @@ const extractAppliedDefaultCoupon = (
...
+  // if we have an appliedCoupon that isn't region restricted and not default
+  // we want to honor it
+  if (appliedCoupon && !appliedCoupon.coupon_region_restricted) {
+    return {
+      couponToApply: {
+        couponCode: appliedCoupon.coupon_code,
+        couponType: 'special',
+      },
+    }
+  }
```
joelhooks, line 50: non-default non-ppp applied coupon is still valid and is being PRESENTED as the price but not carrying through to Stripe

jbranchaud (approved):
Good fix!

Because I was trying to refresh myself on how this works and make sure that the coupon was ultimately getting verified and set server-side, I dug through the code a little.

Here is one of a couple places where the applied coupon gets passed into the stripe redirect:

CleanShot 2023-02-02 at 10 15 35@2x

Here is the stripe redirect making use of that coupon code and passing it to the server:

CleanShot 2023-02-02 at 10 15 56@2x

And then the server does verify the coupon code before handing things off the Stripe.

PR #1292, opened 2023-02-02
let instructors know they are instructors

zacjones93:
We currently ask instructors to sign up for egghead on the membership page because no membership is found. This PR checks if `is_instructor` is set to true and short circuits membership logic to let them know they are an instructor. Also added that if they have an active membership to let egghead staff know and we can cancel it for them.

vip

Desktop
mobile

### DIFF
--- a/src/hooks/use-account.ts
+++ b/src/hooks/use-account.ts
@@ -9,8 +9,6 @@ export const useAccount = () => {
 
   const isInstructor = viewer?.is_instructor
 
-  console.log({isInstructor})
-
   const isActiveAccountMember = userAccounts?.some(
     (account: {members: {id: number}[]}) => {
       return account.members?.find((member: {id: number}) => {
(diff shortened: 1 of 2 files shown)


src/hooks/use-account.ts
```diff
@@ -7,6 +7,8 @@ export const useAccount = () => {
   const {data: userAccounts, status: accountLoadingStatus} =
     trpc.user.accountsForCurrent.useQuery()
 
+  const isInstructor = viewer?.is_instructor
```
zacjones93, line 10: Already had is_instructor property being set on viewer

src/pages/user/membership.tsx
```diff
@@ -20,9 +20,27 @@ const Membership = () => {
...
+              </h3>
+              <p className="mt-4">
+                You have lifetime access to all of our courses. If you have an
+                active membership, please reach out to egghead staff if you'd
+                like it canceled.
+              </p>
+            </div>
+          </div>
+        </>
+      )
```
zacjones93, line 43: first thing to check is if they are instructor

joelhooks (approved):

PR #1295, opened 2023-02-21
Remove download from multi-module course

zacjones93:
This separates multi-module courses into their own template which allows us to remove a good amount of conditional logic from the standard course page layout.

"Multi-module" courses are a playlist of playlists. When users go to download the course on a multi-module course page, the download doesn't work as expected because we don't handle downloading playlists on the back end. This results in an empty folder being downloaded.

This PR removes the download button from the multi-module course page. 

This also truncates the seo page description being set to a max of 155 chars which is ideal.

missing download

**SEO Before:**
image
**SEO After:**
seo description truncated

**Multimodule:**
multi-module course
**typical course:**
typical course

### DIFF
--- a/src/components/layouts/collection-page-layout.tsx
+++ b/src/components/layouts/collection-page-layout.tsx
@@ -7,7 +7,7 @@ import {useQuery} from '@tanstack/react-query'
 import InstructorProfile from 'components/pages/courses/instructor-profile'
 import PlayIcon from 'components/pages/courses/play-icon'
 import getDependencies from 'data/courseDependencies'
-import {get, first, filter, isEmpty, take} from 'lodash'
+import {get, first, filter, isEmpty, take, truncate} from 'lodash'
 import {NextSeo} from 'next-seo'
 import removeMarkdown from 'remove-markdown'
 import {useViewer} from 'context/viewer-context'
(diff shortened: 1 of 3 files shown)


src/components/layouts/collection-page-layout.tsx
```diff
@@ -413,7 +406,7 @@ const CollectionPageLayout: React.FunctionComponent<CoursePageLayoutProps> = ({
   return (
     <>
       <NextSeo
-        description={removeMarkdown(description)}
+        description={truncate(removeMarkdown(description), {length: 155})}
```
zacjones93, line 409: truncate seo description

src/pages/courses/[slug]/index.tsx
```diff
@@ -29,25 +31,33 @@ const Course: React.FC<CourseProps> = (props) => {
...
-    </>
+      )
+  }
+
+  return (
+    <CollectionPageLayout
+      lessons={courseLessons}
+      course={course}
+      ogImageUrl={`https://og-image-react-egghead.now.sh/playlists/${slug}?v=20201103`}
+    />
```
Comment (zacjones93) [inline src/pages/courses/[slug]/index.tsx:60]: add a condition for multi-module courses

joelhooks (approved):

PR #1296, opened 2023-02-22
Fixes EGG-173: display lesson completion

DrShpongle:
This adds lesson completion label on the lessons cards on the search page. However, the approach I used doesn't work with lessons that don't belong to any collection/playlist (like this one: https://egghead.io/lessons/react-using-react-with-the-fullcalendar-jquery-plugin), I didn't find the solution for this case yet.

completed

### DIFF
--- a/src/lib/users.ts
+++ b/src/lib/users.ts
@@ -1,5 +1,6 @@
 import axios from 'axios'
 import gql from 'graphql-tag'
+import {flatten} from 'lodash'
 import getAccessTokenFromCookie from 'utils/get-access-token-from-cookie'
 import {getGraphQLClient} from '../utils/configured-graphql-client'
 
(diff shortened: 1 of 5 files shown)


src/lib/users.ts
```diff
@@ -239,3 +240,46 @@ export async function loadUserCompletedCourses(token?: string): Promise<any> {
...
+        }
+      }
+    }
+  `
+  token = token || getAccessTokenFromCookie()
+  const graphQLClient = getGraphQLClient(token)
+
+  const {user_progress} = await graphQLClient.request(query)
+
+  const completeLessons = flatten(
```
DrShpongle, line 270: here I create the array of slugs of completed lessons from collections that are in progress

PR #1298, opened 2023-02-25
Fixes EGG-173: display lesson completion

DrShpongle:
This puts progress back to the Lesson page. Also amount of watched lessons is shown for the course that has been started, but not completed:

Image 2023-02-24 at 18 27 16

completed GIF


joelhooks (approved):
🎉

PR #1307, opened 2023-04-14
feat: touch up callout component

Creeland:
Made a few tweaks to the Callout component:

Added text prop for easier text rendering, falls back to children if not provided. I noticed that the callout had some whitespace due to "children" not just being the text, but the html that was created from the text when the markdown was compiled. 

I also cleaned up the component and added some additional text styling. 

## Old

image
image

## New

image
image

### DIFF
--- a/src/components/mdx/callout.js
+++ b/src/components/mdx/callout.js
@@ -1,13 +1,9 @@
-import React from 'react'
-
-const container = ''
-
-export default function Callout({children, className}) {
+export default function Callout({children, text, className}) {
   return (
     <div
-      className={`mx-auto my-3 w-full dark:bg-gray-800 bg-blue-100 border-l-4 overflow-hidden border-blue-900 dark:text-blue-200 text-blue-600 sm:px-8 px-4 py-3 rounded-sm`}
+      className={`leading-relaxed mx-auto my-3 w-full dark:bg-gray-800 bg-blue-100 border-l-4 border-blue-900  sm:px-8 px-4 py-6 rounded-sm font-small sm:text-lg text-base tracking-normal shadow-md ${className}`}
     >
-      {children}
+      {text ? text : children}
     </div>
   )
 }
--- a/src/components/mdx/callout.js
+++ b/src/components/mdx/callout.js
@@ -1,9 +1,0 @@
-export default function Callout({children, text, className}) {
-  return (
-    <div
-      className={`p-5 leading-relaxed mx-auto my-3 w-full dark:bg-gray-800 bg-blue-100 border-l-4 border-blue-900 sm:space-x-5 space-x-0 sm:space-y-0 space-y-5 items-center text-left shadow-sm rounded-lg overflow-hidden lg:text-lg sm:text-md ${className}`}
-    >
-      {text ? text : children}
-    </div>
-  )
-}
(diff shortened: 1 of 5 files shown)


src/components/mdx/callout.js
```diff
@@ -1,13 +1,9 @@
-import React from 'react'
-
-const container = ''
-
-export default function Callout({children, className}) {
+export default function Callout({children, text, className}) {
   return (
     <div
-      className={`mx-auto my-3 w-full dark:bg-gray-800 bg-blue-100 border-l-4 overflow-hidden border-blue-900 dark:text-blue-200 text-blue-600 sm:px-8 px-4 py-3 rounded-sm`}
+      className={`leading-relaxed mx-auto my-3 w-full dark:bg-gray-800 bg-blue-100 border-l-4 border-blue-900  sm:px-8 px-4 py-6 rounded-sm font-small sm:text-lg text-base tracking-normal shadow-md ${className}`}
```
vojtaholik, line 4: `font-small` is not a valid tailwind class

Creeland:
Updated the component to be more consistent with our horizontal resource card spacing. I also dropped the class that didn't exist that I had mistakenly used @vojtaholik 

image
  
image

src/components/mdx/callout.js
```diff
@@ -1,13 +1,9 @@
-import React from 'react'
-
-const container = ''
-
-export default function Callout({children, className}) {
+export default function Callout({children, text, className}) {
   return (
     <div
-      className={`mx-auto my-3 w-full dark:bg-gray-800 bg-blue-100 border-l-4 overflow-hidden border-blue-900 dark:text-blue-200 text-blue-600 sm:px-8 px-4 py-3 rounded-sm`}
+      className={`p-5 leading-relaxed mx-auto my-3 w-full dark:bg-gray-800 bg-blue-100 border-l-4 border-blue-900 sm:space-x-5 space-x-0 sm:space-y-0 space-y-5 items-center text-left shadow-sm rounded-lg overflow-hidden lg:text-lg sm:text-md ${className}`}
```
vojtaholik, line 4: using `className` like this without making it an empty string as a fallback will make it return `undefined` by default.

src/components/mdx/callout.js
```diff
@@ -1,13 +1,9 @@
-import React from 'react'
-
-const container = ''
-
-export default function Callout({children, className}) {
+export default function Callout({children, text, className}) {
```
vojtaholik, line 1: why not convert it to TS while you're there? :)

vojtaholik:
I used actual links, fixed horizontal scrolling caused by code blocks by reducing their negative horizontal margin, and made the callouts stretch to full width on mobile.

screenshots

src/components/mdx/callout.tsx
```diff
@@ -0,0 +1,29 @@
...
+  return (
+    <div className="sm:mx-auto -mx-2">
+      <div
+        className={twMerge(
+          'sm:py-5 sm:px-5 py-3 px-5 leading-relaxed my-3 w-full dark:bg-gray-800 bg-blue-200/20 border-l-4 dark:border-blue-600 border-blue-500 items-center text-left shadow-sm sm:rounded-md overflow-hidden lg:text-lg sm:text-md',
+          className,
+        )}
+        aria-label="callout"
+      >
+        {text ? <p>{text}</p> : children}
```
vojtaholik, line 23: we want paragraphs for seo and a11y

vojtaholik (approved):

PR #1308, opened 2023-04-19
align titles and descriptions of sections to the left on curated pages

zacjones93:
When there is no image present for an article, it looks a little off to have the titles and descriptions centered when the topic title and description is left aligned. 

This also adds a little padding to the top of the sections so theres sufficient space when the `Levels` component isn't present.

# before
In-Depth Redux Tutorials for 2023  egghead io

# after
In-Depth Redux Tutorials for 2023  egghead io
In-Depth Redux Tutorials for 2023  egghead io

### DIFF
--- a/src/components/card/new-vertical-resource-card.tsx
+++ b/src/components/card/new-vertical-resource-card.tsx
@@ -18,6 +18,7 @@ import {Textfit} from 'react-textfit'
 import analytics from 'utils/analytics'
 import Heading from './heading'
 import CheckIcon from 'components/icons/check'
+import {twMerge} from 'tailwind-merge'
 
 const VerticalResourceCard: React.FC<{
   resource: CardResource
@@ -54,7 +55,14 @@ const VerticalResourceCard: React.FC<{
       instructor={resource?.instructor?.name}
       tag={resource.tag?.name}
     >
-      <Card {...props} resource={resource} className={className}>
+      <Card
+        {...props}
+        resource={resource}
+        className={twMerge(
+          'rounded-md aspect-w-3 aspect-h-4 w-full h-full transition-all ease-in-out duration-200 relative overflow-hidden group dark:bg-gray-800 bg-white dark:bg-opacity-60 shadow-smooth dark:hover:bg-gray-700 dark:hover:bg-opacity-50',
+          className,
+        )}
+      >
         {resource.background && (
           <Image
             src={resource.background}
--- a/src/components/search/curated/[slug]/index.tsx
+++ b/src/components/search/curated/[slug]/index.tsx
@@ -12,6 +12,7 @@ import {CARD_TYPES} from 'components/search/curated/curated-essential'
 import {useRouter} from 'next/router'
 import {useViewer} from 'context/viewer-context'
 import {loadUserCompletedCourses} from 'lib/users'
+import {twMerge} from 'tailwind-merge'
 
 const useUserCompletedCourses = (viewerId: number) => {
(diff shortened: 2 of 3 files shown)


src/components/card/new-vertical-resource-card.tsx
```diff
@@ -54,7 +55,14 @@ const VerticalResourceCard: React.FC<{
...
     >
-      <Card {...props} resource={resource} className={className}>
+      <Card
+        {...props}
+        resource={resource}
+        className={twMerge(
+          'rounded-md aspect-w-3 aspect-h-4 w-full h-full transition-all ease-in-out duration-200 relative overflow-hidden group dark:bg-gray-800 bg-white dark:bg-opacity-60 shadow-smooth dark:hover:bg-gray-700 dark:hover:bg-opacity-50',
+          className,
+        )}
+      >
```
zacjones93, line 65: merging tailwind classes instead of overriding completely

src/components/search/curated/[slug]/index.tsx
```diff
@@ -27,6 +28,102 @@ type CuratedTopicProps = {
   topic: any
 }
 
+const DynamicCardGrid = ({
```
Comment (zacjones93) [inline src/components/search/curated/[slug]/index.tsx:31]: extracted the default card grid behavior into a class

src/components/search/curated/[slug]/index.tsx
```diff
@@ -129,21 +226,29 @@ const CuratedTopic: React.FC<CuratedTopicProps> = ({topic, topicData}) => {
...
+                        </h2>
+                      </div>
+                      <DynamicCardGrid
+                        section={section}
+                        completedCoursesIds={completedCoursesIds}
+                        location={location}
+                      />
+                    </section>
+                  )
+                case section.resources.length === 2:
```
Comment (zacjones93) [inline src/components/search/curated/[slug]/index.tsx:248]: using a switch case to add special cases we want to handle. 

Right now we want to handle sections:
- that dont have a description or image 
- that have 2 resources
- that have 4 resources and no image

everything else will continue to be default behavior down below

src/components/search/curated/[slug]/index.tsx
```diff
@@ -156,94 +261,125 @@ const CuratedTopic: React.FC<CuratedTopicProps> = ({topic, topicData}) => {
...
-                              completedCoursesIds={completedCoursesIds}
-                            />
-                          )
-                        default:
+                      <div className="grid md:grid-cols-4 gap-4 ">
+                        <div
+                          className={`col-span-2  ${cx({
+                            'md:order-last': i % 2 !== 0,
+                          })}`}
+                        >
```
Comment (zacjones93) [inline src/components/search/curated/[slug]/index.tsx:269]: flipping the text to the opposite column because having text stacked on the left side looked a little off. Maybe we want to do this differently?

This is what it looks like without the flipped text:
In-Depth Redux Tutorials for 2023  egghead io

src/components/search/curated/[slug]/index.tsx
```diff
@@ -156,94 +261,125 @@ const CuratedTopic: React.FC<CuratedTopicProps> = ({topic, topicData}) => {
...
+                            return (
+                              <VerticalResourceCard
+                                className="lg:hidden"
+                                key={resource.title}
+                                resource={resource}
+                                location={location}
+                                completedCoursesIds={completedCoursesIds}
+                              />
+                            )
+                          })}
```
Comment (zacjones93) [inline src/components/search/curated/[slug]/index.tsx:345]: When the screen fits the content, we split the resources and display 2 courses as horizontal cards and 2 as vertical. When it is smaller we go to the default which is vertical for all 4

vojtaholik (approved):

PR #1310, opened 2023-04-21
Update e2e GitHub Actions versions

johnlindquist:
fail fix

The action started failing due to outdated versions. Updating should fix it.


vojtaholik (approved):
🙏 🎉

PR #1311, opened 2023-04-24
[fixes EGG-192]: Convert article resource image on cards to cover instead of contain

zacjones93:
Our cards don't support article images very well, they assume that the image is square and small (like podcast and course resources) where that is not the case.

The result is the images get contorted and look out of place, see down below.

## Before
desktop - before showing square images with awkward padding

## After
desktop after showing image covering top and left side of cards
mobile after showing image covering top and left side of cards

cats chillin

### DIFF
--- a/src/components/card/new-horizontal-resource-card.tsx
+++ b/src/components/card/new-horizontal-resource-card.tsx
@@ -154,7 +154,7 @@ const PreviewImage: React.FC<{title: string; image: any; name: string}> = ({
       case 'talk':
         return 80
       case 'article':
-        return 300
+        return 500
       default:
         return 200
     }
@@ -171,8 +171,8 @@ const PreviewImage: React.FC<{title: string; image: any; name: string}> = ({
       <Image
         aria-hidden
         src={get(image, 'src', image)}
-        width={name === 'article' ? 500 : getSize(name)}
-        height={name === 'article' ? 500 : getSize(name)}
+        width={getSize(name)}
+        height={getSize(name)}
         quality={100}
         alt=""
       />
--- a/src/components/card/new-horizontal-resource-card.tsx
+++ b/src/components/card/new-horizontal-resource-card.tsx
@@ -54,8 +54,12 @@ const HorizontalResourceCard: React.FC<{
       tag={resource.tag?.name}
     >
       <Card {...props} resource={resource} className={defaultClassName}>
-        <CardContent className="grid grid-cols-8 gap-5 items-center px-5 py-2">
-          <CardHeader className="col-span-3 flex items-center justify-center">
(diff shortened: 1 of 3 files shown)


src/components/card/new-horizontal-resource-card.tsx
```diff
@@ -54,8 +54,12 @@ const HorizontalResourceCard: React.FC<{
       tag={resource.tag?.name}
     >
       <Card {...props} resource={resource} className={defaultClassName}>
-        <CardContent className="grid grid-cols-8 gap-5 items-center px-5 py-2">
-          <CardHeader className="col-span-3 flex items-center justify-center">
+        <CardContent className="grid grid-cols-8 gap-5   items-center px-5 py-2">
+          <CardHeader
+            className={`col-span-3 flex items-center justify-center ${cx({
+              '-ml-32 -my-64': resource.name === 'article',
+            })}`}
```
zacjones93, line 61: if resource is an article we want to position the card to the left and applying negative margin on `y` spreads the image across the card vertically

src/components/card/new-horizontal-resource-card.tsx
```diff
@@ -149,6 +153,8 @@ const PreviewImage: React.FC<{title: string; image: any; name: string}> = ({
         return 40
       case 'talk':
         return 80
+      case 'article':
```
zacjones93, line 156: make the image bigger when it's an article

src/components/card/new-vertical-resource-card.tsx
```diff
@@ -73,11 +74,20 @@ const VerticalResourceCard: React.FC<{
...
         )}
         <CardContent className="grid grid-rows-6 xl:p-5 p-2 pt-5">
-          <CardHeader className={`row-span-3 relative `}>
+          <CardHeader
+            className={`
+            row-span-3 
+            relative 
+            ${cx({
+              '-mx-5 -mt-5': resource.name === 'article',
+            })}
```
zacjones93, line 83: spreads the image horizontally across vertical cards and moves the image up top

src/components/card/new-vertical-resource-card.tsx
```diff
@@ -161,18 +171,19 @@ export const ResourceLink: React.FC<{
...
+  resourceType: string
+}> = ({title, image, small, resourceType}) => {
   if (!image) return null
 
   return (
     <Image
       aria-hidden
       src={get(image, 'src', image)}
-      objectFit="contain"
+      objectFit={resourceType === 'article' ? 'cover' : 'contain'}
```
zacjones93, line 186: when resource type is article we want cover, not contain for object fit.

PR #1315, opened 2023-04-27
fixes [egg-83]: allow user to remove GitHub connection

zacjones93:
On egghead-next, users could connect their GitHub accounts to egghead to log in but we weren't giving them the option to disconnect their account if they wanted to.

This adds the disconnect functionality through trpc and graphql.

disconnect

[Loom demo](https://www.loom.com/share/18df73b47de045a68df662766c256320)


src/components/pages/user/tabs-content/profile-tab-content.tsx
```diff
@@ -24,30 +26,47 @@ const ProfileTabContent: React.FC<any> = () => {
...
+        isConnectedToGithub={isConnectedToGithub}
+        authToken={authToken}
+      />
     </div>
   )
 }
 
 export default ProfileTabContent
 
+const GithubDisConnectButton: React.FunctionComponent<{
```
jbranchaud, line 39: I think the `C` in `DisConnect` should be lowercase (otherwise, this PR looks great to me 🙌)

joelhooks (approved):

PR #1321, opened 2023-05-16
fix: load tags from cloudinary

zacjones93:
This doesn't fix the underlying issue of tag images being broken coming from rails (we should probably migrate lessons to sanity but this is a big task) but this fixes images by loading tag images from cloudinary based on tag name that is present.

tag

### DIFF
--- a/src/components/layouts/collection-page-layout.tsx
+++ b/src/components/layouts/collection-page-layout.tsx
@@ -885,6 +885,9 @@ const CollectionPageLayout: React.FunctionComponent<CoursePageLayoutProps> = ({
               <div>
                 <ul>
                   {lessons.map((lesson: LessonResource, index: number) => {
+                    const tagImageUrl = `https://res.cloudinary.com/dg3gyk0gu/image/upload/v1683914713/tags/${lesson?.primary_tag?.name}.png`
+                    const optimizedImageUrl = `https://res.cloudinary.com/dg3gyk0gu/image/fetch/w_72,h_72/${tagImageUrl}`
+
                     const isComplete = completedLessonSlugs?.includes(
                       lesson.slug,
                     )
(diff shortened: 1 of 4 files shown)


src/components/layouts/collection-page-layout.tsx
```diff
@@ -885,6 +885,9 @@ const CollectionPageLayout: React.FunctionComponent<CoursePageLayoutProps> = ({
               <div>
                 <ul>
                   {lessons.map((lesson: LessonResource, index: number) => {
+                    const tagImageUrl = `https://res.cloudinary.com/dg3gyk0gu/image/upload/v1683914713/tags/${lesson?.primary_tag?.name}.png`
+                    const optimizedImageUrl = `https://res.cloudinary.com/dg3gyk0gu/image/fetch/w_72,h_72/${tagImageUrl}`
+
```
zacjones93, line 890: optimized image from cloudinary

src/components/layouts/collection-page-layout.tsx
```diff
@@ -905,10 +908,10 @@ const CollectionPageLayout: React.FunctionComponent<CoursePageLayoutProps> = ({
                                 index + 1
                               )}
                             </div>
-                            {lesson.icon_url && (
+                            {lesson.primary_tag && (
                               <div className="flex items-center flex-shrink-0 w-8">
                                 <Image
-                                  src={lesson.icon_url}
+                                  src={optimizedImageUrl}
```
zacjones93, line 914: use cloudinary url

src/lib/playlists.ts
```diff
@@ -223,6 +223,9 @@ export async function loadPlaylist(slug: string, token?: string) {
             created_at
             updated_at
             published_at
+            primary_tag {
+              name
+            }
```
zacjones93, line 228: fetch primary tag name

src/components/layouts/collection-page-layout.tsx
joelhooks: i'm positive you can just add the transform to the first url instead of passing it to fetch

joelhooks (approved):

PR #1322, opened 2023-05-16
fix: pull default course image from cloudinary

zacjones93:
Realized that the course artwork pulls in the primary tag image as well when there is no custom artwork supplied. This patches that to return the first tag set in sanity as the default course image

## broken prod
broken

## fixed local
fixed

again

### DIFF
--- a/src/lib/courses.ts
+++ b/src/lib/courses.ts
@@ -6,6 +6,7 @@ const courseQuery = groq`
 *[_type == 'resource' && externalId == $courseId][0]{
   title,
   challengeRating,
+  "square_cover_480_url": image,
   description,
   summary,
   essentialQuestions[]->{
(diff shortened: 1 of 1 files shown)


src/lib/courses.ts
```diff
@@ -80,8 +81,11 @@ export async function loadCourseMetadata(id: number) {
...
   const course = await sanityClient.fetch(courseQuery, params)
 
-  if (course?.illustration?.url) {
-    course['square_cover_480_url'] = course.illustration.url
+  if (!course?.square_cover_480_url) {
+    const imageUrl = course?.dependencies
+      ? `https://res.cloudinary.com/dg3gyk0gu/image/upload/v1683914713/tags/${course?.dependencies[0]?.name}.png`
+      : 'https://res.cloudinary.com/dg3gyk0gu/image/upload/v1569292667/eggo/eggo_flair.png'
+    course['square_cover_480_url'] = imageUrl
   }
```
zacjones93, line 89: we don't actually set a `course.illustration.url` so switched this logic out for when no image is set

joelhooks (approved):

PR #1324, opened 2023-05-24
Article Resource Grid and email course CTA

Creeland:
I added an additional CTA and a resource grid for articles. 

The CTA adds / updates a customer in CIO, adding attributes to their profile for each of their selected interests. I've set it up so that users with the `article_cta_portfolio` attribute get put into a segment that is added to the portfolio campaign.

This uses the new graphql query that @joelhooks set up for getting a user's contact. https://github.com/skillrecordings/egghead-rails/pull/5215/files

# CTA

## Default
image

## Small Viewport
image

# Resource Grid

## Three items
image

## One Item
image

## Small Viewport
image


src/components/mdx/email-subscribe-widget.tsx
```diff
@@ -0,0 +1,201 @@
...
+            </li>
+            {/* TODO: Make these campaigns in CIO */}
+            {/* <li className="pb-2">
+              ⭐️ <b>Full-Stack in 2023:</b> Build several full-stack apps using
+              the latest tools and trends (Open AI, Next.js, Remix, and more)
+            </li>
+            <li>
+              ⭐️ <b>TypeScript:</b> Build a solid foundation in TypeScript and
+              learn some advanced techniques along the way
+            </li> */}
```
Creeland, line 181: The initial plan was to create these campaigns but in the interest of saving time I'm starting with just the portfolio campaign since it already exists. 

Phase two of this will be to make this CTA dynamic using data from Sanity.

src/config.ts
```diff
@@ -2,3 +2,5 @@ export const ACCESS_TOKEN_KEY = 'eh_token_2020_11_22'
 export const EGGHEAD_USER_COOKIE_KEY = 'eh_user'
 export const CIO_IDENTIFIER_KEY = 'cio_id'
 export const CIO_CUSTOMER_OBJECT_KEY = 'cio_customer'
+
+export const ARTICLE_NEWSLETTER_INTEREST_KEY = 'article_cta_signed_up'
```
Creeland, line 6: Continuing the pattern used above for creating the cookie name

Creeland: Having a dedicated cookie made the most sense since some existing users may be interested in the email courses offered. So I didn't want to hide the CTA based on subscriber status

src/lib/customer-io.ts
```diff
@@ -0,0 +1,12 @@
+const {RegionUS, APIClient} = require('customerio-node')
+const appApiKey = process.env.CUSTOMER_IO_APPLICATION_API_KEY
+const api = new APIClient(appApiKey, {region: RegionUS})
+
+export const getAttributes = async (user_contact: string) => {
```
Creeland, line 5: put this into a lib since I thought it could potentially be useful elsewhere

joelhooks (approved):
Article Resource Grid and email course CTA by Creeland · Pull Request #1324 · skillrecordingsegghead-next 2023-06-01 at 15 54 31@2x

I think requiring this decision is likely not required

How will results be tracked?

PR #1335, opened 2023-06-21
fixes course metadata

zacjones93:
fixed

## broken
broken metadata

## fixed
fixed metadata

### DIFF
--- a/src/lib/courses.ts
+++ b/src/lib/courses.ts
@@ -5,7 +5,7 @@ import {z} from 'zod'
 
 const courseQuery = groq`
 *[
-  (_type == 'resource' || _type == 'course') && (externalId == '$courseId' || slug.current == $slug)
+  (_type == 'resource' || _type == 'course') && (externalId == $courseId || slug.current == $slug)
 ][0]{
   "id": _id,
   title,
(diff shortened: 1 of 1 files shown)


src/lib/courses.ts
```diff
@@ -5,7 +5,7 @@ import {z} from 'zod'
 
 const courseQuery = groq`
 *[
-  (_type == 'resource' || _type == 'course') && (externalId == '$courseId' || slug.current == $slug)
+  (_type == 'resource' || _type == 'course') && (externalId == $courseId || slug.current == $slug)
```
zacjones93, line 8: passing `$courseId` as a string breaks the query

src/lib/courses.ts
```diff
@@ -80,7 +80,21 @@ const courseQuery = groq`
...
+    "description": summary,
+    path,
+    image,
+    'instructor': collaborators[]->[role == 'instructor']{
+    	'name': person->name,
+  	}
+	},
+	"customOgImage": images[label == 'og-image'][0]{
+    url
+  },
```
zacjones93, line 97: `pairWithResources` was removed so only defaults were showing for courses

joelhooks (approved):

PR #1348, opened 2023-08-15
migrate tippy to radix for lesson page download

zacjones93:
Removes tippy in favor of Radix, need to checkout out preview to see if errors persist related to Next.js 13 upgrade

https://www.loom.com/share/687cf734d5184f0aa0617468b7493647

still hydration errors but less

image

g tippy


src/components/player/download-control.tsx
```diff
@@ -84,24 +101,11 @@ const DownloadControl: FunctionComponent<
...
-            download_url={download_url}
-            state={state}
-          />
-        </div>
-      </Tippy>
+      <DownloadButtonWithTooltip
+        slug={slug}
+        download_url={download_url}
+        state={state}
+      />
```
zacjones93, line 108: put tooltip logic in the button component. Wrapping the functional component in `<Tooltip.Trigger>` had a refs issue that I wasn't quite sure how to fix.

PR #1349, opened 2023-08-15
dynamically import lesson component

zacjones93:
With React 18, when you have components rendering dynamic data you'll get hydration errors like we've been experiencing.

The lesson page has quite a lot of dynamic data being loaded. Splitting it into its own component and dynamically importing it fixes these hydration errors.

reference: https://github.com/vercel/next.js/discussions/35773#discussioncomment-2840696

Not sure if we can be more granular with what we are dynamically loading. hard to tell but I believe it's the components we import from `@skillrecordings/player` that are the most dynamic in nature.

image

g hydration


src/pages/lessons/[slug].tsx
```diff
@@ -1,74 +1,23 @@
...
-import {useSelector} from '@xstate/react'
 import {addCueNote, deleteCueNote} from '../../lib/notes'
-import {CheckCircleIcon as CheckCircleIconOutline} from '@heroicons/react/outline'
-import {CheckCircleIcon} from '@heroicons/react/solid'
-import {trpc} from 'trpc/trpc.client'
-import {LessonProgress} from 'lib/progress'
-import PlayerSidebar from '../../components/player/player-sidebar'
 import {GenericErrorBoundary} from '../../components/generic-error-boundary'
+import dynamic from 'next/dynamic'
+const Lesson = dynamic(() => import('./components/lesson'), {ssr: false})
```
Comment (zacjones93) [inline src/pages/lessons/[slug].tsx:20]: this loads lesson component dynamically

Comment (zacjones93) [reply src/pages/lessons/[slug].tsx:20]: we could also provide a loading state if we wanted to but it loads pretty quick

```
loading: () => <div className='flex items-center justify-center bg-gray-100 dark:bg-gray-1000 dark:border-gray-800 w-full h-[100vh] '>
      <p>Loading...</p>
  </div>,
  ```

PR #1352, opened 2023-08-21
bug: infinite redirect when sanity course slug != to actual course slug

zacjones93:
The slug that is generated in sanity is not based on rails data. When a sanity course document has had it's slug generated on the sanity side (an indeterminate number of courses...) this causes an infinite loop because we have a check `(course && course?.slug !== params.slug)` that retries when the check fails. 

I'm not sure why this behavior changed from before other than having to do with Next/React upgrades..

This fix prefers the course path which is always generated from the rails object.

g checked

### DIFF
--- a/src/pages/courses/[slug]/index.tsx
+++ b/src/pages/courses/[slug]/index.tsx
@@ -115,8 +115,8 @@ export const getServerSideProps: GetServerSideProps = async ({
 
   try {
     const course = params && (await loadPlaylist(params.slug as string))
-
-    if (course && course?.slug !== params?.slug) {
+    console.log({req})
+    if (course && course?.path !== req.url) {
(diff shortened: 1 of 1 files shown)


src/pages/courses/[slug]/index.tsx
```diff
@@ -115,8 +115,7 @@ export const getServerSideProps: GetServerSideProps = async ({
 
   try {
     const course = params && (await loadPlaylist(params.slug as string))
-
-    if (course && course?.slug !== params?.slug) {
+    if (course && course?.path !== req?.url) {
```
Comment (zacjones93) [inline src/pages/courses/[slug]/index.tsx:118]: same check but `path` is more consistent than `slug`

joelhooks (approved):

PR #1356, opened 2023-09-06
Zj/remove with prose remark plugin

zacjones93:
The `withProse` plugin was not functioning correctly. It looks like it was a [common community solution](https://gist.github.com/samselikoff/38c59bdff8ca619a87c28e9c8d8b99f9) a while back that escapes `prose` for MDX components.  We are applying `not-prose` to our MDX components anyway so this plugin isn't needed even if it was working.

This PR removes the plugin in favor of directly applying `prose` to pages and `not-prose` to MDX components

Learn SvelteKit - Part 4 Form Actions  egghead io

g get out


src/components/mdx/article-series-list.tsx
```diff
@@ -21,6 +21,7 @@ const ArticleSeriesList: React.FC<
         {articles.map((article: any, i: number) => {
           return (
             <Link
+              key={i}
```
zacjones93, line 24: an error I found in the console

tailwind.config.js
```diff
@@ -168,7 +168,6 @@ module.exports = {
     require('@tailwindcss/typography'),
     require('@tailwindcss/forms'),
     require('@tailwindcss/aspect-ratio'),
-    require('@tailwindcss/line-clamp'),
```
zacjones93, line 171: another error being logged that was annoying. line-clamp is automatically included now

src/pages/blog/[slug].tsx
```diff
@@ -103,7 +102,7 @@ const Tag = (props: any) => {
               </div>
             )}
           </header>
-          <main>
+          <main className="prose dark:prose-dark sm:prose-lg lg:prose-xl max-w-none dark:prose-a:text-blue-300 prose-a:text-blue-500">
```
Comment (zacjones93) [inline src/pages/blog/[slug].tsx:105]: just apply prose classes to main

Comment (zacjones93) [reply src/pages/blog/[slug].tsx:105]: `not-prose` applied to MDX components down below

``` javascript
CourseWidget: ({course: courseSlug, ...props}: any) => {
  const course = find(resources, {slug: courseSlug})
  return course ? (
    <div className="not-prose my-8">
      <CourseWidget course={course} {...props} />
    </div>
  ) : null
},
```

src/components/mdx/article-series-list.tsx
joelhooks: you shouldn't use index for key in general and use article slug or similar instead
```suggestion
              key={article.slug}
```

PR #1357, opened 2023-09-06
Fix course page rendering

zacjones93:
Upgrading `ReactMarkdown` broke it on the course page 😅 While I was in there I found some hydration errors similar to the lesson page that is fixed with dynamic imports. 

g fixed


zacjones93:
back in business

PR #1359, opened 2023-09-08
migrate sanity to v3

zacjones93:
This PR migrates sanity to v3.

Some plugins aren't ready yet like `sanity-plugin-autocomplete-tags` so I left that commented out. That's not a feature we really use anyway. `sanity-plugin-content-calendar`, `sanity-plugin-graph-view`, `sanity-plugin-nyancat-spinner`, and `sanity-plugin-json-to-docs` are also not currently being used in the v3 studio but I don't think we use those plugins either.

I upgraded the `sanity-plugin-markdown` which works as expected. `sanity-plugin-cloudinary` is also present but not being used.

The biggest change is converting to TypeScript. Documents now use TypeScript but I saw objects in partner products not using TypeScript so kept the same pattern here. 

There is one part that I haven't implemented which is the `path-input.tsx` behavior. I replaced it with the generic `CustomStringField` example but in the studio and I can't tell a difference.

I ran through the whole studio clicking each type of document and their references and all is well. I'm not sure what type of testing should be done on the egghead-next side but I've clicked through pages there as well and things are rendering as expected.

egghead-next – Desk

mario


joelhooks (approved):
nice one!

PR #1366, opened 2023-09-18
Rework usage of aspect ratio class

DrShpongle:
Got rid of `@tailwindcss/aspect-ratio` plugin, substituted `aspect-w-XX` and `aspect-h-XX` classes with `aspect-XX` that is core class in last versions of tailwindcss.

giphy


zacjones93 (approved):
Looks good!

PR #1370, opened 2023-09-25
adds a sanity model for email broadcasts

joelhooks:
broadcast


zacjones93 (approved):
This looks good to me. I assume email templates would be stored in egghead-next?

PR #1372, opened 2023-10-06
migrate trpc to app router

zacjones93:
This allows us to use trpc in the app router/in server components.

The main difference here is client components (and anything under the pages router) will need to use `trpc` exported from `app/_trpc/client` where server components will use `serverClient` from `app/_trpc/serverClient`.

The hard part about this migration is that the Next trpc adapter doesn't support app router as of right now. I had to remove the adaptor which looked like it's primary role was to add properties to trpc context. Now we are manually providing context in `app/api/trpc/[trpc]/route`. Here I'm including prisma, the request (if it's coming from the client, server components don't need that), and then added `userToken` as a convenience. followed [this example](https://github.com/solaldunckel/next-13-app-router-with-trpc/blob/master/apps/nextjs/src/pages/api/trpc/%5Btrpc%5D.ts) which generally looks like it has some good patterns.

cat


src/app/_trpc/serverClient.ts
```diff
@@ -0,0 +1,34 @@
...
+}
+
+export const serverClient = appRouter.createCaller({
+  links: [
+    httpBatchLink({
+      url: `${getBaseUrl()}/api/trpc`,
+    }),
+  ],
+  transformer,
+  prisma,
```
zacjones93, line 33: serverClient doesn't get request

src/app/api/trpc/[trpc]/route.ts
```diff
@@ -0,0 +1,36 @@
...
+        req,
+        userToken: req?.cookies?.get(ACCESS_TOKEN_KEY)?.value,
+      }
+    },
+    onError({error}) {
+      if (error.code === 'INTERNAL_SERVER_ERROR') {
+        console.error('Caught TRPC error:', error)
+      }
+    },
+  })
```
Comment (zacjones93) [inline src/app/api/trpc/[trpc]/route.ts:34]: adds what trpc/next was doing (from what I could tell). I like the convenience of also having `userToken` top level so we don't have to dig for it in each router

src/pages/_app.tsx
```diff
@@ -112,14 +112,16 @@ const App: React.FC<React.PropsWithChildren<AppProps>> = ({
...
+              <TrpcProvider>
+                <QueryClientProvider client={queryClient}>
+                  <MDXProvider components={mdxComponents}>
+                    {getLayout(Component, pageProps)}
+                  </MDXProvider>
+                  <div className="print:hidden">
+                    <ReactQueryDevtools />
+                  </div>
+                </QueryClientProvider>
+              </TrpcProvider>
```
zacjones93, line 124: wrapping app with provider instead of using `withTRPC` that comes from trpc/next

src/server/routers/progress.ts
```diff
@@ -1,12 +1,12 @@
...
+import {router, baseProcedure} from '../trpc'
 import {z} from 'zod'
 import {ACCESS_TOKEN_KEY} from '../../utils/auth'
 import {loadLessonProgress, loadPlaylistProgress} from '../../lib/progress'
 import {loadUserCompletedCourses} from 'lib/users'
 
 export const progressRouter = router({
   completedCourses: baseProcedure.query(async ({ctx}) => {
-    const token = ctx.req?.cookies[ACCESS_TOKEN_KEY]
+    const token = ctx?.userToken
```
zacjones93, line 9: nice to just use `usertoken`

src/server/routers/tips.ts
```diff
@@ -95,7 +93,7 @@ export const tipsRouter = router({
       }),
     )
     .mutation(async ({ctx, input}) => {
-      const ability = await getAbilityFromToken(ctx.req?.cookies[ACCESS_TOKEN_KEY])
+      const ability = await getAbilityFromToken(ctx?.userToken)
```
zacjones93, line 96: we could likely just expose ability top level on trpc context as well

src/server/trpc.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  return next({
+    ctx: {
+      ...ctx,
+      auth: ctx.auth,
+    },
+  });
+});
+
+*/
```
zacjones93, line 30: I saw an example of exporting an 'authed' Procedure so that way we wouldn't have to manually check for a token in every router that needs it. just add that logic in middleware here and we can know we are authed in the router where we use this procedure

```
const token = ctx?.userToken
 if (!token) return null
```

src/app/tips/page.tsx
```diff
@@ -2,14 +2,16 @@ import {Tip} from 'lib/tips'
...
 
 type TipsIndex = {
   tips: Tip[]
 }
 
 const TipsIndex: React.FC<any> = async () => {
-  const tips = await getAllTips()
+  const allTips = await serverClient.tips.all()
+  const publishedTips =
+    allTips.find((tipGroup) => tipGroup.state === 'published')?.tips ?? []
```
zacjones93, line 14: prefer using trpc over directly fetching

joelhooks (approved):
🔥

PR #1375, opened 2023-10-13
adds sanity webhook syncing tip document to rails

zacjones93:
This webhook is triggered whenever a Tip in sanity is created. Sanity sends the instructor ID and title to the webhook so we can create a lesson with that instructors ID in rails.

When we get the lesson ID back from rails, we set the `eggheadRailsLessonId` property on the Tip that was just created.

I met with @jbranchaud and @Creeland to talk through this flow as well as steps towards making Tips self-service for instructors.

This webhook handles the top flow (in purple) where next steps are adding duration to that lesson tied to the tip so we can track completions [[tldraw](https://www.tldraw.com/r/DnKZR78TwtgwpgZ86wboV?viewport=-1565%2C-802%2C2622%2C2967&page=page%3ACMaiQtTQpaTHeLDTFqci9)]

We see there being two more webhooks to create to get this functioning at a basic level. We need to hook into the Mux event when a video is done processing to grab the total duration of that video asset to set it in sanity and then turn around and have a sanity webhook that sets duration on the rails lesson.

image


src/pages/api/webhooks/sanity/tips/create.ts
```diff
@@ -0,0 +1,81 @@
+import type {NextApiRequest, NextApiResponse} from 'next'
+import {isValidSignature, SIGNATURE_HEADER_NAME} from '@sanity/webhook'
+import client from '@sanity/client'
+import axios from 'axios'
+import _get from 'lodash/get'
+
+const secret = process.env.SANITY_WEBHOOK_CREATED_SECRET || ''
+const railsToken = process.env.EGGHEAD_ADMIN_TOKEN || ''
```
zacjones93, line 8: this needed admin permissions on the POST request to rails to create a lesson for any instructor

zacjones93: not sure if I could do this another way but I just grabbed an admin access token locally and set that

src/pages/api/webhooks/sanity/tips/create.ts
```diff
@@ -0,0 +1,81 @@
...
+
+        await sanityClient
+          .patch(_id)
+          .set({
+            eggheadRailsLessonId: lesson.data.id,
+            eggheadRailsCreatedAt: new Date().toISOString(),
+          })
+          .commit()
+
+        return res.status(200).json({success: true})
```
zacjones93, line 61: the work

joelhooks (approved):

PR #1376, opened 2023-10-16
Set Video Duration in Sanity by consuming Mux Webhook events in egghead-next

zacjones93:
This is the second step to tracking views for egghead tips by adding the duration to video assets. Next we will want to add another sanity webhook to sync the tip that had it's video asset duration updated with the rails object that has been created in step 1 (#1375). 

[loom](https://www.loom.com/share/293a1c16e8074667ae744dd41a693a85?sid=35fe0449-6ff5-4bb1-a1cc-c1a52892804e)

watch

### DIFF
--- a/src/app/api/webhooks/tips/mux/route.ts
+++ b/src/app/api/webhooks/tips/mux/route.ts
@@ -1,0 +1,80 @@
+import {NextRequest, NextResponse} from 'next/server'
+import {headers} from 'next/headers'
+import Mux from '@mux/mux-node'
+import client from '@sanity/client'
+import _get from 'lodash/get'
+import groq from 'groq'
+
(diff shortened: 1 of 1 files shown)


src/app/api/webhooks/tips/mux/route.ts
```diff
@@ -0,0 +1,80 @@
...
+
+export async function POST(req: NextRequest) {
+  const signature = headers().get('mux-signature') as string
+  const muxRequestBody = await req.json()
+  const isValid = Mux.Webhooks.verifyHeader(
+    JSON.stringify(muxRequestBody),
+    signature,
+    secret,
+  )
+  console.log({isValid, muxRequestBody})
```
zacjones93, line 31:
```suggestion
```

PR #1381, opened 2023-10-25
Start tracking with Posthog

Creeland:
I've added Posthog and set it up so that's it's tracking all the events that are sent through `track`, just like Mixpanel. 

I also followed this [handy guide](https://posthog.com/tutorials/mixpanel-to-posthog) and imported the last year of data from Mixpanel into Posthog.


src/lib/posthog-client.ts
```diff
@@ -0,0 +1,26 @@
+import posthog from 'posthog-js'
+
+class PosthogClient {
```
Creeland, line 3: I made this class so that posthog only had to be initialized once and we could use the capture method wherever else we'd like.

src/pages/_app.tsx
```diff
@@ -114,12 +119,14 @@ const App: React.FC<React.PropsWithChildren<AppProps>> = ({
             <CioProvider>
               <QueryClientProvider client={queryClient}>
                 <TrpcProvider>
-                  <MDXProvider components={mdxComponents}>
-                    {getLayout(Component, pageProps)}
-                  </MDXProvider>
-                  <div className="print:hidden">
-                    <ReactQueryDevtools />
-                  </div>
+                  <PostHogProvider client={posthog}>
```
Creeland, line 122: This enables us to use their `usePosthog` hook and handle web tracking in our React components. Not using any of those features at the moment though.

Creeland: The providers pyramid grows ever taller

src/lib/posthog-client.ts
```diff
@@ -0,0 +1,26 @@
...
+        },
+        capture_pageview: false, // Disable automatic pageview capture, as we capture manually
+        autocapture: false, // Disable automatic default event capture, as we capture manually
+      })
+    }
+
+    return posthog
+  }
+
+  static capture(event: string, params: any) {
```
joelhooks, line 21: might consider renaming this `track` to be consistent with our existing usage/apis

does posthog have an analog for `identify` to associate events/sessions with a user?

joelhooks (approved):

PR #1382, opened 2023-10-25
track tip completions through download api

zacjones93:
https://www.loom.com/share/1559bf7dec5a4150a20b113af6690452?sid=a7af9dea-1137-4a10-b09f-37c688a0653a

This pr triggers lesson completion logic and displays whether or not a tip has been completed in the UI

This PR depends on [a graphql endpoint being merged in egghead-rails](https://github.com/skillrecordings/egghead-rails/pull/5236)

pieced together


src/server/routers/tips.ts
```diff
@@ -157,4 +157,53 @@ export const tipsRouter = router({
...
+            Authorization: `Bearer ${token}`,
+            'X-SITE-CLIENT': process.env.NEXT_PUBLIC_CLIENT_ID as string,
+            'Content-Type': 'application/json',
+          },
+
+          body: JSON.stringify({
+            id: tipId,
+          }),
+        }).then((res) => res.json())
+      })
```
zacjones93, line 203: I'm chaining two calls here. `/watch/download` creates a `LessonView` that has the proper segments and then `json.complete_url` flips the `did_complete` boolean to `true.

Originally I was using the `/watch/manual_complete` endpoint but that doesn't track segments watched which is what get's instructors paid.

image

src/components/tips/tip-teaser.tsx
```diff
@@ -2,11 +2,23 @@ import {useRouter} from 'next/navigation'
...
+  ArrowsExpandIcon,
+} from '@heroicons/react/outline'
+import {CheckCircleIcon, CheckIcon} from '@heroicons/react/solid'
+import {trpc} from 'app/_trpc/client'
 
 export const TipTeaser: React.FC<{tip: any}> = ({tip}) => {
   const {title, muxPlaybackId} = tip
   const thumbnail = `https://image.mux.com/${muxPlaybackId}/thumbnail.png?width=720&height=405&fit_mode=preserve`
   const router = useRouter()
+  const data = tip?.eggheadRailsLessonId
```
Creeland, line 16: I wonder if instead of data you could just have a boolean on each of these ternary results. Then you wouldn't need the second tipCompleted constant

```
? trpc.tips.loadTipProgress.useQuery({
        id: tip?.eggheadRailsLessonId,
      }).data?.tipCompleted
    : false
```

src/server/routers/tips.ts
```diff
@@ -157,4 +160,80 @@ export const tipsRouter = router({
...
+        {
+          method: 'POST',
+          headers,
+          body: JSON.stringify({
+            lesson_view: {
+              lesson_id: tipId,
+            },
+          }),
+        },
+      ).then((res) => res.json())
```
Creeland, line 192:
```suggestion
      // This is posting to /watch/download since it creates the proper amount of segments watched on the lesson view where /watch/manual_complete does not.
      let res = await fetch(
        `${process.env.NEXT_PUBLIC_AUTH_DOMAIN}/watch/download`,
        {
          method: 'POST',
          headers,
          body: JSON.stringify({
            lesson_view: {
              lesson_id: tipId,
            },
          }),
        },
      ).then((res) => res.json())
```

Creeland: I think leaving a comment here would be beneficial since it's an unusual situation with the API

Creeland (approved):

PR #1384, opened 2023-10-27
Fetch user course completions via prisma

jbranchaud:
@zacjones93 Here is a spike of another potential way to fetch course completions (playlist IDs) for the current user via Prisma. I'd expect this to be quite a bit faster than the graphql endpoint because it doesn't get a bunch of computed values like graphql does, it's just a single SQL query.

This is only a WIP because I'm having issues parsing the current user's ID from the `eh_user` cookie. I might need to find a different way to look up the current user's ID from tRPC.

spike


zacjones93:
The `eh_user` cookie value seems to alternate back and forth in a seemingly random fashion to an object that is extremely trimmed down:
```json
{
  "contact_id": "co_4d23a9d5bc981c26ccb0",
  "email": "zac+testtips@egghead.io",
  "is_pro": true,
  "is_instructor": false
}
``` 

where the full `eh_user` cookie is that we expect is: 

```json
{"name":null,"id":591000,"email":"zac+testtips@egghead.io","full_name":null,"discord_id":null,"contact_id":"co_4d23a9d5bc981c26ccb0","created_at":1698266275,"opted_out":false,"avatar_url":"//gravatar.com/avatar/54b870032ec33a3cbb4124dda3edeef2.png?s=128&d=mp","username":"member","user_url":"http://app.egghead.af:5000/api/v1/users/591000","watch_later_bookmarks_url":"http://app.egghead.af:5000/api/v1/playlists/watch-later-6ce4d5d3-be8a-45fa-95f6-8f7997ef58ff","favorites":["angular","typescript","react","next","javascript","html"],"is_pro":true,"can_comment":true,"timezone":"Central Time (US & Canada)","is_instructor":false,"is_verified":true,"current_course":{"url":"http://app.egghead.af:5000/api/v1/series/angular-dependency-injection-di-explained","slug":"angular-dependency-injection-di-explained","type":"course"},"roles":["user","account_owner","account_member","pro"],"accounts":[{"slug":"acc_4c31a12d","stripe_customer_id":"cus_OszpugBbPttHQ7","account_capacity":"individual","role":"owner","subscriptions":[{"stripe_subscription_id":"sub_1O5Dpn2nImeJXwdJbuf5y9o9"}]}],"providers":["magic"],"purchased":[]}
```

zacjones93 (approved):

src/server/routers/progress.ts
```diff
@@ -1,10 +1,37 @@
...
+              progressable_type: 'Playlist',
+              user_id: userId,
+              last_lesson_watched_at: {
+                not: null,
+              },
+              is_complete: true,
+            },
+          },
+        },
+      })
```
zacjones93, line 29: joins in prisma queries are 'implicit' meaning you set up relationships in the schema and then need to know that you can use those relationships in `where` and `include` clauses

PR #1385, opened 2023-10-27
get 5 tips and sort by updatedAt

zacjones93:
I wanted some more control over the order that tips show up on the home page, it makes sense that if we update a tip it gets bumped to the home page for people to view

g update

### DIFF
--- a/src/components/pages/home/sanity-sections.tsx
+++ b/src/components/pages/home/sanity-sections.tsx
@@ -168,7 +168,7 @@ const DynamicGridComponentWithTips = ({
   const publishedTips =
     data
       ?.find((tipGroup) => tipGroup.state === 'published')
-      ?.tips.slice(0, 4) ?? []
+      ?.tips.slice(0, 5) ?? []
 
   console.log({publishedTips})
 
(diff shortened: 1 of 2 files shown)


src/components/pages/home/sanity-sections.tsx
```diff
@@ -168,7 +168,7 @@ const DynamicGridComponentWithTips = ({
   const publishedTips =
     data
       ?.find((tipGroup) => tipGroup.state === 'published')
-      ?.tips.slice(0, 4) ?? []
+      ?.tips.slice(0, 5) ?? []
```
zacjones93, line 171: display 5 tips

src/lib/tips.ts
```diff
@@ -82,7 +82,7 @@ export type Tip = z.infer<typeof TipSchema>
 export const getAllTips = async (onlyPublished = true): Promise<Tip[]> => {
   const tips = await sanityClient.fetch(groq`*[_type == "tip" ${
     onlyPublished ? `&& state == "published"` : ''
-  }] | order(_createdAt asc) {
+  }] | order(_updatedAt asc) {
         _id,
```
zacjones93, line 86: order on `_updatedAt`

PR #1387, opened 2023-10-31
feat: groq for few tips when possible

jbranchaud:
Rather than sorting and slicing after the fact, we can tell groq to
fetch fewer tips that match the published state.

tip


src/lib/tips.ts
```diff
@@ -79,43 +78,65 @@ export const TipsSchema = z.array(TipSchema)
...
+  onlyPublished: z.boolean(),
+  limit: z.number().optional(),
+})
+
+type GetAllTipsOptions = z.infer<typeof GetAllTipsOptionsSchema>
+
+export const getAllTips = async (
+  options: GetAllTipsOptions,
+): Promise<Tip[]> => {
+  const {onlyPublished, limit} = GetAllTipsOptionsSchema.parse(options)
```
jbranchaud, line 91: parse options for this function: a `limit` and a `onlyPublished` filter

src/lib/tips.ts
```diff
@@ -79,43 +78,65 @@ export const TipsSchema = z.array(TipSchema)
...
+  // determine filtering conditions
+  const filterConditions = ['_type == "tip"']
+  if (onlyPublished) {
+    filterConditions.push('state == "published"')
+  }
+  const filter = filterConditions.join(' && ')
+
+  // determine slice clause limiting number of fetched results
+  const validLimit = limit && limit > 0
+  const sliceClause = validLimit ? `[0...${limit}]` : ''
```
jbranchaud, line 102: grab all tips or limit it to `0...<limit>` if the `limit` is provided

src/pages/tipz/index.tsx
```diff
@@ -6,7 +6,7 @@ import Balancer from 'react-wrap-balancer'
 import {Card, CardContent, CardHeader} from 'ui'
 
 export async function getStaticProps() {
-  const tips = await getAllTips()
+  const tips = await getAllTips({onlyPublished: true})
```
jbranchaud, line 9: updating this function call to only grab published tips which allows us to remove the filter below

src/server/routers/tips.ts
```diff
@@ -116,7 +116,13 @@ export const tipsRouter = router({
 
       throw new Error('Unauthorized')
     }),
+  published: baseProcedure
+    .input(z.object({limit: z.number().optional()}).optional())
+    .query(async ({input}) => {
+      const tips = await getAllTips({onlyPublished: true, limit: input?.limit})
 
+      return tips
+    }),
```
jbranchaud, line 125: leaving `tips.all` alone for the moment and adding this `tips.published` route for the Tips Sanity Section which only grabs a limited number of published tips, rather than all tips and then processing after the fact

src/server/routers/tips.ts
```diff
@@ -126,7 +132,7 @@ export const tipsRouter = router({
         .optional(),
     )
     .query(async ({input}) => {
-      const tips = await getAllTips(false)
+      const tips = await getAllTips({onlyPublished: false})
```
jbranchaud, line 135: no change here, just updating it from an argument to an options object

src/components/pages/home/sanity-sections.tsx
```diff
@@ -164,11 +164,9 @@ const DynamicGridComponentWithTips = ({
   location?: string
   completedCoursesIds?: number[]
 }) => {
-  const {data, status: tipsStatus} = trpc.tips.all.useQuery()
-  const publishedTips =
-    data
-      ?.find((tipGroup) => tipGroup.state === 'published')
-      ?.tips.slice(0, 5) ?? []
+  const {data, status: tipsStatus} = trpc.tips.published.useQuery({limit: 5})
```
jbranchaud, line 167: Replacing the filtering and slicing with a call to the `tips.published` route

src/lib/tips.ts
```diff
@@ -79,43 +78,65 @@ export const TipsSchema = z.array(TipSchema)
...
+
+  const query = groq`*[${filter}] | order(_updatedAt desc) {
```
jbranchaud, line 104: order by `updatedAt`, but in `desc` order so that the most recently updated tips show up at the top of the list.

src/lib/tips.ts
```diff
@@ -79,43 +78,65 @@ export const TipsSchema = z.array(TipSchema)
...
+    "tweetId":  resources[@._type == 'tweet'][0].tweetId,
+    'instructor': collaborators[@->.role == 'instructor'][0]->{
+      title,
+      'slug': person->slug.current,
+      'name': person->name,
+      'path': person->website,
+      'twitter': person->twitter,
+      'image': person->image.url
+    },
+  }${sliceClause}`
```
jbranchaud, line 137: Move the slice to the end because that seems to make results more consistent when also using the order by.

src/components/pages/home/sanity-sections.tsx
zacjones93: this api makes sense

zacjones93 (approved):
Makes way more sense to do filtering and sorting at the data level instead of over-fetching and slicing. LGTM

PR #1393, opened 2023-11-07
Track written tip reads and tip page views

Creeland:
This uses PostHog to track if people are scrolling down and viewing the written lessons. I'm also tracking page views to compare with the number of reads. 

I've also switched from using a class with static methods to a function that initializes the Posthog client. And, added a posthog provider to the app routes

### DIFF
--- a/src/app/providers.tsx
+++ b/src/app/providers.tsx
@@ -4,11 +4,11 @@ import {ViewerProvider} from 'context/viewer-context'
 import {CioProvider} from 'hooks/use-cio'
 import {ThemeProvider} from 'next-themes'
 import TrpcProvider from 'app/_trpc/Provider'
-import PosthogClient from 'lib/posthog-client'
+import PostHogClient from 'lib/posthog-client'
 import {PostHogProvider} from 'posthog-js/react'
 
 export function Providers({children}: {children: React.ReactNode}) {
-  const posthog = PosthogClient.init()
+  const posthog = PostHogClient
 
   return (
     <ThemeProvider attribute="class">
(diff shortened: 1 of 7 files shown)


src/lib/posthog-client.ts
```diff
@@ -1,27 +1,23 @@
...
-  static capture(event: string, params: any) {
-    posthog.capture(event, params)
-  }
+  return posthog
 }
 
-export default PosthogClient
+const PostHogClient = SetupPostHogClient()
+
+export default PostHogClient
```
zacjones93, line 23: should we be invoking the client set up function when we import it from other files?

src/pages/_app.tsx
```diff
@@ -23,7 +23,7 @@ import {ReactQueryDevtools} from '@tanstack/react-query-devtools'
 import TrpcProvider from 'app/_trpc/Provider'
 
 import {PostHogProvider} from 'posthog-js/react'
-import PosthogClient from 'lib/posthog-client'
+import PostHogClient from 'lib/posthog-client'
```
zacjones93, line 26: this could probably be `import posthog from 'lib/posthog-client'` since the file is exporting default variable

src/app/providers.tsx
```diff
@@ -4,13 +4,19 @@ import {ViewerProvider} from 'context/viewer-context'
...
+  const posthog = PostHogClient
+
   return (
     <ThemeProvider attribute="class">
       <ViewerProvider>
         <TrpcProvider>
-          <CioProvider>{children}</CioProvider>
+          <CioProvider>
+            <PostHogProvider client={posthog}>{children}</PostHogProvider>
+          </CioProvider>
```
zacjones93, line 19: We shouldn't need to import PostHogClient in the other files if we are wrapping the tree in a provider

```
import { usePostHog } from 'posthog-js/react'
```

`usePostHog` is the hook that would provide posthog to us

zacjones93 (changes requested):
If we are going to be wrapping the tree in a provider, the components that use it need to be using the hook instead of importing the client directly. `posthog/react` gives us the hook that will expose the client to those components

```
import { usePostHog } from 'posthog-js/react'
```

PR #1399, opened 2023-11-20
actually complete tips

zacjones93:
I commented out tip completion for dev work because 500s were clogging the network tab and console and committed that

g huge face palm

### DIFF
--- a/src/server/routers/tips.ts
+++ b/src/server/routers/tips.ts
@@ -190,6 +190,9 @@ export const tipsRouter = router({
       }),
     )
     .mutation(async ({input, ctx}) => {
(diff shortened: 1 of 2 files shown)


src/server/routers/tips.ts
```diff
@@ -190,6 +190,9 @@ export const tipsRouter = router({
       }),
     )
     .mutation(async ({input, ctx}) => {
+      if (process.env.NODE_ENV === 'development') {
+        return true
+      }
```
zacjones93, line 195: just return true if you're in development

nicollguarnizo (approved):

PR #1400, opened 2023-11-21
loads user from token instead of depending on cookie value

zacjones93:
Getting the instructor id from cookies sent with the request is being unreliable in production. This fetches the user from the current token instead

g disappearing cookie

### DIFF
--- a/src/server/routers/tips.ts
+++ b/src/server/routers/tips.ts
@@ -13,6 +13,7 @@ import {GraphQLClient} from 'graphql-request'
 import {inngest} from '@/inngest/inngest.server'
 import {TIP_VIDEO_UPLOADED_EVENT} from '@/inngest/events/tips'
 import groq from 'groq'
+import {loadCurrentUser} from '@/lib/users'
 
 export const tipsRouter = router({
   create: baseProcedure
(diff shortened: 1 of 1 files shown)


src/server/routers/tips.ts
```diff
@@ -44,16 +45,17 @@ export const tipsRouter = router({
             '1234567890abcdefghijklmnopqrstuvwxyz',
             5,
           )
-
-          const {instructor_id: instructorId} = JSON.parse(
-            ctx.req?.cookies.get('eh_user')?.value ?? "{instructor_id: ''}",
+          const {instructor_id: instructorId} = await loadCurrentUser(
+            ctx?.userToken,
           )
+          if (!instructorId) throw new Error('Instructor ID not found')
```
zacjones93, line 51: providing more useful errors where production is currently giving this error which is opaque:

```
Caught TRPC error: SyntaxError: Unexpected token i in JSON at position 1
    at JSON.parse (<anonymous>)
```

src/server/routers/tips.ts
```diff
@@ -44,16 +45,17 @@ export const tipsRouter = router({
...
 
           let sanityInstructor = await sanityWriteClient.fetch(groq`
           *[_type == 'collaborator' && eggheadInstructorId == '${instructorId}'][0] {
             _id,
-          }
-        `)
+          }`)
+          if (!sanityInstructor)
+            throw new Error('Sanity Instructor ID not found')
```
zacjones93, line 58: sanity instructor could also possibly not be set so this returns an error if that's the case

PR #1401, opened 2023-12-01
feat: tip publishing workflow

joelhooks:
flow

### DIFF
--- a/src/components/tips/tip-template.tsx
+++ b/src/components/tips/tip-template.tsx
@@ -22,14 +22,13 @@ import analytics from '@/utils/analytics'
 
 const TipTemplate = ({
   tip,
-  tips,
   coursesFromTag,
 }: {
   tip: Tip
-  tips: Tip[]
   coursesFromTag: any
 }) => {
   const markComplete = trpc.tips.markTipComplete.useMutation()
+  const {data: tips = []} = trpc.tips.published.useQuery()
 
   const {instructor, tags} = tip
 
@@ -38,7 +37,6 @@ const TipTemplate = ({
     if (tip?.eggheadRailsLessonId) {
       await markComplete.mutateAsync({tipId: tip?.eggheadRailsLessonId})
     }
-    console.log('video ended')
   }
 
   const module: any = {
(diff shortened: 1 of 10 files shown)


src/components/tips/tip-template.tsx
```diff
@@ -22,14 +22,13 @@ import analytics from '@/utils/analytics'
...
   tip,
-  tips,
   coursesFromTag,
 }: {
   tip: Tip
-  tips: Tip[]
   coursesFromTag: any
 }) => {
   const markComplete = trpc.tips.markTipComplete.useMutation()
+  const {data: tips = []} = trpc.tips.published.useQuery()
```
joelhooks, line 31: in this case we don't need to load that list of tips server side, might as well load it in on the client and let them see the page

I think you can do this with Suspense to and get the "streaming" response, but I'm not sure how to accomplish that yet but it would involve dropping `use client` and thinking about it differently.

src/components/tips/tip-uploader.tsx
```diff
@@ -102,11 +102,11 @@ const TipCreationForm: React.FunctionComponent<
       form.setValue('s3Url', '')
 
       await trpcUtils.tips.invalidate()
-      router.refresh()
+      router.push(`/tips/edit/${newTip?.slug}`)
```
joelhooks, line 105: take them to the new tip

src/inngest/functions/mux/mux-webhooks-handlers.ts
```diff
@@ -58,6 +58,10 @@ export const muxVideoAssetReady = inngest.createFunction(
           },
         ])
       })
+
+      await step.run('announce video processed', async () => {
+        return 'TODO: announce in slack'
+      })
```
joelhooks, line 64: needs to load the instructor for the channel id and send a message

src/server/routers/tips.ts
```diff
@@ -99,13 +99,19 @@ export const tipsRouter = router({
...
+            return parsedTip.data
+          } else {
+            throw new Error('Could not create tip')
+          }
         } else {
           throw new Error('Could not create video resource')
         }
       }
 
-      return new Response('Unauthorized', {status: 401})
```
joelhooks, line 108: TRPC doesn't return a `Response` in most cases it just returns data directly

PR #1402, opened 2023-12-13
feat: add cache revalidation to getTip and sanityQuery

Creeland:
I browsed the Sanity slack for caching issues and several people were having problems with Sanity and the app router. 

There were a couple of solutions: route based configuration, and params on individual fetch requests. I've gone ahead and added a `cache: 'no-cache`' param to the request in `getTip`. It seems to be working locally.

I wanted to do this instead of flipping `useCdn` which would affect everything.

https://nextjs.org/docs/app/building-your-application/data-fetching/fetching-caching-and-revalidating#individual-fetch-requests

# Update 1

Instead of disabling cache on fetch, @joelhooks suggested that we instead set the request to be revalidated. I've gone ahead and added this to the `sanityQuery` as well

# Update 2 (2023-12-14)

For `getTip` and `getAllTips`, I've switched to using global fetch in @joelhooks's `sanityQuery` function. I added the `next: { revalidate: 3600 }` configuration to the fetch in `sanityQuery`

### DIFF
--- a/src/utils/sanity.fetch.only.server.ts
+++ b/src/utils/sanity.fetch.only.server.ts
@@ -12,11 +12,14 @@ export async function sanityMutation(mutations: any[]) {
   ).then((response) => response.json())
 }
 
-export async function sanityQuery<T = any>(query: string): Promise<T> {
+export async function sanityQuery<T = any>(
+  query: string,
+  useCdn: boolean = true,
+): Promise<T> {
   return await fetch(
-    `https://${process.env.SANITY_STUDIO_PROJECT_ID}.api.sanity.io/v${
-      process.env.SANITY_STUDIO_API_VERSION
-    }/data/query/${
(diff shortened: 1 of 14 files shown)


src/utils/sanity.fetch.only.server.ts
```diff
@@ -24,6 +24,7 @@ export async function sanityQuery<T = any>(query: string): Promise<T> {
       headers: {
         Authorization: `Bearer ${process.env.SANITY_API_TOKEN}`,
       },
+      next: {revalidate: 3600},
```
joelhooks, line 27: this is an hour, which is a long time if you are expecting an edit to show up

Creeland: I'm not sure what time to set it as. Because in my mind, edits should show up immediately

joelhooks (approved):

PR #1409, opened 2024-02-21
Course page redirects if course not found in sanity

zacjones93:
When trying to load sanity data, we check if the course has a course image but assume the course is in sanity and add a default if not. 

When a course happens to not be represented in sanity this forces a redirect because we are trying to modify a null object.

Adding a check for the course object allows the course to still load

whoops

### DIFF
--- a/src/lib/courses.ts
+++ b/src/lib/courses.ts
@@ -124,7 +124,7 @@ export async function loadCourseMetadata(id: number, slug: string) {
 
   const course = await sanityClient.fetch(courseQuery, params)
 
-  if (!course?.square_cover_480_url) {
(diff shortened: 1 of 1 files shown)


src/lib/courses.ts
```diff
@@ -124,7 +124,7 @@ export async function loadCourseMetadata(id: number, slug: string) {
 
   const course = await sanityClient.fetch(courseQuery, params)
 
-  if (!course?.square_cover_480_url) {
+  if (course && !course?.square_cover_480_url) {
```
zacjones93, line 127: in the rare case a course is not found in sanity, this ensures we aren't attempting to modify a null value

joelhooks (approved):

PR #1411, opened 2024-02-22
enhance pricing widget with yearly percernt off

zacjones93:
During a design session with @vojtaholik, we thought it would make a positive impact if we display the % savings of switching to a yearly plan would give a learner. 

This adds a switch statement that displays `Save X% with yearly billing` when people are on quarterly or monthly options.

What we display on the yearly plan could use some zhuzh 

(my local dev set up has weird pricing)
monthly
quarterly
yearly

g savings

### DIFF
--- a/src/components/pricing/select-plan-new/index.tsx
+++ b/src/components/pricing/select-plan-new/index.tsx
@@ -207,6 +207,41 @@ const GetAccessButton: React.FunctionComponent<
   )
 }
 
+const PlanPercentageOff: React.FunctionComponent<
+  React.PropsWithChildren<{interval: string}>
+> = ({interval}) => {
+  switch (interval) {
+    case 'Yearly':
+      return (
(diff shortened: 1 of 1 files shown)


src/components/pricing/select-plan-new/index.tsx
```diff
@@ -207,6 +207,41 @@ const GetAccessButton: React.FunctionComponent<
...
+          with yearly billing
+        </div>
+      )
+    case 'Monthly':
+      return (
+        <div className="max-w-2xl pt-4 text-sm font-light leading-tight text-gray-700 dark:text-gray-200">
+          Save{' '}
+          <strong className="dark:bg-amber-400 dark:text-black bg-blue-600 text-white px-px font-semibold">
+            50%
+          </strong>{' '}
```
zacjones93, line 236: these percentages are hardcoded, they could be dynamically generated as well

DrShpongle (approved):

PR #1413, opened 2024-02-26
refresh navbar to surface useful links top level

zacjones93:
## WIP 

https://www.loom.com/share/ec738bc47a4a4b189c8e937b42dc2555?sid=8f10607d-3d01-4984-98bb-f47b66620970


src/components/app/header/index.tsx
```diff
@@ -1,13 +1,13 @@
...
 import {FunctionComponent} from 'react'
 import Link from '../../link'
+import NextLink from 'next/link'
 import Image from 'next/legacy/image'
 import Eggo from '@/components/icons/eggo'
 import {useViewer} from '@/context/viewer-context'
 import {track} from '@/utils/analytics'
 import {isEmpty} from 'lodash'
 import FeedbackInput from '@/components/feedback-input'
-import useBreakpoint from '@/utils/breakpoints'
```
vojtaholik, line 10: I went ahead and removed all the js breakpoint stuff and replaced it with with css which lets us avoid `isMounted` check and hydration mismatches.

src/components/app/header/index.tsx
```diff
@@ -249,167 +369,103 @@ const Header: FunctionComponent<React.PropsWithChildren<unknown>> = () => {
...
+            <div className="hidden items-center h-full lg:flex">
+              {navLinks.map((link) => {
+                return <NavLink key={link.name} {...link} />
+              })}
             </div>
           </div>
-        )}
-        {isOpen && <MobileNavigation />}
+          <div className="flex items-center h-full">
+            {loading ? null : (
```
vojtaholik, line 382: all right-hand menu items are awaited during loading to avoid layout shift.

src/components/app/header/index.tsx
```diff
@@ -17,52 +17,319 @@ import SearchBar from './search-bar'
...
+          props.onClick()
+          // do stuff
+        },
+      })
+    )
+
+  return <LinkOrButton />
+}
+
+const NavDropdown: React.FC<
```
vojtaholik, line 217: refactored this into generic dropdown component. I think we should be able to apply logic to transform it into an accordion component on mobile.

src/components/app/header/index.tsx
```diff
@@ -1,82 +1,324 @@
+'use client'
```
zacjones93, line 1: we could get more fine-grain where we `use client` by loading `viewer` data from the layout RSC but also using this component in the pages router means we need to `useViewer` hook top level to do viewer related checks

vojtaholik (approved):

PR #1415, opened 2024-03-01
add default comment and display 'member sign up' CTA to unlock comments

zacjones93:
The logic for showing the member sign up CTA for non paying users was busted with two checks to see if they could comment. 

Fixed the logic (removed nested ternary) and adjusted styles. Also added the comment guidelines as a button because those are still relevant. I also added the comment guidelines as the default first comment if no user comments exist. This gives people a chance to read them over before they comment which I think is nice.

g comment

## non paying / signed out user on lesson with no comments
non paying / signed out user on lesson with no comments
mobile - non paying / signed out user on lesson with no comments

## paying member on lesson with comments
paying member on lesson with comments
mobile - paying member on lesson with comments

## guideline dialog
guideline dialog
mobile guideline dialog


vojtaholik (approved):
🎇

PR #1416, opened 2024-03-01
fix SEO structured data errors

zacjones93:
Saw some errors in the Google Search Console that are quick fixes

## `Parsing error: Missing ',' or '}'` 
any lesson or course that has `""` in the description is erroring out because we don't escape the quote mark. Swapped the double quote for single quote to fix.

 ## `Course info`
Courses have 2 main errors which looks like `next-seo` doesn't support this structured data well. 
- `Missing field "hasCourseInstance"`
- `Missing field "offers"`
`hasCourseInstance` is looking for a place the course is being hosted, looks like 'online' is an option. `offers` is looking for a price which I'm not sure how to gauge for our courses. 

`next-seo` is throwing errors as it doesn't expect this data so I'm opting to remove the structured data for now to resolve errors.

g fixed

### DIFF
--- a/src/components/layouts/collection-page-layout.tsx
+++ b/src/components/layouts/collection-page-layout.tsx
@@ -403,7 +403,9 @@ const CollectionPageLayout: React.FunctionComponent<
   return (
     <>
       <NextSeo
-        description={truncate(removeMarkdown(description), {length: 155})}
+        description={truncate(removeMarkdown(description.replace(/"/g, "'")), {
+          length: 155,
+        })}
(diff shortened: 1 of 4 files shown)


src/components/layouts/collection-page-layout.tsx
```diff
@@ -403,7 +403,9 @@ const CollectionPageLayout: React.FunctionComponent<
   return (
     <>
       <NextSeo
-        description={truncate(removeMarkdown(description), {length: 155})}
+        description={truncate(removeMarkdown(description.replace(/"/g, "'")), {
```
zacjones93, line 406: replacing `"` with `'`

PR #1419, opened 2024-03-06
fix: add contentUrl and duration to structured video data

zacjones93:
the `contentUrl` for videos is getting set to some random string (uuid?) on lesson pages which will 404. This sets content URL to the `hls_url` which is an [accepted format for google](https://developers.google.com/search/docs/appearance/video#supported-video-encodings)

[content url is recommended](https://developers.google.com/search/docs/appearance/structured-data/video#structured-data-type-definitions)
image

image

g index me

### DIFF
--- a/src/components/pages/lessons/lesson/index.tsx
+++ b/src/components/pages/lessons/lesson/index.tsx
@@ -519,6 +519,8 @@ const Lesson: React.FC<React.PropsWithChildren<LessonProps>> = ({
         description={truncate(removeMarkdown(description.replace(/"/g, "'")), {
           length: 155,
         })}
+        contentUrl={lesson?.hls_url}
+        duration={lesson?.duration}
         uploadDate={lesson?.created_at}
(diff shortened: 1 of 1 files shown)


src/components/pages/lessons/lesson/index.tsx
```diff
@@ -519,6 +519,8 @@ const Lesson: React.FC<React.PropsWithChildren<LessonProps>> = ({
         description={truncate(removeMarkdown(description.replace(/"/g, "'")), {
           length: 155,
         })}
+        contentUrl={lesson?.hls_url}
+        duration={lesson?.duration}
```
zacjones93, line 523: added duration as well

joelhooks (approved):

src/components/pages/lessons/lesson/index.tsx
vojtaholik: @zacjones93 
image

PR #1420, opened 2024-03-06
remove extra slash in url

zacjones93:
g extra


src/components/pages/lessons/lesson/index.tsx
```diff
@@ -527,7 +527,7 @@ const Lesson: React.FC<React.PropsWithChildren<LessonProps>> = ({
       <SocialProfileJsonLd
         type="Person"
         name={instructor.full_name}
-        url={`https://egghead.io/${instructorPagePath}`}
+        url={`https://egghead.io${instructorPagePath}`}
```
zacjones93, line 530: `instructorPagePath` is `/q/resources-by-${get(instructor, 'slug', '#')}`

PR #1421, opened 2024-03-08
update tips index to list

zacjones93:
The tips index page doesn't tell you much about what you're actually learning as each tip card is mostly the thumbnail of the tip. This updates to give more information per tip with instructor, tags, and a little bit of the description (if present which we should fix if not).

 g makeoverj

## After

tips desktop
tips mobile

## before
tips before

### DIFF
--- a/src/components/tips/tip-card.tsx
+++ b/src/components/tips/tip-card.tsx
@@ -227,10 +227,12 @@ const TagList: FunctionComponent<
               <Link
                 href={`/q/${tag.name}`}
                 onClick={() => {
-                  // track(`clicked view topic`, {
-                  //   course: courseSlug,
-                  //   topic: tag.name,
-                  // })
+                  analytics.events.activityInternalLinkClick(
+                    'tag',
+                    'tips index',
+                    tag.name,
+                    tag.slug,
+                  )
                 }}
                 className="inline-flex items-center hover:underline"
               >


src/components/tips/tip-card.tsx
```diff
@@ -1,73 +1,78 @@
...
-          className="group  flex items-center justify-center"
-        >
+    <Link
+      className={twMerge(
+        cx(
+          'group relative flex flex-row items-center overflow-hidden rounded border border-transparent bg-white hover:bg-gray-100 dark:bg-gray-400/5 hover:dark:bg-gray-800 transition duration-300 ease-in-out',
+          {
+            'dark:bg-gray-900 bg-gray-50': i % 2 === 0,
+          },
+        ),
```
Creeland, line 39: I like how you've alternated the colors using this logic with the index

src/components/tips/tip-card.tsx
```diff
@@ -176,4 +206,55 @@ export const TipTeaser: React.FC<{tip: Tip}> = ({tip}) => {
...
+              className="inline-flex items-center mr-4 mt-0 dark:text-indigo-100 text-gray-700 lg:text-sm text-[0.75rem] opacity-80 leading-none"
+            >
+              <Link
+                href={`/q/${tag.name}`}
+                onClick={() => {
+                  // track(`clicked view topic`, {
+                  //   course: courseSlug,
+                  //   topic: tag.name,
+                  // })
+                }}
```
Creeland, line 234: Not tracking this?

PR #1424, opened 2024-03-14
clean up tip index

zacjones93:
implements feedback from vojta on tip index style

added in a skeleton loader though index is so fast even on 3g you don't see it for long. It pulses like it's loading which is nice
tips skeleton

upgraded tailwind for the `text-balance` property which is nice to not need the React component `Balancer` to get the same effect:
tips

g squeaky clean


src/app/(content)/tips/_components/tips-lists.tsx
```diff
@@ -0,0 +1,18 @@
...
+  return (
+    <>
+      {publishedTips
+        .filter(({state}) => state === 'published')
+        .map((tip, i) => {
+          return <TipCard tip={tip} key={tip.slug} i={i} />
+        })}
+    </>
+  )
+}
```
zacjones93, line 16: Load tips in it's own component

src/app/(content)/tips/page.tsx
```diff
@@ -32,15 +31,66 @@ const TipsIndex: React.FC<any> = async () => {
...
       <main className="max-w-screen-lg mx-auto relative z-10 flex flex-col items-center justify-center pb-16">
         <div className="flex w-full flex-col gap-0 sm:gap-3">
-          {publishedTips
-            .filter(({state}) => state === 'published')
-            .map((tip, i) => {
-              return <TipCard tip={tip} key={tip.slug} i={i} />
-            })}
+          <Suspense fallback={<Skeleton />}>
+            <TipList />
+          </Suspense>
```
zacjones93, line 36: wrap list in suspense in tip index page

src/components/tips/tip-card.tsx
```diff
@@ -29,24 +27,15 @@ const TipCard: React.FC<{tip: Tip; i: number}> = ({tip, i}) => {
   const tipTopic = tip?.tags ? tip.tags[0]?.name : ''
 
   return (
-    <Link
+    <article
       className={twMerge(
         cx(
-          'group relative flex flex-row items-center overflow-hidden rounded border border-transparent bg-white hover:bg-gray-100 dark:bg-gray-400/5 hover:dark:bg-gray-800 transition duration-300 ease-in-out',
+          'group relative flex flex-row items-center overflow-hidden rounded border border-transparent bg-white dark:bg-gray-400/5 transition duration-300 ease-in-out px-4 py-2 min-h-[160px] ',
```
zacjones93, line 33: min height here is nice because ~160px is around the height of the average tip with title being one row and description being two

src/components/tips/tip-card.tsx
```diff
@@ -29,24 +27,15 @@ const TipCard: React.FC<{tip: Tip; i: number}> = ({tip, i}) => {
   const tipTopic = tip?.tags ? tip.tags[0]?.name : ''
 
   return (
-    <Link
+    <article
```
zacjones93, line 30: don't wrap the whole item in a link because you get nested a tags which is a no no

src/components/tips/tip-card.tsx
```diff
@@ -55,26 +44,42 @@ const TipCard: React.FC<{tip: Tip; i: number}> = ({tip, i}) => {
...
-            <Image
-              src={thumbnail}
-              alt=""
-              objectFit="cover"
-              layout="fill"
-              aria-hidden="true"
-              className="brightness-90 transition duration-300 group-hover:brightness-75 dark:brightness-50"
-            />
+            <Link
+              tabIndex={-1}
```
zacjones93, line 48: tabIndex -1 has keyboard skip the image as a link and go straight to the next title link which is nice

src/components/tips/tip-card.tsx
```diff
@@ -100,108 +119,40 @@ const TipCard: React.FC<{tip: Tip; i: number}> = ({tip, i}) => {
                 />
               </div>
             )}
-            <span className="text-left dark:text-indigo-100 text-gray-700 lg:text-sm text-[0.75rem] opacity-80 leading-none">
```
zacjones93, line 103: this text color looked off, replaced with a consistent gray

src/components/tips/tip-card.tsx
```diff
@@ -100,108 +119,40 @@ const TipCard: React.FC<{tip: Tip; i: number}> = ({tip, i}) => {
...
-          <TagList tags={tip.tags} />
         </div>
         {tip?.description && (
-          <ReactMarkdown className="prose dark:prose-dark dark:prose-dark-sm dark:prose-a:text-blue-300 prose-a:text-blue-500 prose-sm line-clamp-2 max-w-[50ch] pt-2">
+          <ReactMarkdown
+            className="prose dark:prose-dark dark:prose-dark-sm dark:prose-a:text-blue-300 prose-a:text-blue-500 prose-sm line-clamp-2 pt-2"
+            components={{
+              a: (props) => <span>{props.children}</span>,
+              p: (props) => <p className="text-balance">{props.children}</p>,
+            }}
```
zacjones93, line 150: didn't know about the components prop for `ReactMarkdown` but this gives you control of how different elements are presented when converted to html. Specifically.. we don't want links in the markdown to actually render because it was causing hydration errors also `text-balance` on description is nice.

src/components/tips/tip-card.tsx
```diff
@@ -100,108 +119,40 @@ const TipCard: React.FC<{tip: Tip; i: number}> = ({tip, i}) => {
...
+          >
             {tip.description}
           </ReactMarkdown>
         )}
       </CardContent>
-    </Link>
-  )
-}
-
-export const TipTeaser: React.FC<{tip: Tip}> = ({tip}) => {
```
zacjones93, line 127: This was an unused component that we also have defined in its own file. Artifact of the initial transfer of tips into egghead

nicollguarnizo (approved):

PR #1426, opened 2024-03-19
remove post purchase survey

zacjones93:
Removing this survey allows me to cancel my tally subscription, not seeing value in the subscription right now.

The last time a user filled this survey out was Feburary 22nd, with it at the bottom of the sign up page it can be removed and we rethink where to survey. 

shutting down


src/components/pages/confirm/membership/index.tsx
```diff
@@ -7,7 +7,6 @@ import Spinner from '@/components/spinner'
 import {IconTwitter} from '@/components/share'
 import usePurchaseAndPlay from '@/hooks/use-purchase-and-play'
 import {Topic} from '@/types'
-import PostPurchase from '@/components/survey/tally/post-purchase'
```
jbranchaud, line 10: are you keeping this file around for now or can that get deleted too?

jbranchaud: not too hard to find and restore it from git history either

PR #1427, opened 2024-03-22
Creeland/egg 233 filters on egghead search page not working

Creeland:
This fixes [EGG-233](https://linear.app/skillrecordings/issue/EGG-233/filters-on-egghead-search-page-not-working) and more. 

Search should behave completely as expected now:

- Flip filters on and off
- Get the expected Curated Page
- Curated pages fall back to graphql data if sanity data isn't present
- Curated pages are displayed and removed dynamically when filters are used. You no longer need to refresh the page.
- Use both query params and the search box

I merged together use-selected-topic and use-loaded-topic into one and got things in sync. I made sure that the effect updated on change to the search state as well. 

I also implemented an optimistic update to the TagList items. 

skullandbones


src/hooks/use-load-topic-data.ts
```diff
@@ -1,26 +1,74 @@
...
 import {sanityClient} from '@/utils/sanity-client'
+import {topicExtractor} from '@/utils/search/topic-extractor'
+import {first} from 'lodash'
+import {loadTag} from '@/lib/tags'
+
+function useLoadTopicData(
+  initialTopicGraphqlData: any,
+  initialTopicSanityData: any,
+  searchState: any,
+) {
```
Creeland, line 12: I merged the two loader hooks into this one so that I could have more control of the timing

src/hooks/use-load-topic-data.ts
```diff
@@ -1,26 +1,74 @@
...
+          topicQuery,
+          {
+            slug: newTopic,
+          },
+          {signal},
+        )
+
+        if (sanityData) {
+          setTopicSanityData(sanityData)
+          setTopicGraphqlData({name: sanityData.slug})
```
Creeland, line 42: A topic with a name is always expected. In the Search component we are saving data from our GraphQL endpoint under `topic` and the sanity data as `topicData`. A next step could be to adjust the names to be more clear

src/hooks/use-load-topic-data.ts
```diff
@@ -1,26 +1,74 @@
...
+      } finally {
+        setLoading(false)
+      }
+    }
+
+    if (newTopic) {
+      getData()
+    } else {
+      setTopicSanityData(null)
+      setTopicGraphqlData({name: ''})
```
Creeland, line 63: A topic is always expected. A next step could be to handle null in the template. I don't like having to set this empty string.

I wanted to stay within the scope of fixing the search page bug in this PR

src/components/search/refinement-list.tsx
```diff
@@ -30,6 +30,7 @@ const TagItem: FunctionComponent<
         href={createURL(item.value)}
         onClick={(event) => {
           event.preventDefault()
+          item.isRefined = !item.isRefined
```
Creeland, line 33: This was done because algolia's `refine` function was correctly refining the list, but it wasn't flipping the `isRefined` boolean on these items.

zacjones93 (approved):

PR #1431, opened 2024-03-29
add slack chat notification when tip is created

zacjones93:
This depends on [rails PR](https://github.com/skillrecordings/egghead-rails/pull/5241) adding the `slack_group_id` to the graphql api.

In the sanity tip created webhook we add an inngest function call to send a slack message via fetch

image

notified

### DIFF
--- a/src/app/api/webhooks/tips/sanity/route.ts
+++ b/src/app/api/webhooks/tips/sanity/route.ts
@@ -44,39 +44,36 @@ export async function POST(req: NextRequest) {
         name: SEND_SLACK_MESSAGE_EVENT,
         data: {
           instructorId,
-          messageType: 'resource-created',
           message: `_Created egghead tip_: *<https://egghead.io/tips/${slug.current}|${title}>*\n\n`,
         },
       })
 
       // create a lesson in rails
       // patch lesson resources array with ref using the video resource id
-      // try {
-      //   const body = new URLSearchParams({
-      //     'lesson[instructor_id]': instructorId,
-      //     'lesson[title]': title,
-      //   })
+      try {
+        const body = new URLSearchParams({
+          'lesson[instructor_id]': instructorId,
+          'lesson[title]': title,
+        })
 
-      //   const lesson = await eggAxios.post('/api/v1/lessons', body)
+        const lesson = await eggAxios.post('/api/v1/lessons', body)
 
-      //   await sanityClient
-      //     .patch(_id)
-      //     .set({
-      //       eggheadRailsLessonId: lesson.data.id,
-      //       eggheadRailsCreatedAt: new Date().toISOString(),
-      //     })
-      //     .commit()
+        await sanityClient
+          .patch(_id)
+          .set({
+            eggheadRailsLessonId: lesson.data.id,
+            eggheadRailsCreatedAt: new Date().toISOString(),
+          })
+          .commit()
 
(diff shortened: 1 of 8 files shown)


src/inngest/functions/index-lessons.ts
```diff
@@ -2,13 +2,15 @@ import {inngest} from '@/inngest/inngest.server'
...
+const base64EncodedServiceAccount =
+  process.env.GOOGLE_SERVICE_ACCOUNT_ENCODED ?? ('' as string)
 const decodedServiceAccount = Buffer.from(
   base64EncodedServiceAccount,
   'base64',
 ).toString('utf-8')
-const credentials = JSON.parse(decodedServiceAccount)
+const credentials = decodedServiceAccount
+  ? JSON.parse(decodedServiceAccount)
+  : ''
```
zacjones93, line 13: build was throwing type errors in dev because I don't have t`GOOGLE_SERVICE_ACCOUNT_ENCODED` defined so this makes it not break

src/inngest/functions/index-lessons.ts
```diff
@@ -21,6 +23,8 @@ export const indexLessonsForever = inngest.createFunction(
   {id: `index-all-lessons-forever`, name: 'Index 200 Lessons'},
   {event: INDEX_LESSONS_FOREVER},
   async ({event, step}) => {
+    if (process.env.NODE_ENV === 'development') return 'no index in dev'
```
zacjones93, line 26: also skipping the work when in dev

src/inngest/functions/send-slack-message.ts
```diff
@@ -0,0 +1,94 @@
...
+    const {instructor} = await step.run(
+      'Get Rails Instructor object',
+      async () => {
+        try {
+          return await graphQLClient.request(query, {slug: instructorId})
+        } catch (error) {
+          console.error('Error fetching instructor', error)
+        }
+      },
+    )
```
zacjones93, line 36: get instructor information

src/inngest/functions/send-slack-message.ts
```diff
@@ -0,0 +1,94 @@
...
+                    {
+                      type: 'plain_text',
+                      text: instructor.full_name,
+                      emoji: true,
+                    },
+                  ],
+                },
+              ],
+            }),
+          },
```
zacjones93, line 81: send slack message. @nicollguarnizo and I were having the worst time with the slack client and it's bot scopes so I opted for using fetch with the `SLACK_ADMIN_API_KEY` that we use in rails as well

src/app/api/webhooks/tips/sanity/route.ts
```diff
@@ -38,6 +40,14 @@ export async function POST(req: NextRequest) {
       const {_id, title, slug, instructorId} = sanityRequestBody.data
       console.info('processing Sanity webhook: Lesson created', _id)
 
+      await inngest.send({
+        name: SEND_SLACK_MESSAGE_EVENT,
+        data: {
+          instructorId,
+          message: `_Created egghead tip_: *<https://egghead.io/tips/${slug.current}|${title}>*\n\n`,
```
zacjones93, line 47: passing in message here so we can re-use this event for other instances (like egghead-feedback)

nicollguarnizo (approved):

PR #1435, opened 2024-04-03
add lifetime member case on /membership

zacjones93:
/membership page needs to check for `lifetime_subscriber` role being applied to accounts

image

g together forever


jbranchaud (approved):
looks good, also perfect gif

PR #1437, opened 2024-04-08
Rate limit search page to 60 requests a minute

zacjones93:
I picked 60 requests a minute but not sure if that should be lowered at all.

This follows vercel's [rate limiting example](https://github.com/vercel/next.js/pull/19509/files). Because the use-case is simple, I opted for this over introducing [upstash ratelimit](https://github.com/upstash/ratelimit)

g stop

### DIFF
--- a/src/pages/q/[[...all]].tsx
+++ b/src/pages/q/[[...all]].tsx
@@ -1,5 +1,6 @@
 import * as React from 'react'
 import {useRouter} from 'next/router'
+import Image from 'next/image'
 import algoliasearchLite from 'algoliasearch/lite'
 import Search from '@/components/search'
 import {NextSeo} from 'next-seo'
@@ -27,6 +28,7 @@ import {
   SearchBox,
 } from 'react-instantsearch'
 import {renderToString} from 'react-dom/server'
+import rateLimit from '@/utils/rate-limit'
 
 const tracer = getTracer('search-page')
 
@@ -58,6 +60,7 @@ const getInstructorSlugFromInstructorList = (instructors: string[]) => {
 }
 
 type SearchIndexProps = {
+  error: string
   initialSearchState: any
   resultsState: any
   pageTitle: string
@@ -68,6 +71,7 @@ type SearchIndexProps = {
 }
 
 const SearchIndex: any = ({
+  error,
   initialSearchState,
   resultsState,
   pageTitle,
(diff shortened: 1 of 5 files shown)


src/pages/q/[[...all]].tsx
```diff
@@ -88,6 +92,20 @@ const SearchIndex: any = ({
...
+        <Image
+          src="https://res.cloudinary.com/dg3gyk0gu/image/upload/v1659039546/eggodex/basic_eggo.png"
+          alt="egghead search error"
+          width={200}
+          height={200}
+        />
+        <p className="prose dark:prose-dark text-xl">{error}</p>
+      </div>
+    )
+  }
```
Comment (zacjones93) [inline src/pages/q/[[...all]].tsx:107]: if we return an error from `getServerSideProps` we display it. Currently just returning `'Rate limit exceeded, please try again in 60 seconds.'`

along with eggo eggo says stop

src/pages/q/[[...all]].tsx
```diff
@@ -175,11 +193,27 @@ function BrandPage({serverState}: any) {
...
+  uniqueTokenPerInterval: 500,
+})
+
 export const getServerSideProps: GetServerSideProps = async function ({
   req,
   query,
   res,
 }) {
+  try {
+    await limiter.check(res, 60, 'SEARCH_CACHE_TOKEN') // 60 requests per minute
```
Comment (zacjones93) [inline src/pages/q/[[...all]].tsx:207]: limits a user to 60 searches a minute. Unsure if we should drop this

src/pages/q/[[...all]].tsx
```diff
@@ -175,11 +193,27 @@ function BrandPage({serverState}: any) {
...
+  } catch (error) {
+    res.setHeader('Retry-After', 60)
+    res.statusCode = 429
+
+    return {
+      props: {error: 'Rate limit exceeded, please try again in 60 seconds.'},
+    }
```
Comment (zacjones93) [inline src/pages/q/[[...all]].tsx:214]: if there is an error, we set status code to 429 and `Retry-After` header to 60 seconds which is the interval we have set.

src/utils/search/topic-extractor.ts
```diff
@@ -2,8 +2,8 @@ import {QueryReturnType} from '@/lib/search-url-builder'
 import {first, isEmpty} from 'lodash'
 
 export const topicExtractor = (searchState: QueryReturnType) => {
-  if (!isEmpty(searchState.refinementList?._tags)) {
-    return searchState.refinementList?._tags
+  if (!isEmpty(searchState?.refinementList?._tags)) {
+    return searchState?.refinementList?._tags
```
zacjones93, line 6: sometimes we don't have `searchState` if we are returning an error so this does a safe look up

joelhooks (approved):

PR #1447, opened 2024-05-15
migrate to pnpm away from yarn

zacjones93:
Yarn v1 started giving build errors so I previously upgraded to Yarn v4 in #1443. All our other products use pnpm so this migrates us over 

An interesting note I ran into is that we were using a few packages that were not explicitly installed (e.g. `@reach/auto-id`, `@skillrecordings/time`) where it looks like pnpm is much stricter about that than yarn was

gif


src/components/upload/video-uploader.tsx
```diff
@@ -32,7 +32,7 @@ const VideoUploader = ({
         // filename with no extension
         const filename = fullFilename.replace(/\.[^/.]+$/, '')
         // remove stuff s3 hates
-        const scrubbed = `${filename}-${uuid.generate()}`
+        const scrubbed = `${filename}-${nanoid()}`
```
jbranchaud, line 35: Looking into shortid, I think the default was a 7 character identifier, so for consistency:

```suggestion
        const scrubbed = `${filename}-${nanoid(7)}`
```

src/module-builder/get-unique-filename.ts
```diff
@@ -1,11 +1,11 @@
-import uuid from 'shortid'
+import {nanoid} from 'nanoid'
 import fileExtension from 'file-extension'
 
 export const getUniqueFilename = (fullFilename: string) => {
   // filename with no extension
   const filename = fullFilename.replace(/\.[^/.]+$/, '')
   // remove stuff s3 hates
-  const scrubbed = `${filename}-${uuid.generate()}`
+  const scrubbed = `${filename}-${nanoid()}`
```
jbranchaud, line 8: Looking into shortid, I think the default was a 7 character identifier, so for consistency:

```suggestion
  const scrubbed = `${filename}-${nanoid(7)}`
```

jbranchaud (approved):

PR #1448, opened 2024-05-16
feat: skip rate-limit check in dev

jbranchaud:
skip


zacjones93 (approved):
This makes sense to me

PR #1451, opened 2024-05-29
send message to slack on cio api error

zacjones93:
When we get an error from CIO, send a slack message so we can resolve it

Not sure what a useful link would be here but started with a link to the person in CIO

image

g alert


src/inngest/functions/send-slack-message.ts
```diff
@@ -59,6 +59,19 @@ export const sendSlackMessage = inngest.createFunction(
...
+          webhookUrl: channel,
+          username: 'Eggo',
+          text: message,
+          attachments,
+        })
+      } catch (error) {
+        console.error('Error sending slack message', error)
+      }
+    } else if (messageType === 'error') {
+      channel = process.env.SLACK_EGGHEAD_MONITORING_CHANNEL_WEBHOOK ?? ''
```
zacjones93, line 73: setting up a webhook for a channel is very straight forward, figuring out bot permissions has been a maze of ambiguity. I figured with something like this error reporting we will only need this one channel anyway.

Creeland (approved):

PR #1454, opened 2024-06-05
Fix bookmark button

zacjones93:
We've had a few reports of users not being able to bookmark courses because they are asked to sign in again. Looks like `getAccessTokenFromCookie` returns false in a seemingly random way (sometimes it works and sometimes doesn't). We have the access token available to us so passing that in as an additional param makes this query more reliable 

RSS button is also not working as expected but seems to be an issue on the rails side not returning that url properly. That needs more investigation

image

This also fixes the login form if a user is truly not logged in:

## Before
image

## After
image

gif


src/pages/courses/[course]/index.tsx
```diff
@@ -122,6 +119,13 @@ export const getServerSideProps: GetServerSideProps = async ({
   setupHttpTracing({name: getServerSideProps.name, tracer, req, res})
   try {
     const course = params && (await loadPlaylist(params.course as string))
+    const authedCourseAttributes =
+      params &&
+      (await loadAuthedPlaylistForUser(
+        params.course as string,
+        req.cookies[ACCESS_TOKEN_KEY],
+      ))
```
Comment (zacjones93) [inline src/pages/courses/[course]/index.tsx:127]: passes in access token

src/lib/playlists.ts
```diff
@@ -115,14 +118,13 @@ export async function loadAuthedPlaylistForUser(slug: string) {
     }
   `
   const token = getAccessTokenFromCookie()
-  const graphQLClient = getGraphQLClient(token)
+  const graphQLClient = getGraphQLClient(accessToken ?? token)
```
zacjones93, line 121: use passed in access token if available

PR #1458, opened 2024-06-26
feat: switch typesense sorting to presets

zacjones93:
Since migrating to TypeSense, whenever a user inputs a search or clicks a filter the url is populated by the sorting that we need to do for the content that makes sense.

e.g. `https://egghead.io/q?q=react&sortBy=content_production/sort/_eval([%20(type:playlist):4,%20(type:lesson):3,%20(type:podcast):2],%20(type:talk):1):desc,search_research_sort:asc,rank:asc`

The first sorting function is particularly important as it rank sorts playlists to the top before other types of content. 

The other way to do this is with presets but @Creeland and I were having a rough time actually setting a preset apart from initial load. Turns out that the TypeSense adapter has an `updateConfiguration` function that we can call when we want to switch to a different preset as recommended [here](https://github.com/typesense/typesense-instantsearch-adapter/issues/195). 

I'm updating configuration in a select and calling instantsearches `refresh` to trigger another search. 

gif

### DIFF
--- a/src/components/search/index.tsx
+++ b/src/components/search/index.tsx
@@ -10,7 +10,7 @@ import {
   Configure,
   InstantSearch,
   ClearRefinements,
-  SortBy,
+  useInstantSearch,
 } from 'react-instantsearch'
 import {get, isEmpty} from 'lodash'
 import {useToggle} from 'react-use'
@@ -28,7 +28,11 @@ import cx from 'classnames'
 import NewCuratedTopicPage from './curated/[slug]'
 import Link from 'next/link'
 import analytics from '@/utils/analytics'
-import {SORT_PRESETS} from '@/utils/typesense'
+import {
+  typsenseAdapterConfig,
+  TYPESENSE_COLLECTION_NAME,
+} from '@/utils/typesense'
+import {typesenseAdapter} from '@/pages/q/[[...all]]'
 
 type SearchProps = {
   searchClient?: any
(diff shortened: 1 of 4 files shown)


src/components/search/index.tsx
```diff
@@ -157,6 +161,53 @@ const Search: FunctionComponent<React.PropsWithChildren<SearchProps>> = ({
...
+          typesenseAdapter.updateConfiguration({
+            ...typsenseAdapterConfig,
+            additionalSearchParameters: {
+              query_by: 'title,description,_tags,instructor_name,contributors',
+              preset: e.target.value,
+            },
+          })
+
+          refresh()
+        }}
```
zacjones93, line 181: where the magic happens. updates configuration and triggers a search

src/pages/q/[[...all]].tsx
```diff
@@ -36,7 +36,9 @@ const tracer = getTracer('search-page')
 
 const createURL = (state: any) => `?${qs.stringify(state)}`
 
-const searchClient = typesenseInstantsearchAdapter().searchClient
+export const typesenseAdapter = typesenseInstantsearchAdapter()
+
+const searchClient = typesenseAdapter.searchClient
```
Comment (zacjones93) [inline src/pages/q/[[...all]].tsx:41]: need access to the `typesenseAdapter` to update configuration

src/utils/typesense.ts
```diff
@@ -1,34 +1,29 @@
 import TypesenseInstantSearchAdapter from 'typesense-instantsearch-adapter'
 
+export const typsenseAdapterConfig = {
```
zacjones93, line 3: stuffed the config in an object to spread when we update presets

nicollguarnizo (approved):

PR #1460, opened 2024-07-09
actually use the index for initial search state

zacjones93:
gif

### DIFF
--- a/src/components/search/index.tsx
+++ b/src/components/search/index.tsx
@@ -299,8 +299,11 @@ const Search: FunctionComponent<React.PropsWithChildren<SearchProps>> = ({
           indexName={TYPESENSE_COLLECTION_NAME}
           searchClient={searchClient}
           onStateChange={onSearchStateChange}
+          future={{
+            preserveSharedStateOnUnmount: true,
+          }}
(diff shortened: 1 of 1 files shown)


src/components/search/index.tsx
```diff
@@ -299,8 +299,11 @@ const Search: FunctionComponent<React.PropsWithChildren<SearchProps>> = ({
           indexName={TYPESENSE_COLLECTION_NAME}
           searchClient={searchClient}
           onStateChange={onSearchStateChange}
+          future={{
+            preserveSharedStateOnUnmount: true,
+          }}
```
zacjones93, line 304: this is going to be default behavior for widgets soon so algolia recommends this flag

src/components/search/index.tsx
```diff
@@ -299,8 +299,11 @@ const Search: FunctionComponent<React.PropsWithChildren<SearchProps>> = ({
...
           initialUiState={{
-            TYPESENSE_COLLECTION_NAME: {
+            [TYPESENSE_COLLECTION_NAME]: {
```
zacjones93, line 306: 😭 gotta use the actual index name

PR #1476, opened 2024-08-14
fix: filter out unpublished lessons in lesson list

zacjones93:
As I was adding lessons for Johns course, I notice we don't do any sort of filter of unpublished lessons which will show up in the list and navigate you to home when you click on them.

gif


Creeland (approved):
Nice improvement

PR #1479, opened 2024-08-27
fix: correctly set up ssr for /q

zacjones93:
We were never actually using `InstantSearchSSRProvider` correctly. It looks like there was a separate `BrandPage` component that had the SSR provider wrapped that seems like it was from some example code. This wraps our `Search` Component in the provider so the server will pass results to the client

https://www.loom.com/share/1a315170ae874435814045ff2606040a?sid=54199555-e00e-4081-99d5-8a8eb32e08be

gif)

### DIFF
--- a/src/pages/q/[[...all]].tsx
+++ b/src/pages/q/[[...all]].tsx
@@ -1,5 +1,5 @@
 import * as React from 'react'
-import {useRouter} from 'next/router'
+import singletonRouter, {useRouter} from 'next/router'
 import Image from 'next/image'
 import Search from '@/components/search'
 import {NextSeo} from 'next-seo'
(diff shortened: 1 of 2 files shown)


src/pages/q/[[...all]].tsx
```diff
@@ -77,8 +77,6 @@ const SearchIndex: any = ({
   const [instructor, setInstructor] = React.useState(initialInstructor)
   const [noIndex, setNoIndex] = React.useState(noIndexInitial)
   const debouncedState = React.useRef<any>()
-  const router = useRouter()
```
Comment (zacjones93) [inline src/pages/q/[[...all]].tsx:80]: because this gets rendered server-side, we can't `useRouter`

src/pages/q/[[...all]].tsx
```diff
@@ -121,7 +119,7 @@ const SearchIndex: any = ({
       const href: string = createUrl(searchState)
       setNoIndex(queryParamsPresent(href))
 
-      router.push(href, undefined, {
+      singletonRouter.push(href, undefined, {
```
Comment (zacjones93) [inline src/pages/q/[[...all]].tsx:124]: singletonRouter works here instead of router from `useRouter`

joelhooks (approved):
🎉

PR #1480, opened 2024-09-04
Properly forward ref to S3uploader

zacjones93:
The ref on the uploader was `null` which looking into it we need to properly forward the ref with `forwardRef` instead of passing it down as a regular prop.

gif

### DIFF
--- a/src/components/upload/modern-react-s3-uploader.tsx
+++ b/src/components/upload/modern-react-s3-uploader.tsx
@@ -121,7 +121,7 @@ const ModernReactS3Uploader: React.FC<ModernReactS3UploaderProps> = ({
     ...props,
   }
 
-  return <input className={className} {...inputProps} />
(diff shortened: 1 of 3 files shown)


src/components/upload/modern-react-s3-uploader.tsx
```diff
@@ -43,85 +43,94 @@ const clearInputFile = (f: HTMLInputElement) => {
...
+      scrubFilename = (filename) => filename.replace(/[^\w\d_\-\.]+/gi, ''),
+      s3path = '',
+      autoUpload = true,
+      ...props
+    },
+    ref,
+  ) => {
+    const [myUploader, setMyUploader] = useState<any>(null)
+
+    const fileRef = typeof ref !== 'function' ? ref : null
```
zacjones93, line 87: forwarded refs can be wrapped in a callback so we need to check for function here.

PR #1489, opened 2024-10-15
feat: add transcripts to sanity lesson publish flow

zacjones93:
While we are still publishing courses through sanity, this ensures transcripts and SRTs are added when a lesson is created. It will add the transcript to the rails object as well as the video resource in sanity.

The API for transcribing was already in place because of the work we've done with tips, just had to wire this up into the sanity publish flow

gif


src/inngest/functions/create-lesson.ts
```diff
@@ -164,6 +165,18 @@ export let createLesson = inngest.createFunction(
...
 
+    const {deepgram} = await step.run(
+      'Order Transcript [Deepgram]',
+      async () => {
+        return await orderDeepgramTranscript({
+          moduleSlug: event.data.body._id,
+          mediaUrl: event.data.body.videoResource.videoFile.url,
+          videoResourceId: event.data.body.videoResource._id,
+          eggheadLessonId: lessonObject.id,
+        })
```
zacjones93, line 176: This will kick off the `transcript-ready` event with necessary data. Added `eggheadLessonId` as a optional prop

src/inngest/functions/transcript-ready.ts
```diff
@@ -52,14 +54,41 @@ export const transcriptReady = inngest.createFunction(
...
+
+            let body = new URLSearchParams(lessonParams)
+
+            return await eggAxios.put(
+              `/api/v1/lessons/${event.data.eggheadLessonId}`,
+              body,
+            )
+          },
+        )
+      }
```
zacjones93, line 81: if `eggheadLessonId` is present then we want to post srt and transcript to that lesson object

PR #1490, opened 2024-10-16
feat: add upsert to publish flow

Creeland:
insert

### DIFF
--- a/src/inngest/functions/create-course/index.ts
+++ b/src/inngest/functions/create-course/index.ts
@@ -2,6 +2,7 @@ import {inngest} from '@/inngest/inngest.server'
 import {SANITY_COURSE_DOCUMENT_CREATED} from '@/inngest/events/sanity-course-document-created'
 import axios from 'axios'
 import {createClient} from '@sanity/client'
+import {upsertCourseToTypesense} from './upsertCourseToTypesense'
 
 const AXIOS_PARAMS = {
   baseURL: process.env.NEXT_PUBLIC_AUTH_DOMAIN || '',
@@ -105,6 +106,10 @@ export let createCourse = inngest.createFunction(
 
     await step.run('add-rails-id-to-sanity', async () => {
       return await saveCourseDataToSanity(sanityBody, courseObject.data)
+    })
+
+    await step.run('upsert-course-to-typesense', async () => {
+      return await upsertCourseToTypesense(courseObject.data)
     })
   },
 )
--- a/src/inngest/functions/create-course/upsertCourseToTypesense.ts
+++ b/src/inngest/functions/create-course/upsertCourseToTypesense.ts
@@ -1,0 +1,44 @@
+import Typesense from 'typesense'
+
+let client = new Typesense.Client({
+  nodes: [
+    {
+      host: process.env.NEXT_PUBLIC_TYPESENSE_HOST!,
+      port: 443,
+      protocol: 'https',
+    },
+  ],
+  apiKey: process.env.TYPESENSE_WRITE_API_KEY!,
+  connectionTimeoutSeconds: 2,
+})
+
+export async function upsertCourseToTypesense(course: any) {
+  await client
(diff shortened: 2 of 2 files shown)


src/inngest/functions/create-course/index.ts
```diff
@@ -89,22 +90,36 @@ export let createCourse = inngest.createFunction(
...
+        let message = `Course is not indexed, skipping upsert to Typesense. Course: ${sanityBody.title}`
+        console.log(message)
+        return message
+      } else if (sanityBody.productionProcessState !== 'published') {
+        let message = `Course is not published, skipping upsert to Typesense. Course: ${sanityBody.title}`
+        console.log(message)
+        return message
+      } else {
+        return await upsertCourseToTypesense(courseObject.data)
+      }
```
Creeland, line 122: Since this is only being run on the initial publish of the Sanity document, I'm checking for the indexed and published attributes on the document instead of deleting any existing documents in TypeSense since there won't be any.

joelhooks: deleting is a safe operation regardless. might make sense to have a `syncWithTypeSense` inngest function that can be run anytime/anywhere that handles removing and replacing.

you could call it here with `await step.invoke` ([docs for invoke](https://www.inngest.com/docs/reference/functions/step-invoke)) as a for instance

joelhooks: I'd probably, generally, use the `lesson` as the source of truth for all things display etc versus using sanity fields for the current state. mostly for consistency

src/inngest/functions/create-course/index.ts
```diff
@@ -89,22 +90,36 @@ export let createCourse = inngest.createFunction(
...
-        sanityBody.lessonIds,
-        courseObject.data.id,
-      )
-    })
+    // await step.run('add-lessons-to-course', async () => {
+    //   return await addLessonsToCourse(
+    //     sanityBody.lessonIds,
+    //     courseObject.data.id,
+    //   )
+    // })
```
joelhooks, line 98: it's not clear why these steps have been disabled

src/inngest/functions/create-course/upsertCourseToTypesense.ts
```diff
@@ -0,0 +1,84 @@
...
+        }),
+      ),
+      title: z.string(),
+      description: z.string(),
+      path: z.string(),
+    })
+    .safeParse(course)
+}
+
+export async function upsertCourseToTypesense(course: any) {
```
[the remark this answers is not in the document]
joelhooks: this whole file is in the inngest functions folder, but should probably be in `/lib` as an "api" that we use for performing this action

src/inngest/functions/create-course/utils/upsertCourseToTypesense.ts
```diff
@@ -0,0 +1,76 @@
...
+      label: z.string(),
+      image_url: z.string(),
+      http_url: z.string(),
+      description: z.string(),
+    }),
+  ),
+  title: z.string(),
+  description: z.string(),
+  path: z.string(),
+})
```
joelhooks, line 48: one thing I notice is that every property in this schema is required

is that the actual case? if not can probably loosen the type up a bit.

src/inngest/functions/typesense/delete-document.ts
```diff
@@ -0,0 +1,39 @@
...
+})
+
+export const deleteDocument = inngest.createFunction(
+  {id: 'delete-document', name: 'Delete Document'},
+  {event: COURSE_UNPUBLISHED_EVENT}, // Use the correct event here
+  async ({event, step}) => {
+    const {courseId, reason} = event.data
+
+    await step.run('delete-document', async () => {
+      console.log(`Deleting document with ID ${courseId}}`)
```
joelhooks, line 24: if you're done with them the logs can be removed since technically the logging is handled in inngest

PR #1492, opened 2024-11-05
Zj cp/end lifetime sale

zacjones93:
## allowPurchase=true
if `allowPurchase=true` is set as a query param then the user can purchase lifetime subscription
image

## no allowPurchase=true
if `allowPurchase` is anything other than `true` then this form will be displayed that will prompt the user to enter email to be notified when a future sale goes live. If the user is logged in there will just be a button.

This will add them to a segment in customerio `Interested in Lifetime` that is data-driven and checks if the attribute `interested_in_lifetime_since` exists. When the user signs up a data is set so we know when they signed up and can filter them out of the segment if we run a sale.

image
image

gif


src/components/pricing/lifetime/upcoming-sale.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+    // there will only be a submittedEmail if there is no viewer.
+    // subscriber can still exist because it looks at cookies in the browser but would be confusing for the user if a different email was used than the one they submitted
+    if (submittedEmail) {
+      const {contact_id} = await requestSignInEmail(submittedEmail)
+
+      cioIdentify(contact_id, {
+        email: submittedEmail,
+        [INTERESTED_IN_LIFETIME_SINCE]: Math.floor(Date.now() / 1000),
+      })
```
zacjones93, line 30: if a user isn't signed in, they can enter their email which will send a sign in request to rails (effectively creating an account if there isn't one) which we can use the `contact_id` to identify the user in CIO

src/components/pricing/lifetime/upcoming-sale.tsx
```diff
@@ -0,0 +1,116 @@
...
+    } else if (subscriber) {
+      cioIdentify(subscriber.id, {
+        [INTERESTED_IN_LIFETIME_SINCE]: Math.floor(Date.now() / 1000),
```
zacjones93, line 33: CIO needs you to divide by 1000 for their time format

src/utils/cio/subscriber-attributes.ts
```diff
@@ -0,0 +1,3 @@
+const INTERESTED_IN_LIFETIME_SINCE = 'interested_in_lifetime_since'
+
+export {INTERESTED_IN_LIFETIME_SINCE}
```
zacjones93, line 3: we didn't want to have special strings for CIO attributes in the file we were working with so created this util file to organize any attributes we are setting in egghead-next

Creeland (approved):

PR #1495, opened 2024-12-04
feat: move invoices to membership layout

zacjones93:
Links to invoices were only showing up if you had an active recurring membership. This doesn't make sense as anyone who's made a previous transaction might want to see it even if they are a lifetime member or no longer a paying member.

Also updates `Invoices` component to use trpc router (already defined) instead of loading transactions in a `useEffect`

M4AHOES1E0HZZ


joelhooks (approved):
🙏🏻

PR #1509, opened 2025-02-20
fix: find active account in useAccount

zacjones93:
image

gif

### DIFF
--- a/src/hooks/use-account.ts
+++ b/src/hooks/use-account.ts
@@ -32,9 +32,15 @@ export const useAccount = () => {
 
   const account =
     isAccountOwner &&
-    userAccounts?.find((account: {owner: {id: number}}) => {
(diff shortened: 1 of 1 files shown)


src/hooks/use-account.ts
```diff
@@ -32,9 +32,15 @@ export const useAccount = () => {
 
   const account =
     isAccountOwner &&
-    userAccounts?.find((account: {owner: {id: number}}) => {
-      return account.owner?.id === viewer.id
-    })
+    userAccounts?.find(
+      (account: {owner: {id: number}; subscriptions: {status: string}[]}) => {
+        const activeSubscription = account?.subscriptions?.find(
+          (sub) => sub.status === 'active',
```
zacjones93, line 38: we could add `sub.type === 'paid'` as well if we wanted to

PR #1549, opened 2025-12-16
fix: correct GROQ query overwriting lessons with wrong resources filter

Creeland:
The courseQuery GROQ had two "lessons" definitions - the second one using resources[@->["_type"] == "lesson"] was overwriting the first lessons[]-> definition, causing wrong lesson titles from other courses to appear.

Changes:
- Use lessons[]-> as primary source (the actual lessons array field)
- Add _resourceLessons fallback from resources[]-> for courses that use resources
- JavaScript fallback logic to use _resourceLessons if lessons is empty

This fixes the Beginner's Guide to React course showing wrong lesson titles like "Angular 1.x Redux: Introduction" instead of the correct React lessons.

## Summary by CodeRabbit

* **New Features**
  * Enhanced lesson metadata with slug, path, tags, and primary tag for clearer course organization.
  * Added a fallback lessons source and automatic fallback to populate lessons when primary entries are empty.

* **Bug Fixes / Improvements**
  * Improved slug-handling to reduce incorrect legacy matching and make course lookups more reliable.

* **Chores**
  * Filters out broken lesson references and removes temporary fields to keep course data consistent.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/src/lib/get-course-builder-metadata.ts
+++ b/src/lib/get-course-builder-metadata.ts
@@ -1,6 +1,6 @@
 import * as mysql from 'mysql2/promise'
-import { ConnectionOptions, RowDataPacket, Pool } from 'mysql2/promise'
-import type { Post } from '@/schemas/post'
+import {ConnectionOptions, RowDataPacket, Pool} from 'mysql2/promise'
+import type {Post} from '@/schemas/post'
 
 const access: ConnectionOptions = {
   uri: process.env.COURSE_BUILDER_DATABASE_URL,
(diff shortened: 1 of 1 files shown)


Creeland:
In src/lib/get-course-builder-metadata.ts, I made Course Builder queries only use the loose LIKE matching when the slug includes a ~<id> suffix (real Course Builder slugs).

For legacy slugs (no ~), queries now use exact match only (id = ? or fields.slug = ?), so they can’t accidentally match unrelated records and overwrite titles.

PR #1608, opened 2026-05-20
EGG-435: Remove legacy tips feature

zacjones93:
## Summary
- remove legacy /tips and /tipz routes, components, lib helpers, and tRPC router
- remove tip-specific Mux webhook and Inngest event/function wiring
- remove Sanity tip schema/desk entry and stale sitemap/docs references
- fix Course Builder metadata test mock typing so TypeScript can inspect mock call args

## Verification
- pnpm exec eslint src/server/routers/_app.ts src/inngest/inngest.config.ts src/inngest/inngest.server.ts src/app/api/webhooks/sanity/route.ts src/schemas/post.ts src/lib/format-content-as-prompt.ts src/lib/posts/get-post.ts src/lib/guides-from-course-builder.ts --ext ts,tsx
- pnpm exec tsc --noEmit
- git diff --check

## Summary by CodeRabbit

* **Documentation**
  * Updated form submission and background job flow diagrams.
  * Removed tip creation flow documentation.

* **Chores**
  * Removed tips feature pages, components, and server routes.
  * Removed tip-related tRPC procedures and Inngest background jobs.
  * Removed Sanity schema, webhooks, and CMS configuration for tips.
  * Updated sitemap and robots.txt configuration.
  * Updated content type validation and post type mappings.

* **Tests**
  * Added test case for tip row handling in post retrieval.

[Review Change Stack](https://app.coderabbit.ai/change-stack/skillrecordings/egghead-next/pull/1608?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


joelhooks (changes requested):
Blocking one stale-data path before this merges.

`src/schemas/post.ts` removes `postType: 'tip'`, but `src/lib/posts/get-post.ts` still selects any `type = 'post'` row for a direct `[post]` request before parsing with `PostSchema`.

Because `src/pages/[post].tsx` has `fallback: 'blocking'`, old/root tip slugs can still hit `getPost()`. If the DB row has `fields.postType = 'tip'`, `PostSchema.safeParse()` fails and `getPost()` throws, so the deleted feature turns into a 500 instead of a clean 404.

Small fix: add the same allowed `postType` filter to the single-post SQL (or explicitly return `null` for `postType === 'tip'`) and add a regression test that a tip row returns `null`/404. Then this removal looks good.

